PR #3, opened 2026-06-14
Organize design docs into OpenSpec capability specs

davitf:
Reorganize the existing SPEC.md, ARCHITECTURE.md, and COMPARISON.md design
docs into 20 OpenSpec capability specs under openspec/specs/, plus a
cross-cutting openspec/project.md. This is a first-pass transcription that
restructures existing material into the OpenSpec requirement/scenario format;
the source docs remain the prose reference and individual specs will be
refined later.

Capabilities: archive-reading, archive-writing, archive-data-model,
access-intent-and-cost, safe-extraction, bomb-protection, format-detection,
backend-registry, error-handling, progress-and-logging, per-format specs
(zip/tar/single-file-compressors/7z/rar/iso/directory),
seekable-decompressor-streams, testing-contract, and cli.

https://claude.ai/code/session_0183r6yxzNcJfo9QZXaXELBY


openspec/specs/format-detection/spec.md
```diff
@@ -0,0 +1,156 @@
+# Format Detection
+
+## Purpose
+
+Archivey can identify the archive format of a source (file path or binary stream) without fully opening it. Detection returns a `FormatInfo` dataclass carrying the detected format, a confidence score, and an encoding hint. Detection never consumes or discards bytes from the source.
+
+## Requirements
```
davitf, line 7: This seems to be missing the compressed tar detection we have in DEV and I think in an earlier spec.

And I'd also like to add detection for SFX archives (rar, 7z; what other formats do we support that may also have that?). maybe we need to look at the first bytes and see if they look like an executable, and if so seek further to determine if there's an archive in there, or seek to the end. I think the DEV detector had some unfinished code for that. And we should add support for them in the native parsers

openspec/specs/format-detection/spec.md
```diff
@@ -0,0 +1,156 @@
...
+#### Scenario: ISO file detected via extended peek
+
+- **WHEN** the source is a path with extension `.iso` and the stream is at least 32 774 bytes
+- **AND** the bytes `43 44 30 30 31` appear at offset 32 769
+- **THEN** `detect_format()` returns `FormatInfo(format=ArchiveFormat.ISO, confidence=1.0, detected_by="magic")`
+
+#### Scenario: stream too short for ISO detection
+
+- **WHEN** ISO detection is attempted on a stream shorter than 32 774 bytes
+- **THEN** `detect_format()` raises `FormatDetectionError`
```
davitf: no, it should just understand that the file can't be an ISO; many regular archives can be less than that

openspec/specs/format-iso/spec.md
```diff
@@ -0,0 +1,83 @@
+# ISO 9660 Archive Support (via pycdlib)
+
+## Purpose
+
+The system reads ISO 9660 disc images through the `pycdlib` library (optional `[iso]` extra). Because ISO 9660 supports multiple filesystem namespaces — plain ISO 9660, Joliet, and Rock Ridge — the backend auto-selects the richest available namespace and reports the selection so callers can reason about filename and metadata fidelity.
```
davitf, line 5: Can we also support bin files from the bin/cue format? It seems we'd just have to add a stream wrapper that strips the metadata from each 2048-byte block and pass the unwrapped stream to pycdlib. I think that's of lower priority, we can drop the requirement if it gets to be more complex than that

openspec/specs/format-detection/spec.md
```diff
@@ -0,0 +1,156 @@
...
+
+---
+
+### Requirement: Detection never consumes or discards bytes
+
+The system SHALL ensure that the bytes read during detection remain available to the backend that subsequently opens the archive.
+
+- For **seekable streams**: the stream position SHALL be restored to its original position (via `seek(0)`) after detection.
+- For **non-seekable streams**: the system SHALL wrap the stream in a `PeekableStream` that buffers the first `DETECTION_LIMIT` bytes and replays them transparently on subsequent reads.
```
davitf: that doesn't work if the function receives a non-peekable stream and the caller than keeps using it (so the "### Scenario: non-seekable stream is wrapped" below doesn't work). Either we need to wrap in a PeekableStream before calling the function, or the function itself returns the peekable stream after wrapping. What's best?

openspec/specs/format-rar/spec.md
```diff
@@ -0,0 +1,228 @@
...
+  O(member_size).
+- **Solid random access:** decode from the archive start up to the requested
+  member, or extract once with `unrar x` into a temporary directory and serve
+  subsequent reads from disk (cleaned up on `close()`).
+- **`extract_all`:** MAY use a one-shot `unrar x` to a temporary directory.
+
+#### Scenario: random access on a non-solid RAR
+
+- **WHEN** `ar.open(member)` is called on a non-solid RAR archive
+- **THEN** the backend reads just that member's data via `unrar`, doing O(member_size) work
```
davitf, line 112: rarfile has a hack for speeding up reading these files. unrar may parse all file metadata before actually extracting the file, which slows it down; to avoid it, for small members, rarfile creates a temporary rar file with just that file's contents and asks unrar to open that one. We should investigate with benchmarks if that's actually worth it, and what thresholds make sense. it may be a lower priority task when implementing the parser

openspec/specs/error-handling/spec.md
```diff
@@ -0,0 +1,123 @@
...
+- **WHEN** an archive uses a recognized feature the reader does not handle (e.g. multi-volume, RAR2, or the BCJ2 coder)
+- **THEN** `UnsupportedFeatureError` is raised rather than producing incorrect output
+
+---
+
+### Requirement: Required Attributes on Every ArchiveyError
+
+The system SHALL ensure that every instance of `ArchiveyError` (and any subclass) carries the following attributes:
+
+| Attribute | Type | Description |
```
davitf, line 88: archive_name as well? it may be helpful when printing error messages (although it may not always be available; if we open the archive from a stream, some include the filename or another name, but possibly not all)

openspec/specs/error-handling/spec.md
```diff
@@ -0,0 +1,123 @@
...
+```python
+try:
+    raw_member = self._zf.getinfo(name)
+except KeyError as exc:
+    raise ReadError(f"Member '{name}' not found", format=ArchiveFormat.ZIP) from exc
+except zipfile.BadZipFile as exc:
+    raise CorruptionError("ZIP central directory corrupt", format=ArchiveFormat.ZIP) from exc
+```
+
+For the common case, a `@translate_errors(format)` decorator handles exception wrapping uniformly across backend methods:
```
davitf: I don't fully understand this. Does translate_errors internally know all the library exceptions to translate? Are they scoped by format? Each library (not format) has its own exceptions, so maybe it should be a per-library function. If this format field is just to set the source_format field, I think this is weird. We shouldn't be setting the fields in the exception everywhere manually, to avoid repetitive errorprone code; the base class methods that may raise these exceptions should catch them, set the archive/member(/format?) fields and re-raise

openspec/specs/error-handling/spec.md
```diff
@@ -0,0 +1,123 @@
+# Error Handling
```
davitf, line 1: One clarification I think is missing here is that actual runtime unpredictable filesystem-level errors (e.g. a problem reading the archive, closed network connection) should still raise the original exception, as they're not directly related to archive decoding. Worth adding a requirement and/or scenario for this?

openspec/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,230 @@
...
+Provides a uniform interface for opening and reading archives across all supported formats. The `ArchiveReader` class presents ZIP, TAR, RAR, 7z, ISO, plain directories, and single-file compressed streams as interchangeable objects with consistent metadata, iteration, and data-access semantics.
+
+## Requirements
+
+### Requirement: Opening an archive for reading
+
+The system SHALL expose a top-level `archivey.open()` function that accepts a file path, `Path`, or binary stream and returns an `ArchiveReader`.
+
+```python
+archivey.open(
```
davitf: Didn't we decide to use `open_archive` for clarity? if someone imports the method directly, it will be clearer that it's not the built-in open

openspec/specs/format-7z/spec.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Scenario: unrecognized method ID
+
+- **WHEN** a folder uses a coder whose method ID is not recognized
+- **THEN** `UnsupportedFeatureError` is raised naming the unknown method ID
+
+#### Scenario: multi-volume 7z archive
+
+- **WHEN** a multi-volume 7z archive is opened
+- **THEN** `UnsupportedFeatureError` is raised rather than producing an incorrect member list
```
davitf: Someone mentioned that a multivolume 7z file is just a single stream, split into multiple files, and a simple stream joiner should make it possible to support it.

We should think about how to support multivolume rar/7z files: open_archive could receive a list of files/streams and treat them like a multi-format archive. additionally, if receiving a filename, it could detect it's a multivolume archive and open the whole set

openspec/specs/format-rar/spec.md
```diff
@@ -0,0 +1,228 @@
...
+The native parser SHALL raise `UnsupportedFeatureError` for archive shapes it does
+not support — multi-volume RAR sets and legacy RAR2 archives (extract version ≤ 20)
+— rather than mis-parsing them. For RAR5 members that carry only a Blake2sp hash and
+no CRC32, `Member.hashes` SHALL contain a `"blake2sp"` entry (bytes) and no
+`"crc32"` key — never a guessed CRC.
+
+#### Scenario: multi-volume RAR set
+
+- **WHEN** a multi-volume RAR archive is opened
+- **THEN** the system raises `UnsupportedFeatureError` rather than a partial or garbled member list
```
davitf: As in the 7z comment, let's add multivolume support

openspec/specs/format-single-file-compressors/spec.md
```diff
@@ -0,0 +1,65 @@
...
+
+Single-file compressors (GZ, BZ2, XZ, ZST) are presented as a one-member pseudo-archive through the unified `ArchiveReader` / `ArchiveWriter` interface. The compressed stream is treated as an archive containing exactly one file, with the member name inferred from the source filename. This allows single-file compressed streams to participate in the same iteration, extraction, and conversion workflows as multi-member archives.
+
+## Requirements
+
+### Requirement: Present a single-file compressor as a one-member archive
+
+The system SHALL present any GZ, BZ2, XZ, or ZST source as an archive containing exactly one `Member` of type `MemberType.FILE`. No directory members are synthesized. The member's name is inferred by stripping the compression extension from the source filename (e.g., `data.txt.gz` → `data.txt`). If no filename is available (e.g., the source is an anonymous stream), the member name defaults to `"data"`.
+
+#### Scenario: Member name inferred from filename
```
davitf: gzip archives may optionally store the original filename in a header field. I don't think it's good to use it for filename by default (maybe there can be a config option for this), but it should be added in raw_filename at least

openspec/specs/format-single-file-compressors/spec.md
```diff
@@ -0,0 +1,65 @@
...
+- **THEN** iterating the reader yields exactly one `Member`
+
+### Requirement: Report single-file compressor format properties
+
+The system SHALL expose the following cost and capability properties for every opened single-file compressor archive:
+
+| Property | Value |
+|----------|-------|
+| Listing cost | O(1) — one member always |
+| Access cost | SOLID (must decompress from start to reach any byte offset) |
```
davitf: Not true, some formats and/or libraries support limited seeking. Please see DEV implemenation and specs

openspec/specs/packaging-and-extras/spec.md
```diff
@@ -0,0 +1,144 @@
...
+- **AND** ZIP, TAR (all stdlib variants), GZ/BZ2/XZ, directory, and 7-Zip **reading of common codecs** all work, and RAR listing works (the `[7z]` bundle adds PPMd/Deflate64/Zstd/Brotli/AES for 7z; `[rar]` adds encrypted RAR5 headers and Blake2sp verification; without them a member with only a Blake2sp hash still reads, just unverified; RAR data reads need the system `unrar` binary)
+
+#### Scenario: writing 7z requires the [7z-write] extra
+
+- **WHEN** the core-only install is used to *write* a `.7z` archive
+- **THEN** the operation is unavailable until the `[7z-write]` extra is installed (see `backend-registry` for the exact `UnsupportedFormatError` + install-hint behavior); 7z *reading* still works with no extra
+
+---
+
+### Requirement: Optional Extras Enable Specific Formats
```
davitf, line 43: We should have a recommended set, some libraries included in all may not actually be useful (e.g. alternative zstandard library)

openspec/specs/format-detection/spec.md
davitf: Went with "wrap before calling": `open_archive()` now creates one `PeekableStream` for a non-seekable source and threads that same wrapper through both detection and the backend, so detection stays pure — it returns `FormatInfo` only and consumes nothing (peeks via `.peek()`). `detect_format()` standalone is documented as non-consuming for paths/seekable streams; for a raw non-seekable stream the caller must pass a `PeekableStream` (else the peeked prefix is lost). Updated the requirement and reworked the "non-seekable" scenarios accordingly.

---

openspec/specs/error-handling/spec.md
davitf: Agreed — reworked it. Translation is now two separable concerns: (1) **type translation is per-library** (each lib's exceptions → the right `ArchiveyError` subclass, by inspecting the exception, with no format/path/member knowledge — a lib shared across formats like `lzma` is translated once), and (2) **context is stamped centrally** — the `ArchiveReader` ABC wraps each public op and fills `source_format`/`archive_name`/`member_name` from the context it already holds, so backends never set those fields by hand. Dropped the old `@translate_errors(format)` decorator. Also added a requirement that genuine runtime/IO errors (`OSError`, dropped connections, `KeyboardInterrupt`) propagate **unwrapped** — only decoding-library errors get translated.

---

openspec/specs/safe-extraction/spec.md
```diff
@@ -458,14 +460,16 @@ The system SHALL return a `list[ExtractionResult]` from `archivey.extract()` and
 @dataclass
 class ExtractionResult:
     member: ArchiveMember
-    path: Path | None           # None if skipped
-    status: ExtractionStatus    # EXTRACTED, SKIPPED, REJECTED
+    path: Path | None            # the written path, or None if not written
+    status: ExtractionStatus
+    error: ArchiveyError | None = None   # the failure, for FAILED/REJECTED under OnError.CONTINUE
```
davitf: we could have non-Archivey exceptions here in case of filesystem errors while reading the source archive or writing the decompressed file

openspec/project.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Deferred / out of scope (v1)
+
+These are **decided** deferrals (a conscious "not for v1"):
+
+- In-place archive modification (append/update).
+- Encryption write for 7z/RAR.
+- Native sparse-file extraction (detected and flagged, extracted dense).
+- NTFS junction recreation on non-Windows.
+- Joining multi-volume archives (reported via `is_multivolume`, joining left to caller).
```
davitf: I think we ended up including this in the specs for rar/7z

CLAUDE.md
```diff
@@ -0,0 +1,63 @@
...
+library: read, stream, and safely extract ZIP / TAR / RAR / 7z / ISO / directory
+/ single-file-compressed archives behind one uniform interface.
+
+## Where things live
+
+- `SPEC.md`, `ARCHITECTURE.md`, `COMPARISON.md`, `PLAN.md` — the original prose
+  design docs (reference). `PLAN.md` is the phased implementation roadmap.
+- `openspec/specs/<capability>/spec.md` — the authoritative capability specs
+  (OpenSpec format: requirements + WHEN/THEN scenarios). When specs and the prose
+  docs disagree, the specs win.
```
davitf: actually, I'd rather you pause and ask me when there are discrepancies

openspec/changes/phase-1-scaffold-and-spine/tasks.md
```diff
@@ -0,0 +1,157 @@
...
+- [ ] 6.3 `logging`: *library emits no output by default*.
+- [ ] 6.4 `testing-contract`: framework stands up (matrix harness importable; oracle
+      hooks wired but skipped when libs absent).
+- [ ] 6.4b `format-directory`: opening a directory yields members with correct
+      filesystem metadata; `read`/`open` return file content; in-directory symlinks
+      follow; `cost` reports `INDEXED`/`DIRECT`/`SEEKABLE` — the spine validated
+      end-to-end against a real backend.
+
+**Gates**
+- [ ] 6.5 `uv run mypy src/` clean under `--strict`.
```
davitf: Actually, for type-checking, I'd rather use Pyrefly + ty. The library should be clean on both. Is it worth still keeping mypy and asking it to be clean on all three, or overkill? Apart from code correctness, I want users of the library to have correct checks + IDE autocompletion. let's decide and update mypy references throughout

openspec/changes/phase-2-stream-layer/proposal.md
```diff
@@ -0,0 +1,63 @@
...
+
+Because this is clean-slate, there is **no `io_helpers.py` to reorganize** and no
+re-export shim — the package is laid out correctly from the start, and the
+simplified `BinaryIOWrapper` is written with straightforward delegation (no
+`self.read = self._raw.read` method-replacement trick) rather than retrofitted.
+
+## What Changes
+
+- **`src/archivey/internal/streams/`**, with focused modules:
+  - `detect.py` — `RecordableStream`, `RewindableStreamWrapper` (ported;
```
davitf: do we need these, or are they being replaced with the PeekableStream that we defined in the SPEC.md?

openspec/changes/phase-2-stream-layer/tasks.md
```diff
@@ -0,0 +1,67 @@
...
+
+- [ ] 1.1 Create `src/archivey/internal/streams/` (with `__init__.py`).
+- [ ] 1.2 `detect.py` — port `RecordableStream`, `RewindableStreamWrapper`.
+- [ ] 1.3 `slice.py` — port `SlicingStream`.
+- [ ] 1.4 `compat.py` — port `is_seekable`, `is_stream`, `is_filename`,
+      `ensure_binaryio`, `ensure_bufferedio`, `fix_stream_start_position`,
+      `read_exact`; write a **simplified `BinaryIOWrapper`** fresh (plain delegation
+      + `readinto` fallback; **no** `self.read = self._raw.read`).
+- [ ] 1.5 Port the decompressor streams as `decompress.py` / `xz.py` / `lzip.py`;
+      leave `archive_stream.py` where it is.
```
davitf, line 34: We should write extensive tests for these streams and helpers, as they're essential building blocks. We should think of all the possible corner cases, and also look at the DEV stream tests to find any other scenarios that should be tested

openspec/changes/phase-2-stream-layer/tasks.md
```diff
@@ -0,0 +1,67 @@
...
+## 5. Verify — acceptance criteria
+
+**Spec scenarios covered**
+- [ ] 5.1 All of `compressed-streams`.
+- [ ] 5.2 All of `seekable-decompressor-streams`.
+
+**Gates**
+- [ ] 5.3 `uv run mypy src/` clean under `--strict`.
+- [ ] 5.4 `uv run ruff check` clean.
+- [ ] 5.5 New stream tests green; frozen oracle no worse.
```
davitf, line 84: Should we add a test coverage check, or can we trust the implementer to test all relevant cases? (I don't really like coverage checks, but might be an additional check) In this case, do we need to add a task for setting it up? (here or in phase 1?)

openspec/changes/phase-1-scaffold-and-spine/proposal.md
davitf, line 1: I'm missing a proposal / task for setting up CI tests on different environments (mac/windows/linux with different python and library versions. DEV has some 18 envs, which I think is overkill; but we should define what makes sense

CLAUDE.md
```diff
@@ -0,0 +1,63 @@
+# Archivey v2 — Agent Guide
```
davitf, line 1: I'm missing a file with coding standards (commenting, tests -- checking behavior not internal implementation, except in specific cases; when fixing bugs, use red-green TDD, and so on). would this come here, or be a separate one? do we even need it?

PR #5, opened 2026-06-17
feat: Phase 1 scaffold, spine, and directory backend

davitf:
Implement archivey v2 package scaffold with:
- pyproject.toml with hatchling build, optional extras, dev deps
- Core data model: ArchiveFormat (with named class-attr instances), ArchiveMember (mutable, unhashable, with .replace()), ArchiveInfo
- Error hierarchy (ArchiveyError and 17 subclasses)
- Intent/CostReceipt/ListingCost/AccessCost/StreamCapability enums
- BackendRegistry singleton with register_reader/register_writer
- BaseArchiveReader ABC with link resolution, member_id assignment, stream_members, open/read/get API
- DirectoryReader + DirectoryReadBackend (self-registers, uses os.walk)
- open_archive() public API (auto-detects directories)
- CI workflow for 12 matrix combinations (Linux/macOS/Windows x Python 3.11-3.14)
- 39 passing tests covering all directory backend functionality
- Dev oracle tests copied to tests/_dev_oracle/ (collection skipped, not yet ported)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01FEtJXXqjxWfrmPQaePVXyo


.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Install all extras + dev group
+        if: matrix.install == 'all'
+        run: uv sync --group dev --extra all
+
+      - name: Install unrar (Linux only, for RAR data tests)
+        if: matrix.install == 'all' && runner.os == 'Linux'
+        run: sudo apt-get update && sudo apt-get install -y unrar || true
+
+      - name: Run tests
+        if: matrix.install == 'all'
```
davitf, line 111: We should run the tests even in core-only, the non-dependent tests should work and the others should fail cleanly with PackageNotInstalledError (the tests should handle this expectation properly)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+    name: ${{ matrix.os }} / py${{ matrix.python-version }} / ${{ matrix.install }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Linux: every supported Python, both as a clean zero-dep core and full [all].
+          - { os: ubuntu-latest, python-version: "3.11", install: core-only }
+          - { os: ubuntu-latest, python-version: "3.12", install: core-only }
+          - { os: ubuntu-latest, python-version: "3.13", install: core-only }
```
davitf: I think we don't need all versions for core-only, 3.11 and 3.14 should be enough

src/archivey/formats/__init__.py
```diff
@@ -0,0 +1 @@
+from archivey.formats import directory_reader as _directory_reader  # noqa: F401
```
davitf, line 6: is importing these here needed for registration? if so, please add a comment explaining it, so we don't accidentally remove it

openspec/changes/phase-1-scaffold-and-spine/tasks.md
```diff
@@ -0,0 +1,184 @@
...
+> Python, the no-deps core install, and the OS-specific path/symlink/junction behaviour.
+
+- [x] 2b.1 GitHub Actions workflow (`.github/workflows/ci.yml`) running, per job,
+      `uv run ruff check`, `uv run pyrefly check`, `uv run ty check`, and
+      `uv run pytest` (with `pytest-cov` producing a **report only**, never failing the
+      build). Use `uv` for env setup; cache the uv environment keyed on `uv.lock`.
+- [x] 2b.2 Matrix (~12 jobs):
+      - **Linux** × Python `{3.11, 3.12, 3.13, 3.14}` × install `{core-only (no extras),
+        [all]}` = 8 jobs — `core-only` asserts the zero-dep core imports and runs with
+        no third-party packages; `[all]` runs the full suite (extras present; `unrar`
```
davitf: As mentioned above, we should run the test suite with core-only and the tests should be skipped or have an expected failure if the libs are not available. You can look at some DEV tests for similar behavior

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def _iter_members(self) -> Iterator[ArchiveMember]:
+        for dirpath, dirnames, filenames in os.walk(self._root, followlinks=False):
+            rel_dirpath = os.path.relpath(dirpath, self._root)
+            if rel_dirpath == ".":
+                rel_dirpath = ""
+
+            # Sort for deterministic order
+            dirnames.sort()
+            filenames_sorted = sorted(filenames)
```
davitf: why not sort filenames in-place? the original order is not used anywhere, right?

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,194 @@
...
+
+        size = st.st_size if member_type == MemberType.FILE else None
+
+        return ArchiveMember(
+            type=member_type,
+            name=name,
+            raw_name=name.encode("utf-8", errors="surrogateescape"),
+            size=size,
+            compressed_size=size,  # no compression
+            modified=modified,
```
davitf, line 147: what about the other timestamps, accessed and created? can they be filled as well?

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,194 @@
...
+        self,
+        root: Path,
+        intent: Intent,
+        archive_name: str | None,
+    ) -> None:
+        super().__init__(ArchiveFormat.DIRECTORY, intent, archive_name)
+        self._root = root
+
+    def _iter_members(self) -> Iterator[ArchiveMember]:
+        for dirpath, dirnames, filenames in os.walk(self._root, followlinks=False):
```
davitf: Use os.scandir() instead of walk, so we retrieve all the stats for the dir at once instead of having to call os.stat() for each one (or some other similar approach). Or, as root() is a Path, use the pathlib-equivalent functions if they exist; it would also simplify the path manipulation below

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,194 @@
...
+        link_target: str | None,
+    ) -> ArchiveMember:
+        # Use forward slashes
+        name = rel_path.replace(os.sep, "/")
+
+        # Timestamps: use mtime as modified
+        modified = datetime.fromtimestamp(st.st_mtime, tz=timezone.utc)
+
+        uid: int | None = getattr(st, "st_uid", None)
+        gid: int | None = getattr(st, "st_gid", None)
```
davitf: Why getattr? can the stat_result object not have these fields defined? I don't see this indication in the python docs. if it's a case that could happen, please add a comment

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,194 @@
...
+        uname: str | None = None
+        gname: str | None = None
+        try:
+            import grp
+            import pwd
+
+            if uid is not None:
+                uname = pwd.getpwuid(uid).pw_name
+            if gid is not None:
+                gname = grp.getgrgid(gid).gr_name
```
davitf: Should we cache these values with a local dict inside this iteration or DirectoryReader, since most files will have the same user/group? or is it quick to retrieve them from the os every time, or do getpwuid/getgruid already cache them internally?

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+
+class MemberType(Enum):
+    FILE = "file"
+    DIRECTORY = "directory"
+    SYMLINK = "symlink"
+    HARDLINK = "hardlink"
+    OTHER = "other"
+
+
+class CompressionAlgo(Enum):
```
davitf: let's name this CompressionAlgorithm please. (update the spec as well)

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+    BZ2: ClassVar[ArchiveFormat]
+    XZ: ClassVar[ArchiveFormat]
+    ZST: ClassVar[ArchiveFormat]
+    SEVEN_Z: ClassVar[ArchiveFormat]
+    RAR: ClassVar[ArchiveFormat]
+    ISO: ClassVar[ArchiveFormat]
+    DIRECTORY: ClassVar[ArchiveFormat]
+    UNKNOWN: ClassVar[ArchiveFormat]
+
+    def file_extension(self) -> str:
```
davitf, line 58: where will this be used? should we have a special case for directories (and if so, what? "/"?)

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+    def __repr__(self) -> str:
+        # Return the named instance name if available
+        for name, val in _FORMAT_INSTANCES.items():
+            if val == self:
+                return f"ArchiveFormat.{name}"
+        return f"ArchiveFormat({self.container!r}, {self.stream!r})"
+
+
+# Registry of named format instances
+_FORMAT_INSTANCES: dict[str, ArchiveFormat] = {}
```
davitf: do we need this? can't ArchiveStream look at its own fields to find the named instances? (then we wouldn't need a separate name arg in _add_format also)

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+    BCJ2 = "bcj2"
+    DELTA = "delta"
+    UNKNOWN = "unknown"
+
+
+@dataclass(frozen=True)
+class CompressionMethod:
+    algo: CompressionAlgo
+    level: int | None = None
+    properties: bytes | None = None
```
davitf: add docs/comments explaining the fields, in a way that will be surfaced by the doc generator when we have that. I think SPEC.md has explanations.

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+    VFAT = 14
+    ALTERNATE_MVS = 15
+    BEOS = 16
+    TANDEM = 17
+    OS_400 = 18
+    OS_X_DARWIN = 19
+    UNKNOWN = 255
+
+
+def _normalize_name(
```
davitf: Where is this used? I couldn't find it (but it may be just Github not finding it). But it doesn't seem to belong in types.py as it's not a type definition and doesn't seem to be referred anywhere else in this file

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+    def is_link(self) -> bool:
+        return self.type in (MemberType.SYMLINK, MemberType.HARDLINK)
+
+    @property
+    def is_other(self) -> bool:
+        return self.type == MemberType.OTHER
+
+    @property
+    def is_junction(self) -> bool:
+        return self.type == MemberType.SYMLINK and bool(self.extra.get("zip.is_junction"))
```
davitf: won't this only work for zip files, if the extra dict is scoped per format? 7z and RAR also support junctions. I think this scoping might not actually be appropriate for this field. what do you propose? and we should define a constant for this instead of hardcoding the string here.

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def replace(self, **kwargs: Any) -> "ArchiveMember":
+        """Return a copy with the given fields changed; never mutates self."""
+        from dataclasses import replace as dc_replace
```
davitf: why not put this as a top-level import? it's not that expensive

src/archivey/internal/types.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @property
+    def is_other(self) -> bool:
+        return self.type == MemberType.OTHER
+
+    @property
+    def is_junction(self) -> bool:
+        return self.type == MemberType.SYMLINK and bool(self.extra.get("zip.is_junction"))
+
+    def replace(self, **kwargs: Any) -> "ArchiveMember":
```
davitf, line 297: I think we might include a field indicating that it is a mutated ephemeral member, maybe even a reference to the original. what do you think?

src/archivey/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    ArchiveMember,
+    CompressionAlgo,
+    CompressionMethod,
+    ContainerFormat,
+    CreateSystem,
+    MemberType,
+    StreamFormat,
+)
+
+__all__ = [
```
davitf, line 51: Is this required? won't the module already export everything it imports above if we don't specify it manually? I'm concerned of having to update two places, it seems more error-prone. 

if we need it (for mypy / IDE autocompletion / docs), can we generate it dynamically? Google suggested `__all__ = [name for name in globals() if not name.startswith('_')]` might work, but I don't know enough.

please push back on this if you think it's worth keeping as-is.

src/archivey/internal/types.py
davitf: Renamed `CompressionAlgo` → `CompressionAlgorithm` (and updated `CompressionMethod.algo`, the `__init__` export, and `__all__`).

One heads-up: the matching spec (`openspec/specs/archive-data-model/spec.md`) doesn't exist on this branch — it lives only on PR #3 (`claude/openspec-organize-docs-yljuo4`), which hasn't merged. I didn't want to fork the spec text here and create a merge conflict, so I left the spec edit for PR #3. Want me to push the rename to that branch too, or fold it in once #3 merges?

---

src/archivey/internal/types.py
davitf: It isn't called by our v2 code yet — only the frozen DEV oracle uses it (naming a converted/nested output file by extension, and as a param id). It's specced for extension-based naming/detection, so I kept it but fixed it properly:

- Found a latent bug: `GZ = (RAW_STREAM, GZIP)` returned `"raw_stream.gz"`. Now `RAW_STREAM` returns just the codec ext → `"gz"`/`"bz2"`/`"xz"`.
- `DIRECTORY` and `UNKNOWN` have no on-disk file form, so they return `""` (rather than `"directory"` or `"/"`).
- Added a docstring documenting the intended use and these cases.

If you'd rather not carry an unused method at all, I'm happy to drop it until a backend needs it.

---

src/archivey/internal/types.py
davitf: Agreed the `"zip."` scoping is wrong — junctions show up in 7z and RAR too. My proposal: a single unscoped key, since it's a cross-format concept. I added `EXTRA_IS_JUNCTION = "is_junction"` in `types.py` and `is_junction` now reads `extra.get(EXTRA_IS_JUNCTION)`.

The spec (`archive-data-model`, on PR #3) still says `extra["zip.is_junction"]`, so it'll need the same change there.

Longer term, if junctions turn out common enough we could promote this to a real boolean field on `ArchiveMember` instead of an `extra` key — but a named constant felt like the right weight for now. Say the word if you'd prefer the dedicated field.

---

src/archivey/internal/types.py
davitf: I left this one **un-implemented** pending your call, since it's a real design fork rather than a cleanup.

My lean is to **not** add an "ephemeral/mutated" flag or a back-reference to the original right now:
- The current contract is clean — `.replace()` returns an independent copy and the original is untouched. A back-reference reintroduces shared mutable state (and a retain cycle / GC concern) that the immutability-by-convention model was meant to avoid.
- We don't have a consumer yet. The natural consumer is the extraction/iteration **filters** (Phase 4), where knowing "what did the filter change vs. the original" could matter for logging/auditing.

So my suggestion: revisit when filters land in Phase 4, and if we do want it, model it explicitly (e.g. `replaced_from: ArchiveMember | None` set by `.replace()`, excluded from `__eq__`) rather than a bare boolean. Happy to add it now instead if you'd rather bake it into the data model from the start — just let me know.

---

src/archivey/__init__.py
davitf: Pushing back gently: I'd keep `__all__` explicit. The dynamic `[name for name in globals() if not name.startswith('_')]` trick is the thing that bites here — at the point it runs, the module namespace also contains `version` and `PackageNotFoundError` (imported from `importlib.metadata` to compute `__version__`), so a naive sweep would publish those as public API. Any future top-level helper import would silently leak the same way.

What I did instead to address the maintenance concern:
- `del PackageNotFoundError, version` after `__version__` is set, so the namespace stays clean.
- Added `tests/test_public_api.py`, which fails if `__all__` drifts — every entry must be a real attribute, and every public (non-module, non-dunder) symbol must be listed. So adding an export without updating `__all__` (or vice-versa) breaks CI rather than going unnoticed.

That gives the "can't forget to update two places" safety without the fragility of generating it. If you still prefer generated, I can switch to `[n for n, o in vars().items() if not n.startswith('_') and not ismodule(o)]` now that the namespace is clean — but I think the explicit list + guard test reads better.

---

openspec/changes/phase-2-stream-layer/tasks.md
```diff
@@ -5,6 +5,29 @@
...
+      `_stamp_error_context()` so format/archive/member context is attached. Add a test
+      that a raw decode error surfaces as a stamped `ArchiveyError`.
+- [ ] 0.2 **(Reminder for the streaming readers, Phase 5+)** the base
+      `_iter_with_data()` default eagerly registers all members and is for indexed
+      backends only; forward-only/solid readers MUST override it (correctness, not
+      efficiency) and must not call `_get_members_registered()`. See the docstring in
+      `internal/reader.py`. Nothing to do in Phase 2 — just don't build on the default
+      when the first streaming backend lands.
+- [ ] 0.3 **(Reminder for Phase 3) open-time fail-fast on non-seekable sources.** The
+      method-level intent enforcement is done (`SEQUENTIAL` is forward-only; see the
```
davitf: change to the new streaming arg

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _iter_members(self) -> Iterator[ArchiveMember]:
+        yield from self._scan(self._root, "")
+
+    def _scan(self, directory: Path, rel_prefix: str) -> Iterator[ArchiveMember]:
+        # os.scandir yields DirEntry objects whose stat() is cached, so we avoid a
+        # separate os.stat()/os.lstat() syscall per entry.
+        try:
+            with os.scandir(directory) as it:
+                entries = sorted(it, key=lambda e: e.name)
```
davitf, line 65: order non-dirs first, so the iterator outputs all regular files and then starts going into the subdirs

src/archivey/formats/directory_reader.py
```diff
@@ -0,0 +1,193 @@
...
+            return
+
+        for entry in entries:
+            rel_path = rel_prefix + entry.name
+            try:
+                st = entry.stat(follow_symlinks=False)
+            except OSError:
+                continue
+
+            if entry.is_symlink():
```
davitf, line 80: what about Windows junctions? it seems that starting in python 3.12, we could use entry.is_junction() to detect them:

```
is_junction()
Return True if this entry is a junction (even if broken); return False if the entry points to a regular directory, any kind of file, a symlink, or if it doesn’t exist anymore.

The result is cached on the os.DirEntry object. Call [os.path.isjunction()](https://docs.python.org/3/library/os.path.html#os.path.isjunction) to fetch up-to-date information.

Added in version 3.12.
```

please add a test for that which runs in the Windows CI

src/archivey/internal/open_archive.py
```diff
@@ -0,0 +1,59 @@
...
+from pathlib import Path
+from typing import BinaryIO
+
+from archivey.internal.errors import FormatDetectionError
+from archivey.internal.reader import ArchiveReader
+from archivey.internal.registry import get_registry
+from archivey.internal.types import ArchiveFormat
+
+
+def open_archive(
```
davitf, line 27: I'd move this to an open_archive.py or another named file. I don't like putting actual logic in __init__.py files, it tends to hide them

src/archivey/internal/api.py
```diff
@@ -0,0 +1,59 @@
...
+    Format detection (Phase 3) is not yet wired; only DIRECTORY is auto-detected here.
+    """
+    # Import formats package to ensure backends are registered
+    import archivey.formats  # noqa: F401
+
+    if isinstance(password, str):
+        password = password.encode()
+
+    archive_name: str | None = None
+    if isinstance(source, (str, Path)):
```
davitf: should we get the stream name in case it's a stream?

src/archivey/internal/api.py
```diff
@@ -0,0 +1,59 @@
...
+    If source is a directory path, opens it as a directory pseudo-archive.
+    Format detection (Phase 3) is not yet wired; only DIRECTORY is auto-detected here.
+    """
+    # Import formats package to ensure backends are registered
+    import archivey.formats  # noqa: F401
+
+    if isinstance(password, str):
+        password = password.encode()
+
+    archive_name: str | None = None
```
davitf: This is being used only to raise the format exception below, so do it inside the if block. I see that the DirectoryReadBackend.open_read() gets the archive name from the path in a similar way, and other backends will likely do it as well, so probably better to move it to a common helper function (or pass this name as an arg to open_read)

src/archivey/internal/cost.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from enum import Enum
+
+
+class ListingCost(Enum):
+    INDEXED = "indexed"
+    REQUIRES_SCANNING = "requires_scanning"
+    REQUIRES_DECOMPRESSION = "requires_decompression"
```
davitf, line 26: Add doctrings to these enum values (in all enums in this file) explaining in more detail what they mean, that will be surfaced on the generated docs.

src/archivey/internal/cost.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class StreamCapability(Enum):
+    SEEKABLE = "seekable"
+    FORWARD_ONLY = "forward_only"
+
+
+@dataclass(frozen=True)
+class CostReceipt:
+    listing_cost: ListingCost
```
davitf, line 60: Add field docstrings

PR #7, opened 2026-06-19
feat: Phase 2 stream layer (compressed + seekable)

davitf:
Build the internal/streams/ package and the shared codec/crypto/verify
stack that every later format backend composes (the compressed-streams and
seekable-decompressor-streams capabilities), plus wire the
exception-translation spine left dangling in Phase 1.

- streams/compat.py: ported is_seekable/is_stream/ensure_*/read_exact and a
  freshly written simplified BinaryIOWrapper (plain delegation, no method-swap)
- streams/slice.py: SlicingStream (ported)
- streams/archive_stream.py: ArchiveStream exception translate/stamp carrier
- streams/decompress.py / xz.py / lzip.py: seekable decompressor streams
  (XZ block-index, lzip trailer-scan), exceptions mapped to the v2 hierarchy
- streams/codecs.py: codec registry with one default backend per spec-table
  codec, resolve-without-open, accelerator dispatch behind [seekable], and
  per-codec error translation
- streams/crypto.py: single wrapped AES stage (interface + [crypto] gating;
  concrete decrypt deferred to Phase 7)
- streams/verify.py: decompressed-output digest verification stage
- internal/config.py: StreamConfig + tri-state AcceleratorMode (internal)
- reader.py: _translate_exception hook + _wrap_member_stream (task 0.1);
  logs.py: streams + integrity loggers

New tests cover the primitive units, all compressed-streams scenarios, the
seekable scenarios, and the spine translation. pyrefly + ty + ruff clean;
137 passed, 2 skipped.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XpZehm2wE9ZT5WwB2vpwvq


src/archivey/internal/streams/binaryio.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    def __init__(self, raw: Any) -> None:
+        super().__init__()
+        self._raw = raw
+
+    def read(self, size: int = -1, /) -> bytes:
+        data = self._raw.read(size)
+        # A blocking stream returns b"" at EOF; normalise a None (non-blocking "no data
+        # yet") to b"" so the wrapper presents a plain blocking BinaryIO.
```
davitf: isn't it dangerous to turn None into b"""? Maybe it report an eof when it hasn't actually been reached. Can you find out in what causes this could legitimately happen, so we can think about the proper behavior?

src/archivey/internal/streams/binaryio.py
```diff
@@ -0,0 +1,205 @@
...
+        tell = getattr(self._raw, "tell", None)
+        if tell is None:
+            raise io.UnsupportedOperation("tell")
+        return tell()
+
+    def readable(self) -> bool:
+        return True
+
+    def writable(self) -> bool:
+        return hasattr(self._raw, "write")
```
davitf: won't a file open for reading still have a write method, that just raises an error? should we check the raw stream's writable method if it has one?

src/archivey/internal/streams/binaryio.py
```diff
@@ -0,0 +1,205 @@
...
+def ensure_bufferedio(obj: Any) -> io.BufferedIOBase:
+    """Return ``obj`` as a buffered reader, without taking ownership of it.
+
+    A raw stream is wrapped in a non-closing ``BufferedReader``; an already-buffered
+    stream is returned unchanged.
+    """
+    if isinstance(obj, io.BufferedIOBase):
+        return obj
+    raw: io.RawIOBase
+    if isinstance(obj, io.RawIOBase):
```
davitf, line 326: maybe worth commenting on why this is checked here instead of just using enaure_binaryio: IIRC the BufferedStream constructor fails if the object is not a RawIOBase, even if it's stream-like. 

Now that I think about it, maybe ensure_binaryio should ensure a RawIOBase (or we should have a separate ensure_rawiobase)?

tests/test_stream_primitives.py
davitf, line 1: does it make sense to split this into test_binaryio for the functions and classes in there, and another test for the remaining things?

src/archivey/internal/streams/binaryio.py
davitf: Good catch — you're right, coercing `None`→`b""` is dangerous. Investigated: a `read()` returns `None` **only** for a non-blocking stream that has no data available *right now*; at EOF both blocking and non-blocking streams return `b""`. So `None` never means EOF, and turning it into `b""` would signal a false EOF and silently truncate the data.

archivey's readers pull synchronously and can't make progress on a non-blocking source anyway, so I changed `read()` (and `readinto()`) to raise `BlockingIOError` on `None` rather than fabricate EOF — it surfaces the real condition loudly instead of losing data. (`BlockingIOError` is an `OSError` subclass, so it propagates unchanged through the exception translators, which is correct for a usage error rather than archive corruption.)

I went with "raise" over "propagate `None`" because propagating would just move the failure downstream (callers and codec libs treat the result as `bytes`) and would force `read() -> bytes | None`, which conflicts with the `BinaryIO` contract. Happy to switch to propagation if you'd prefer the faithful-RawIOBase route. Done in 7107ddc; covered by `test_wrapper_read_none_raises_blocking_not_eof` / `test_wrapper_readinto_none_raises_blocking`.

---

src/archivey/internal/streams/binaryio.py
davitf: Confirmed and fixed. Every `io.IOBase` defines `write()` even when opened read-only — it just raises (`UnsupportedOperation`, or `OSError` for `GzipFile`), so `hasattr(raw, "write")` reported `True` for read-only files. `writable()` now delegates to `raw.writable()` when present, and only falls back to `hasattr(write)` for partial file-likes that don't implement `writable()` (e.g. urllib3's `HTTPResponse`, which omits `write()` entirely). Made `readable()` symmetric. Regression test: `test_wrapper_writable_trusts_raw_writable_not_hasattr_write` opens a real read-only file and asserts `writable() is False` despite `hasattr(write)`. Done in 7107ddc.

---

src/archivey/internal/streams/binaryio.py
davitf: Added the comment — exactly your reasoning: `BufferedReader` rejects an object that isn't an `io.RawIOBase`, so a merely stream-like source is adapted via `BinaryIOWrapper` (which *is* a `RawIOBase`) first. That's why this branches on `RawIOBase` rather than calling `ensure_binaryio`.

On the broader question: I left `ensure_binaryio` as "give me a `BinaryIO`" (its result may be a `BufferedIOBase` like a passed-through `BytesIO`, which `BufferedReader` would reject) and documented that callers needing a `RawIOBase` should use `ensure_bufferedio`, which handles it internally. I didn't add a separate `ensure_rawiobase` since `ensure_bufferedio` is the only consumer with that requirement today — but if more callers need a raw stream later, factoring the "coerce to RawIOBase" step out is the natural move. Let me know if you'd rather I add it now. Done in 7107ddc.

---

tests/test_stream_primitives.py
davitf: Done — split into `test_binaryio.py` (the `binaryio.py` functions/classes: `read_exact`, `is_*`, `BinaryIOWrapper`, `ensure_*`) and `test_slice.py` (`SlicingStream` + `fix_stream_start_position`). Also added `test_stream_inputs.py`, a parametrized cross-library matrix that runs the helpers against every stream type a caller might realistically supply. Done in 7107ddc.

---

openspec/specs/format-single-file-compressors/spec.md
```diff
@@ -1,14 +1,24 @@
...
+only ever appear inside a container (ZIP/7z), so they are reachable through
+`compressed-streams` but not as single-file compressors here.
 
 ## Requirements
 
 ### Requirement: Present a single-file compressor as a one-member archive
 
-The system SHALL present any GZ, BZ2, XZ, or ZST source as an archive containing exactly one `ArchiveMember` of type `MemberType.FILE`. No directory members are synthesized. The member's name is inferred by stripping the compression extension from the source filename (e.g., `data.txt.gz` → `data.txt`). If no filename is available (e.g., the source is an anonymous stream), the member name defaults to `"data"`.
+The system SHALL present any GZ, BZ2, XZ, ZST, LZ4, LZIP, ZLIB, BR, or Z source as an archive containing exactly one `ArchiveMember` of type `MemberType.FILE`. No directory members are synthesized. The member's name is inferred by stripping the compression extension from the source filename (e.g., `data.txt.gz` → `data.txt`; likewise `.bz2`, `.xz`, `.zst`, `.lz4`, `.lz`, `.zz`, `.br`, `.Z`). If no filename is available (e.g., the source is an anonymous stream), the member name defaults to `"data"`.
```
davitf: what happens if the compressed file doesn't have the proper extension? we shouldn't remove arbitrary extensions. otoh. the extension may be wrong. my suggestion: if the extension matches a known compressed stream extension, remove it. otherwise, add .uncompressed . what do you think?

openspec/specs/format-single-file-compressors/spec.md
```diff
@@ -1,14 +1,24 @@
...
+Single-file compressors (GZ, BZ2, XZ, ZST, LZ4, LZIP, ZLIB, BR, Z) are presented as a one-member pseudo-archive through the unified `ArchiveReader` / `ArchiveWriter` interface. The compressed stream is treated as an archive containing exactly one file, with the member name inferred from the source filename. This allows single-file compressed streams to participate in the same iteration, extraction, and conversion workflows as multi-member archives.
+
+The set is the standalone-stream side of the `compressed-streams` codec table: every codec
+that can stand alone as a bare `.ext` stream (no container) appears here, including
+**unix-compress (`.Z`, LZW)** via the `uncompresspy` backend (the `[unix-compress]`
+extra). unix-compress has two quirks: it requires a **seekable** source (the backend
```
davitf, line 10: I checked the library code once, I think it would be simple to make it work with non seekable streams. we could add a low priority future task to fork it into our code and fix it (or send a fix upstream)

src/archivey/internal/streams/binaryio.py
davitf: can we add a test demonstrating why we can't let the underlying non buffering io be closed, if we don't have it yet? Iirc the main problem is that the buffered io would be closed on destruction due to gc or context closing, but if the user gives the library a stream we should never close it once we're done decompressing. we should probably explain this reasoning here. try creating a failing test with this behavior removed, and green with it present

davitf: does the binary io wrapper above work correctly regarding closing as well? what behavior should it have?

src/archivey/internal/streams/decompress.py
```diff
@@ -0,0 +1,422 @@
+"""Seekable decompressor-stream base class and the zlib/deflate backend.
```
davitf: and brotli

src/archivey/internal/streams/verify.py
```diff
@@ -0,0 +1,154 @@
...
+        return _Crc32Hasher
+    if name in hashlib.algorithms_available:
+        return lambda: hashlib.new(name)
+    return None
+
+
+def _matches(hasher: _IncrementalHasher, expected: int | bytes) -> bool:
+    digest = hasher.digest()
+    # Digests may be stored as raw bytes or as a big-endian integer (e.g. CRC32 as int).
+    if isinstance(expected, int):
```
davitf: why don't we do this conversion inside the crc32 hasher digest()? I think it would be cleaner

davitf (commented):
x

src/archivey/internal/streams/binaryio.py
davitf: Both behaviours are intentional and now tested + documented (f7193fa).

`BinaryIOWrapper` already left the wrapped stream open (`close()` only marks the wrapper closed, never touching `self._raw`) — correct for the same reason as the buffer: it adapts a stream the *caller* owns, and the close is usually implicit (a `with` block, or GC finalizing the wrapper), so it must not take the user's stream down with it. Expanded the comment to say so; `test_wrapper_does_not_close_underlying` covers it.

For the buffer, I added the contrast test you asked for: `test_plain_bufferedreader_closes_source_demonstrates_why_we_detach` shows a plain `io.BufferedReader` *does* close its source on `close()`, alongside `test_ensure_bufferedio_does_not_close_raw_source` proving ours doesn't — so the detach-on-close guard is demonstrably meaningful. The reasoning (implicit close via GC/`with` must never close a caller-owned stream) is now in the `_NonClosingBufferedReader` docstring.

---

openspec/specs/format-single-file-compressors/spec.md
davitf: Agreed — blindly stripping is wrong. Updated the spec (f7193fa) to your rule: if the source filename ends in a **recognized** single-file-compressor extension (`.gz`/`.bz2`/`.xz`/`.zst`/`.lz4`/`.lz`/`.zz`/`.br`/`.Z`, case-insensitive) it's stripped; otherwise the extension is left intact and `.uncompressed` is appended (`mystery.bin` → `mystery.bin.uncompressed`, `backup` → `backup.uncompressed`). No filename at all still defaults to `"data"`. Added a scenario for the non-compression-extension case.

---

openspec/specs/format-single-file-compressors/spec.md
davitf: Added as a backlog item in IDEAS.md (f7193fa): make unix-compress (`.Z`) work on non-seekable sources — `uncompresspy`'s decode path looks straightforward to make forward-only, so fork that bit into archivey or fix it upstream. Until then a `.Z` pipe/socket reports `StreamNotSeekableError`.

---

src/archivey/internal/streams/decompress.py
davitf: Done (f7193fa) — line 1 now reads "the zlib/deflate and Brotli backends", and the body sentence already lists Brotli alongside zlib/deflate.

---

src/archivey/internal/streams/verify.py
davitf: Good catch — for CRC32 we were converting int→bytes in `digest()` then bytes→int back in `_matches`, a pointless round-trip. `digest()` has to stay `bytes` (the hashlib contract), so rather than push the conversion into the hasher I normalized the **expected** value to bytes once, at construction, in the hasher's own width (`_expected_as_bytes`, via `digest_size`). Verification is now a plain `hasher.digest() == expected` compare with no per-read conversion (f7193fa).

---

PR #13, opened 2026-06-22
Phase 3 Stage 1: foundation + ZIP backend (detect→select→open→read→cost)

davitf:
Implements the first Phase-3 stage: the shared detection/registry machinery
plus the ZIP backend, end-to-end through open_archive().

- PeekableStream (internal/streams/peekable.py): buffers a non-seekable
  source's prefix so detection never consumes bytes the backend still needs;
  peek/replay/forward-only semantics, never closes the underlying stream.
- detect_format() + FormatInfo/DetectionConfidence (internal/detection.py):
  magic-first (CERTAIN) -> extension (GUESS); tables aggregated from each
  backend's MAGIC/EXTENSIONS data; magic/extension conflict warns on
  archivey.detection; non-consuming for paths/seekable/peekable sources.
- Registry rework: always-register; tri-state compositional FormatSupport
  (FULL/PARTIAL/NONE) via the module-or-None sentinel + codec availability;
  MissingComponent/FormatAvailability; format_availability /
  list_supported_formats / list_known_formats; install-hint selection errors.
- ReadBackend ABC: EXTENSIONS now Mapping[str, ArchiveFormat], MAGIC carries
  the implied format per entry, plus INSTALL_HINT.
- ZIP backend (formats/zip_reader.py): stdlib zipfile; INDEXED/DIRECT/SEEKABLE;
  ZipInfo->ArchiveMember mapping (mode, extended-timestamp precedence, type,
  compression, is_encrypted, raw_name); O(1) name lookup; non-seekable
  fail-fast; multi-volume rejection.
- open_archive(): detects, wraps non-seekable sources in PeekableStream,
  enforces REQUIRES_SEEK fail-fast, threads encoding hint.
- Public API exposure + new test suites (peekable, detection, registry, zip).

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01QVdJG1mJyrPRfYVbU6PT59

### DIFF
--- a/src/archivey/formats/single_file_reader.py
+++ b/src/archivey/formats/single_file_reader.py
@@ -8,7 +8,8 @@ logic lives in a small **dispatch table of metadata hooks** (gzip's stored filen
 xz/lzip decompressed size), so a new standalone codec becomes readable by adding the codec
 + enum + detection entry — no new backend class (see ``format-single-file-compressors``).
 
-ZST/LZ4 standalone reading is deferred to Phase 8 and intentionally not in ``FORMATS``.
+ZST and LZ4 are first-class standalone formats here (their codecs already exist from
+Phase 2); only their *seekable-decompressor* refinements remain for Phase 8.
 """
 
 from __future__ import annotations
@@ -125,6 +126,13 @@ class SingleFileReader(BaseArchiveReader):
                 archive_name=archive_name,
             )
 
+        # A non-seekable source cannot be randomly accessed, so a random-access accelerator
+        # (rapidgzip/indexed_bzip2) would only fail trying to seek it — keep the codec in
+        # sequential mode for such a source regardless of the archive's streaming flag.
+        self._codec_config = StreamConfig(
+            streaming=self._streaming or not self._seekable
+        )
+
         self._member = self._build_member(archive_name)
 
     def _build_member(self, archive_name: str | None) -> ArchiveMember:
@@ -173,7 +181,7 @@ class SingleFileReader(BaseArchiveReader):
         if not isinstance(self._source, Path):
             return None
         try:
-            backend = resolve_codec(self._codec, StreamConfig(streaming=self._streaming))
+            backend = resolve_codec(self._codec, self._codec_config)
             stream = backend.open(str(self._source))
         except (ArchiveyError, OSError, ValueError):
             return None
@@ -198,7 +206,7 @@ class SingleFileReader(BaseArchiveReader):
         return open_codec_stream(
             self._codec,
             codec_source,
-            config=StreamConfig(streaming=self._streaming),
+            config=self._codec_config,
             stamp=lambda exc: self._stamp_error_context(exc, member.name),
         )
 
--- a/src/archivey/formats/single_file_reader.py
+++ b/src/archivey/formats/single_file_reader.py
@@ -43,7 +43,6 @@ from archivey.internal.types import (
     ArchiveFormat,
     ArchiveInfo,
     ArchiveMember,
-    ContainerFormat,
     MemberType,
     StreamFormat,
 )
@@ -54,6 +53,8 @@ _EXTENSIONS: dict[str, ArchiveFormat] = {
     ".gz": ArchiveFormat.GZ,
     ".bz2": ArchiveFormat.BZ2,
     ".xz": ArchiveFormat.XZ,
+    ".zst": ArchiveFormat.ZST,
+    ".lz4": ArchiveFormat.LZ4,
     ".lz": ArchiveFormat.LZIP,
     ".zz": ArchiveFormat.ZLIB,
     ".br": ArchiveFormat.BROTLI,
@@ -67,6 +68,8 @@ _FORMATS: tuple[ArchiveFormat, ...] = (
     ArchiveFormat.GZ,
     ArchiveFormat.BZ2,
     ArchiveFormat.XZ,
+    ArchiveFormat.ZST,
+    ArchiveFormat.LZ4,
     ArchiveFormat.LZIP,
     ArchiveFormat.ZLIB,
     ArchiveFormat.BROTLI,
@@ -275,6 +278,8 @@ class SingleFileBackend(ReadBackend):
         (0, b"\x1f\x8b", ArchiveFormat.GZ),
         (0, b"BZh", ArchiveFormat.BZ2),
         (0, b"\xfd7zXZ\x00", ArchiveFormat.XZ),
+        (0, b"\x28\xb5\x2f\xfd", ArchiveFormat.ZST),
+        (0, b"\x04\x22\x4d\x18", ArchiveFormat.LZ4),
         (0, b"LZIP", ArchiveFormat.LZIP),
         (0, b"\x1f\x9d", ArchiveFormat.Z),
         # zlib's 2-byte CMF/FLG header — a weak signal (see detection's _WEAK_MAGIC_FORMATS).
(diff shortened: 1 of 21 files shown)


src/archivey/formats/single_file_reader.py
```diff
@@ -0,0 +1,317 @@
...
+        self._seekable = not is_stream(source) or is_seekable(source)
+
+        # unix-compress (.Z) decodes via random access, so it needs a seekable source; the
+        # other single-file formats read fine from a non-seekable source.
+        if format.stream == StreamFormat.UNIX_COMPRESS and not self._seekable:
+            raise StreamNotSeekableError(
+                "unix-compress (.Z) archives require a seekable source.",
+                source_format=format,
+                archive_name=archive_name,
+            )
```
davitf: The uncompresspy opener already raises it when we try to open a non-seekable source, and the translator already converts it into the proper error. So it seems to be redundant here.

Is this due to the library only actually opening the stream when _open_member is called, which doesn't happen during initialization? This may cause other invalid-format issues in other libraries to get triggered only on open as well. I believe DEV handled this by eagerly opening the stream during init, so any init-time errors would already show up. (I think it would cache and serve it on the first open, but create new ones for further opens, so multiple streams still worked)

src/archivey/formats/single_file_reader.py
```diff
@@ -0,0 +1,317 @@
...
+            return
+        xlen = int.from_bytes(header[pos : pos + 2], "little")
+        pos += 2 + xlen
+    if flg & 0x08:  # FNAME: null-terminated stored filename
+        end = header.find(b"\x00", pos)
+        if end != -1:
+            name_bytes = header[pos:end]
+            member.raw_name = name_bytes
+            member.extra["gzip.original_filename"] = name_bytes.decode(
+                "utf-8", errors="replace"
```
davitf: According to RFC 1952:
            If FNAME is set, an original file name is present,
            terminated by a zero byte.  The name must consist of ISO
            8859-1 (LATIN-1) characters; on operating systems using
            EBCDIC or any other character set for file names, the name
            must be translated to the ISO LATIN-1 character set.  This
            is the original name of the file being compressed, with any
            directory components removed, and, if the file being
            compressed is on a file system with case insensitive names,
            forced to lower case. There is no original file name if the
            data was compressed from a source other than a named file;
            for example, if the source was stdin on a Unix system, there
            is no file name.

So utf-8 doesn't seem to be the correct encoding. Let's fix this with red/green TDD: please add a test with a file generated with the gzip command line containing a filename with non-ascii characters, verify that it fails with the current incorrect encoding, and then fix the encoding and check that it passes.

src/archivey/formats/single_file_reader.py
```diff
@@ -0,0 +1,317 @@
...
+    member.size = reader._probe_decompressed_size()
+
+
+_METADATA_HOOKS: dict[
+    StreamFormat, Callable[[SingleFileReader, ArchiveMember], None]
+] = {
+    StreamFormat.GZIP: _gzip_metadata,
+    StreamFormat.XZ: _sized_stream_metadata,
+    StreamFormat.LZIP: _sized_stream_metadata,
+}
```
davitf: Shouldn't all of these live inside the SingleFileReader? It's weird that they're passed the reader and then call private reader methods

src/archivey/internal/detection.py
```diff
@@ -0,0 +1,193 @@
...
+from archivey.internal.streams.codecs import Codec, is_codec_available
+from archivey.internal.streams.peekable import DETECTION_LIMIT, PeekableStream
+from archivey.internal.streams.streamtools import is_seekable, read_exact
+from archivey.internal.types import ArchiveFormat
+
+# The zlib 2-byte header (CMF/FLG) is not a true magic number — the same prefix begins
+# many raw-deflate streams and can occur in arbitrary data — so a zlib magic match is the
+# weakest entry in the table: it is consulted only after the stronger magics and the
+# content probe, and a conflicting extension overrides it (see ``format-detection``).
+_WEAK_MAGIC_FORMATS: frozenset[ArchiveFormat] = frozenset({ArchiveFormat.ZLIB})
```
davitf: Shouldn't this live inside the magic tuple in single_file_reader.py (add a new is_weak item to the tuple) instead of here? It would make this detector more format-agnostic.

src/archivey/internal/detection.py
```diff
@@ -0,0 +1,193 @@
...
+            logger.warning(
+                "Format conflict for %r: extension suggests %r but magic bytes indicate "
+                "%r; using the magic-byte result.",
+                name,
+                ext_fmt,
+                strong_fmt,
+            )
+        return FormatInfo(strong_fmt, DetectionConfidence.CERTAIN, "magic")
+
+    # 2. Weak magic (zlib's 2-byte header): a real but low-confidence signal. A conflicting
```
davitf: Should we do a content probe for zlib, similar to what we do for brotli?

something like:
1. check magic, return False if no match
2. feed 256 bytes into the decompressor, see if it works or fails

davitf (commented):
I'm wondering if we should have a StreamBackend class to encapsulate the logic that is currently scattered in codecs.py, single_file_reader.py and detection.py. For each backend, we could specify:
1. opener
2. translator
3. magic bytes
4. content probe
5. extra metadata extraction (pass an ArchiveMember and let it fill any fields it can)
6. required libraries and install hints

and then both the detection and single-file-reader could be totally (or mostly) format-agnostic. WDYT?

davitf (commented):
Also, I see that zstd support is using the zstandard library. IIRC, DEV supported both zstandard and pyzstd, and the latter is what the 3.14 builtin support is based on. zstandard was finicky (the no-backward-seek issue, and it also wouldn't raise some errors I think), so it was relegated to a secondary option. Can you investigate pyzstd vs zstandard, or at least create a new spec for doing this investigation and deciding which one we should use (or whether to support both)?

src/archivey/formats/zip_reader.py
```diff
@@ -0,0 +1,301 @@
...
+    else:
+        try:
+            dos_modtime = datetime(*info.date_time)
+        except ValueError:
+            logger.warning("Invalid ZIP date_time for %r: %r", info.filename, info.date_time)
+            dos_modtime = None
+
+    pos = 0
+    extra = info.extra or b""
+    while pos + 4 <= len(extra):
```
davitf, line 102: I believe the extended timestamp field provides mtime / ctime / atime. We should fill all of them in the ArchiveMember if they're present

src/archivey/formats/zip_reader.py
```diff
@@ -0,0 +1,301 @@
...
+        password: bytes | None,
+        encoding: str | None,
+        archive_name: str | None,
+    ) -> None:
+        super().__init__(ArchiveFormat.ZIP, streaming, archive_name)
+        self._password = password
+        self._encoding = encoding
+        # Normalized member name -> ZipInfo, for O(1) data access without re-deriving the
+        # stored filename. Populated as members are materialized in _iter_members().
+        self._info_by_name: dict[str, zipfile.ZipInfo] = {}
```
davitf: Mapping names to members is a function of the BaseArchiveReader, it already has exactly this in a get(name) method.

Here we might need to get member -> info, right? I seem to recall in the spec that the ArchiveMember should have a private field (or maybe even public) with the backend info object, so by getting the member we should have the info object needed to open the file. Is that not in the spec?

src/archivey/formats/zip_reader.py
```diff
@@ -0,0 +1,301 @@
...
+            extra={"zip.compress_type": info.compress_type},
+        )
+
+    def _open_member(self, member: ArchiveMember) -> BinaryIO:
+        if not self._info_by_name:
+            # Materialize the member list so the name->ZipInfo map is populated (e.g. when
+            # opening a member object without a prior listing pass).
+            self._get_members_registered()
+        try:
+            info = self._info_by_name[member.name]
```
davitf: In the (rare) case of multiple members with the same name, this could get the wrong info. As stated in the other comment, we should use the info object stored in the actual member. (or, if that's not an option, at least we should store infos by member_id and not name, to avoid these conflicts; and it should be something that the base reader handles)

src/archivey/formats/zip_reader.py
```diff
@@ -0,0 +1,301 @@
...
+        self._encoding = encoding
+        # Normalized member name -> ZipInfo, for O(1) data access without re-deriving the
+        # stored filename. Populated as members are materialized in _iter_members().
+        self._info_by_name: dict[str, zipfile.ZipInfo] = {}
+
+        if is_stream(source) and not is_seekable(source):
+            raise StreamNotSeekableError(
+                "ZIP archives cannot be read from a non-seekable source (the central "
+                "directory lives at the end of the file). Buffer the source to disk or a "
+                "BytesIO and reopen.",
```
davitf: Don't provide this suggestion in the exception, it's not user-actionable (how will they "reopen"). Just state the problem, trust the developer to know what to do with it. Also in other exceptions in this module.

PR #14, opened 2026-06-23
fix(codecs): translate accelerator corruption + backstop rapidgzip truncation

davitf:
Fixes bad-input handling for the optional random-access accelerators (`rapidgzip` / `indexed_bzip2`, the `[seekable]` extra). `test_codecs.py` only ever exercised the **stdlib** decompressor paths, so the accelerators' different exception taxonomies were uncovered — and a review of the Stage-2 single-file work surfaced this.

## What was wrong

- **rapidgzip, corrupt header** → leaked a raw `RuntimeError` (`"Failed to parse gzip/zlib header"` / `"Invalid gzip/zlib wrapper"`), violating the `compressed-streams` translation contract.
- **indexed_bzip2, corrupt block** → leaked a raw `RuntimeError` (`"[BZip2 block header] Invalid Huffman…"`).
- **rapidgzip, truncation** → does **not** reliably raise. Investigation: it raises (`std::exception`) for a cut that leaves a partially-decodable block, but **silently returns short/zero output** otherwise (its `EndOfFileReached` doesn't always reach Python; `parallelization=1` makes no difference). So it's a core behavior inconsistency, not a wrapper-swallow or a config toggle.

## Fix

- Translate the two corrupt-input patterns → `CorruptionError`.
- **ISIZE truncation backstop** in `_open_gzip` for seekable **path** sources: on a full read to EOF, compare the decompressed length (`mod 2³²`) to the gzip ISIZE trailer; mismatch ⇒ `TruncatedError`. Concatenated **multi-member** gzip (whose trailer is only the *last* member's size) is disambiguated by a conservative scan for a further gzip header — so a valid file is **never** false-flagged (the only cost is a rare missed truncation whose partial deflate coincidentally contains a gzip header). A caller `seek` disables the check (random access invalidates the sequential total; partial reads are never verified).

## Tests

`tests/test_accelerator_corruption.py` — both accelerators on corrupt / truncated / intact / multi-member input, forced ON and skipped when the package is absent. Gates green locally (ruff/ty/pyrefly, 376 passed / 7 skipped).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XpZehm2wE9ZT5WwB2vpwvq)_

### DIFF
--- a/src/archivey/internal/streams/codecs.py
+++ b/src/archivey/internal/streams/codecs.py
@@ -79,30 +79,6 @@ _rapidgzip = _optional("rapidgzip")
 _rapidgzip_bzip2 = getattr(_rapidgzip, "IndexedBzip2File", None)
 
 
-def _close_accelerator(inner: BinaryIO) -> None:
-    """Close a ``rapidgzip`` accelerator object so its worker threads are stopped.
-
-    Used as a :func:`weakref.finalize` callback: it takes the raw accelerator object as an
-    argument (never a closure over the wrapper), so ``finalize`` keeps that object alive
-    until this runs.
-
-    ``close()`` is what actually stops the worker thread — and it is sufficient on its own:
-    ``join_threads()`` alone does **not** stop it (a joined-but-not-closed object still aborts
-    the interpreter at finalization with "Detected Python finalization from running … thread"),
-    whereas ``close()`` reliably does, so we just close. The shutdown canary in
-    ``tests/test_accelerator_shutdown.py`` enforces this (a never-closed object aborts; a
-    closed/guard-closed one does not).
-
-    ``inner`` is a :class:`BinaryIO` (the accelerator object after ``ensure_binaryio``), so
-    ``close()`` is guaranteed to exist; the ``try`` only guards against it *raising* during
-    finalization, where a propagated exception would just become unraisable-hook noise.
-    """
-    try:
-        inner.close()
-    except Exception:  # noqa: BLE001 - best-effort; the object is going away regardless
-        pass
-
-
 class _AcceleratorStream(io.RawIOBase, BinaryIO):
     """Wrap a threaded accelerator (``rapidgzip`` / ``indexed_bzip2``) so its underlying object
     is always *closed* before it is freed.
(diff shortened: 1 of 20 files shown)


src/archivey/internal/streams/codecs.py
```diff
@@ -70,7 +71,101 @@ def _optional(name: str) -> ModuleType | None:
...
+    def __init__(self, inner: BinaryIO) -> None:
+        super().__init__()
+        self._inner = inner
+        self._finalize = weakref.finalize(self, _close_accelerator, inner)
+
+    def read(self, n: int = -1, /) -> bytes:
+        return self._inner.read(n)
+
+    def readinto(self, b: "WriteableBuffer", /) -> int:
+        raw_readinto = getattr(self._inner, "readinto", None)
```
davitf: doesn't ensure_binaryio ensure this is true?

src/archivey/internal/streams/codecs.py
```diff
@@ -70,7 +71,101 @@ def _optional(name: str) -> ModuleType | None:
...
-_indexed_bzip2 = _optional("indexed_bzip2")
+# bzip2 random access is provided by rapidgzip's *bundled* IndexedBzip2File, NOT the separate
+# ``indexed_bzip2`` package. Loading both rapidgzip and indexed_bzip2 into one process corrupts
+# the heap and aborts on macOS (they statically bundle an overlapping C++ core, whose symbols
+# collide under dyld). Routing both gzip and bzip2 through rapidgzip keeps a single accelerator
+# library in the process, which is safe on every platform. See docs/known-issues.md.
+_rapidgzip_bzip2 = getattr(_rapidgzip, "IndexedBzip2File", None)
+
+
+def _close_accelerator(inner: BinaryIO) -> None:
```
davitf: could we fold this inside _AcceleratorStream.close()?

src/archivey/internal/streams/codecs.py
```diff
@@ -70,7 +71,101 @@ def _optional(name: str) -> ModuleType | None:
...
+
+    def close(self) -> None:
+        if self.closed:
+            return
+        # Run the guard (joins + closes the inner object) once; it is then disarmed.
+        self._finalize()
+        super().close()
+
+
+def _track_accelerator(inner: BinaryIO) -> _AcceleratorStream:
```
davitf: Do we still need this? could we just let the callers create the _AcceleratorStream directly?

src/archivey/internal/streams/codecs.py
```diff
@@ -375,7 +571,12 @@ def _open_gzip(source: CodecSource, params: CodecParams, config: StreamConfig) -
                 "The 'rapidgzip' package is required for gzip random access "
                 "(install the 'seekable' extra)."
             )
-        return ensure_binaryio(_rapidgzip.open(source, parallelization=0))
+        stream = _track_accelerator(ensure_binaryio(_rapidgzip.open(source, parallelization=0)))
```
davitf: could we move the ensure_binaryio call to inside the _AcceleratorStream construction? (and as per the other comment, build _AcceleratorStream here directly)?

src/archivey/internal/streams/codecs.py
```diff
@@ -366,6 +472,96 @@ def _open_stored(source: CodecSource, params: CodecParams, config: StreamConfig)
...
+            return
+        raise TruncatedError(
+            "gzip stream is truncated: the decompressed size does not match the ISIZE "
+            "trailer (the rapidgzip accelerator does not surface this truncation itself)"
+        )
+
+    def _has_additional_gzip_member(self) -> bool:
+        try:
+            with open(self._source_path, "rb") as f:
+                blob = f.read()
```
davitf: This is reading the whole file into memory, which must not be done! Maybe we should scan in blocks. But also, we should ideally support multiple blocks, and add the sizes to compare (but how to avoid false positives?).

Before committing to this full thing, I'd like to do some more testing. on a naive test, rapidgzip only failed to report any error when my test file was exactly 10 bytes long:

0 Failed to detect a valid file format.
1 Failed to detect a valid file format.
2 Failed to detect a valid file format.
3 Failed to detect a valid file format.
4 Failed to detect a valid file format.
5 Failed to detect a valid file format.
6 Failed to detect a valid file format.
7 Failed to detect a valid file format.
8 Failed to detect a valid file format.
9 Failed to detect a valid file format.
10 OK 0

[stderr] Unexpected end of file when getting block at 10 B 0 b (block index: 0) on demand
11 std::exception

[stderr] Unexpected end of file when getting block at 10 B 0 b (block index: 0) on demand
12 std::exception

...and so on. so maybe it's a specific corner case we can detect easily, or it's so specific/rare that we can ignore it? 

Maybe we just do the scanning fix for now to avoid extending this PR too much, but create a new openspec change for determining exactly how rapidgzip handles corruption/truncation and potentially improving or removing this

PR #15, opened 2026-06-24
openspec: codec-descriptor refactor + compression-library evaluation proposals

davitf:
Two OpenSpec change proposals coming out of the PR #13 review discussion. **Proposals only** — `proposal.md` + `tasks.md` with inline spec deltas, mirroring the `rapidgzip-truncation-investigation` layout. No code changes; the existing suite stays green.

## 1. `codec-descriptor-refactor`

Your "should we have a `StreamBackend` class?" idea. Today the per-codec logic is split across four modules:

- `codecs.py` — open fn + exception translator + availability sentinel
- the format backends — `MAGIC` / `EXTENSIONS` / `CONTENT_PROBE_FORMATS` data
- `single_file_reader.py` — the per-codec metadata hooks
- `registry.py` — `_CODEC_REQUIREMENT` (package / extra / install hint)

The proposal consolidates all of it into one **`StreamCodec` descriptor** per codec, so `detect_format()`, `SingleFileBackend`, and the registry availability all become data-driven (a new standalone codec = one descriptor, not four edits). Explicitly **behavior-preserving** — the existing detection/single-file/registry suites are the regression net and must stay green unchanged. Container formats (ZIP/TAR/ISO/7z) stay separate.

## 2. `compression-library-evaluation`

Your zstd/xz library question, generalized. Investigates the candidate libraries per codec and records the decision in a new **`docs/library-analysis.md`** (per-codec matrix: non-seekable support · efficient seeking · corruption detection · truncation detection · error-reporting fidelity · install/availability · maintenance).

Open questions it resolves:
- **zstd** — `zstandard` (current, with the no-backward-seek workaround) vs `pyzstd` (the intended migration target) vs stdlib `compression.zstd` (3.14+) vs `backports.zstd` vs `indexed_zstd` (seekable, ratarmount's choice). Notes that pyzstd's `SeekableZstdFile` only reads the Seekable Zstd format, not arbitrary `.zst`.
- **xz** — our own `xz.py` parser vs `python-xz`; record *why* (the exploration isn't written down anywhere), and resolve the dangling dep.
- Records the chosen + rejected library for every other codec too.

It also cleans up the **dead `python-xz` / `pyzstd` pins** in the `[all]` extra (both confirmed unused in `src/`). Actual library swaps are deferred to their own follow-up changes once the decisions land here.

The two changes coordinate: the descriptor's `open` / `translate` / `requirement` fields are where any library swap chosen by the evaluation ultimately gets wired.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVdJG1mJyrPRfYVbU6PT59)_

### DIFF
--- a/openspec/changes/compression-library-evaluation/specs/packaging-and-extras/spec.md
+++ b/openspec/changes/compression-library-evaluation/specs/packaging-and-extras/spec.md
@@ -1,0 +1,22 @@
+# Packaging and Extras — delta (compression-library evaluation)
+
+## MODIFIED Requirements
+
+### Requirement: Optional extras map to exactly the libraries the code uses
+
+The extras → capability mapping SHALL list only libraries the library actually imports; an
+extra MUST NOT pin a dependency that no code path in `src/` uses. The per-codec library
+choice and its rationale SHALL be recorded in `docs/library-analysis.md`, which is the
+source of truth for why each library is used or rejected. The currently-dead `python-xz`
+and `pyzstd` pins in the `[all]` extra SHALL be removed (or wired up if the evaluation
+decides to use them).
+
+#### Scenario: no dead optional dependency
+
+- **WHEN** the `[all]` extra (or any extra) is audited against `src/` imports
+- **THEN** every pinned package is reachable from some code path, or it is removed
+
+#### Scenario: the zstd extra matches the chosen backend
+
+- **WHEN** the evaluation selects the zstd decode backend
+- **THEN** the `[zstd]` extra pins exactly that package (plus any adopted seekable-zstd backend), and `docs/library-analysis.md` records the choice
--- a/openspec/changes/compression-library-evaluation/specs/packaging-and-extras/spec.md
+++ b/openspec/changes/compression-library-evaluation/specs/packaging-and-extras/spec.md
@@ -4,19 +4,31 @@
 
 ### Requirement: Optional extras map to exactly the libraries the code uses
 
-The extras → capability mapping SHALL list only libraries the library actually imports; an
-extra MUST NOT pin a dependency that no code path in `src/` uses. The per-codec library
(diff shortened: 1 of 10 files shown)


openspec/changes/compression-library-evaluation/specs/packaging-and-extras/spec.md
```diff
@@ -0,0 +1,22 @@
+# Packaging and Extras — delta (compression-library evaluation)
+
+## MODIFIED Requirements
+
+### Requirement: Optional extras map to exactly the libraries the code uses
+
+The extras → capability mapping SHALL list only libraries the library actually imports; an
```
davitf: There may be libraries imported just for testing (e.g. rarfile, py7zr). How do they fit here? Are they in the dev dependencies?

davitf: Good distinction — clarified in `d3e3fbd`. Test-only libraries live in the `dev` dependency group, not in user-facing extras, and the "no dead optional dependency" rule applies to **extras vs `src/` imports**. Current state and the cleanup it implies:

- `rarfile`, `py7zr` (decode oracles), `ncompress` (`.Z` fixture generator) — already in the `dev` group. ✓ Correct, stay there.
- `pyzstd` — currently mis-placed in the `[all]` **extra**, but it's used only by the dev test oracle to *generate* zstd fixtures → **move to `dev`** (unless the zstd evaluation promotes it to the runtime `[zstd]` backend, in which case it belongs in that extra).
- `python-xz` — imported nowhere (not `src/`, not the tests) → **remove** from `[all]`.

I added a "test-only library lives in the dev group, not an extra" scenario to the `packaging-and-extras` delta and a guard task so a user-facing extra can't silently pin a non-`src/` dependency again.

---

PR #16, opened 2026-06-24
refactor(codecs): unify per-codec logic behind one StreamCodec descriptor

davitf:
Implements the `codec-descriptor-refactor` OpenSpec change: a behavior-preserving refactor that makes each single-stream codec fully data-driven.

## What changed

Each single-stream codec (gzip, bzip2, xz, lzip, zlib, zstd, lz4, brotli, unix-compress, …) is now described by a single `StreamCodec` descriptor in `internal/streams/codecs.py`, bundling everything the rest of the library needs to know about it:

| Field | Replaces |
|---|---|
| `open` / `translate` | `_CodecSpec` + `_REGISTRY` |
| `stream_format` | `_STREAM_FORMAT_CODECS` |
| `magic` (incl. `weak`) / `content_probe` / `extensions` | `SingleFileBackend.MAGIC` / `CONTENT_PROBE_FORMATS` / `EXTENSIONS` |
| `extract_metadata` | `SingleFileReader._METADATA_HOOKS` |
| `requirement` | `registry._CODEC_REQUIREMENT` |

`detect_format()`, `SingleFileBackend`/`SingleFileReader`, and `format_availability()` all derive from the descriptor registry instead of three parallel hand-maintained tables. Adding a standalone codec is now "register one descriptor" rather than editing four files.

## Notable decisions

- **Import cycle:** `MissingComponent` moved to the leaf `internal/types.py` (re-exported from `registry`, so `archivey.MissingComponent` is unchanged), letting the descriptor carry its `requirement` without a registry↔codecs cycle — as the change's task 1.2 anticipated.
- **No reader coupling:** metadata hooks receive a `MetadataContext` (`peek_header` + `probe_decompressed_size`) rather than the reader instance, so the codec layer keeps no dependency on `formats/`.
- **Detection subtlety preserved:** `content_probe` is reserved for *magic-less* formats (Brotli). zlib stays a *weak-magic* match the detector confirms via its existing probe — it is deliberately **not** added to the magic-less probe list, which would have changed detection.

## Verification

- The **467 pre-existing** detection / single-file / registry / zip tests pass **unchanged** (the regression net for "behavior-preserving").
- Adds `tests/test_codec_descriptor.py`: registers a synthetic descriptor and asserts it becomes detectable, readable as a one-member archive, and availability-reported with no other edits (469 tests total).
- `uv run ruff check`, `uv run pyrefly check`, `uv run ty check` all clean.
- `openspec validate codec-descriptor-refactor --strict` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVdJG1mJyrPRfYVbU6PT59)_

### DIFF
--- a/src/archivey/formats/single_file_reader.py
+++ b/src/archivey/formats/single_file_reader.py
@@ -99,6 +99,9 @@ class SingleFileReader(BaseArchiveReader):
         # Keep the codec sequential for such a source regardless of the archive's streaming flag.
         self._codec_config = StreamConfig(streaming=self._streaming or not self._seekable)
 
+        # The compressed-source header, read at most once and cached (only the gzip metadata
+        # hook needs it; codecs without header metadata never trigger a read). See _peek_header.
+        self._header_cache: bytes | None = None
         self._member = self._build_member(archive_name)
         # Open the decompression stream eagerly so format/seekability errors (e.g. a
         # non-seekable unix-compress source, which the codec rejects via the translator)
@@ -135,7 +138,18 @@ class SingleFileReader(BaseArchiveReader):
         )
 
     def _peek_header(self, length: int) -> bytes:
-        """The first ``length`` bytes of the compressed stream, without consuming the source."""
+        """The first ``length`` bytes of the compressed source, read once and cached.
+
+        The first call (or one needing more than is cached) reads the source a single time;
+        later calls serve from the cache without re-opening or re-seeking. For a non-seekable
+        source this reuses the prefix detection already buffered in the ``PeekableStream``; for
+        a path it opens a fresh handle; for a seekable stream it reads and rewinds once.
+        """
+        if self._header_cache is None or len(self._header_cache) < length:
+            self._header_cache = self._read_source_prefix(length)
+        return self._header_cache[:length]
+
+    def _read_source_prefix(self, length: int) -> bytes:
         from archivey.internal.streams.peekable import PeekableStream
 
         src = self._source
--- a/src/archivey/formats/single_file_reader.py
+++ b/src/archivey/formats/single_file_reader.py
@@ -16,9 +16,8 @@ Phase 2); only their *seekable-decompressor* refinements remain for Phase 8.
 from __future__ import annotations
 
 import os
-from datetime import datetime, timezone
 from pathlib import Path
-from typing import BinaryIO, Callable, ClassVar, Iterator, Mapping
+from typing import BinaryIO, Iterator
 
 from archivey.internal.config import StreamConfig
 from archivey.internal.cost import (
@@ -34,9 +33,11 @@ from archivey.internal.errors import (
 from archivey.internal.reader import BaseArchiveReader, ReadBackend
 from archivey.internal.registry import register_reader
 from archivey.internal.streams.codecs import (
-    codec_for_stream_format,
+    SINGLE_FILE_CODECS,
+    MetadataContext,
     open_codec_stream,
     resolve_codec,
+    stream_codec_for_format,
 )
 from archivey.internal.streams.decompressor_stream import DecompressorStream
 from archivey.internal.streams.streamtools import is_seekable, is_stream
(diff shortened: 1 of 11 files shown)


src/archivey/internal/streams/codecs.py
```diff
@@ -798,31 +764,374 @@ def _open_deflate64(source: CodecSource, params: CodecParams, config: StreamConf
...
+    """(format, probe) pairs for standalone formats recognized by a content probe.
+
+    These are the formats not identified by exact magic (Brotli, which has no signature;
+    zlib, whose 2-byte header the probe gates on before decoding).
+    """
+    return tuple(
+        (d.single_file_format, d.content_probe)
+        for d in single_file_descriptors()
+        if d.content_probe is not None and d.single_file_format is not None
+    )
```
davitf: I think all these little functions are confusing. Would it be cleaner if the SingleFileReader and/or detector got the list of stream codecs directly (make it not-private) and accessed the fields? e.g. you don't need single_file_reader to retrieve and keep its own copy of these lists/tuples, just iterate over the codecs as needed.

Don't blindly accept these, let's think about what the cleanest structure would be

src/archivey/internal/streams/codecs.py
```diff
@@ -798,31 +764,374 @@ def _open_deflate64(source: CodecSource, params: CodecParams, config: StreamConf
...
+def _probe_zlib(prefix: bytes) -> bool:
+    """Recognize a zlib stream: a known CMF/FLG header (fail-fast) that then decodes."""
+    return prefix[:2] in _ZLIB_HEADERS and _decodes_as(Codec.ZLIB, prefix)
+
+
+# --- the codec descriptor + registry ---------------------------------------------------
+
+
+@dataclass(frozen=True)
+class StreamCodec:
```
davitf: Maybe instead of defining the StreamCodec as a dataclass built in the list, they could be actual classes, in which each one defines the relevant fields / methods. it will help encapsulation (all the gzip functions will be inside the class, instead of being top-level functions and then referred to in the list). what do you think?

src/archivey/formats/single_file_reader.py
```diff
@@ -146,67 +120,32 @@ def _build_member(self, archive_name: str | None) -> ArchiveMember:
...
+
+    def _peek_header(self, length: int) -> bytes:
+        """The first ``length`` bytes of the compressed stream, without consuming the source."""
         from archivey.internal.streams.peekable import PeekableStream
 
         src = self._source
         if isinstance(src, Path):
             with open(src, "rb") as f:
-                return f.read(_HEADER_PEEK)
+                return f.read(length)
```
davitf, line 158: This re-opens the file and re-reads the header every time someone calls peek_header. Should we cache the result?

src/archivey/formats/single_file_reader.py
```diff
@@ -146,67 +120,32 @@ def _build_member(self, archive_name: str | None) -> ArchiveMember:
...
         if isinstance(src, PeekableStream):
-            return src.peek(_HEADER_PEEK)
+            return src.peek(length)
         if is_seekable(src):
-            data = src.read(_HEADER_PEEK)
+            data = src.read(length)
             src.seek(0)
```
davitf, line 163: Also here, if the stream itself is slow-seek (internally needing to reopen itself), this is sub-optimal. I'm wondering if the reader should do sort of what we were doing before, reading the header eagerly and providing it to the codecs inside MetadataContext itself, or at least buffering the first ~512 bytes and having _peek_header return from that unless the request is larger (and then maybe increase the buffer with any newly-read data, up to a limit).

In a way that's what PeekableStream already does, so maybe we should use that for everything? (not sure about this part, please weigh pros and cons)

src/archivey/internal/streams/codecs.py
```diff
@@ -735,94 +496,678 @@ def close(self) -> None:
...
+    ``open`` + ``translate``.
+    """
+
+    codec: ClassVar[Codec]
+    # The single-file/TAR StreamFormat this codec decodes, when it is a stream format at all
+    # (raw container coders such as DEFLATE/LZMA have none).
+    stream_format: ClassVar[StreamFormat | None] = None
+    # The standalone single-file ArchiveFormat (RAW_STREAM + stream_format) when this codec is
+    # presented as a bare one-member archive; ``None`` for container-only codecs.
+    single_file_format: ClassVar[ArchiveFormat | None] = None
```
davitf: Do we need this? Can't we build it on demand as `ArchiveFormat(ContainerFormat.RAW_STREAM, stream_format)`?

src/archivey/formats/single_file_reader.py
```diff
@@ -146,67 +120,32 @@ def _build_member(self, archive_name: str | None) -> ArchiveMember:
...
 
-    def _peek_header(self) -> bytes:
-        """The first bytes of the compressed stream, without consuming the source."""
+    def _metadata_context(self) -> MetadataContext:
+        return MetadataContext(
+            peek_header=self._peek_header,
+            probe_decompressed_size=self._probe_decompressed_size,
+        )
+
+    def _peek_header(self, length: int) -> bytes:
```
davitf, line 140: This seems sort of redundant with detection._peek_prefix. Why do we even need peeking here? can' t we leave all the detection to that module, and only use the result here?

src/archivey/internal/streams/codecs.py
```diff
@@ -735,94 +496,678 @@ def close(self) -> None:
...
+    # Exact magic signals for the standalone format, aggregated by the detector.
+    magic: ClassVar[tuple[MagicSignature, ...]] = ()
+    # Standalone file extensions, for member-name inference and extension detection.
+    extensions: ClassVar[tuple[str, ...]] = ()
```
davitf: This is sort of redundant with `ArchiveFormat.file_extension()`, which seems to be what detection.py actually uses. However, I see here that this supports multiple extensions per codec (so we can have .zst and .zstd), but that's probably handled better at the StreamFormat / ArchiveFormat level (maybe a dict of extra extensions to formats)

PR #17, opened 2026-06-25
refactor(streams): unify read-only stream wrappers behind a shared base

davitf:
Implements OpenSpec change `stream-wrapper-base`. Introduces two small internal bases so the ~10 read-only stream wrappers stop re-implementing the `BinaryIO` surface, and folds the rewind-warning wrapper into `ArchiveStream`.

## What changed

**New bases** in `internal/streams/streamtools/base.py`:
- `ReadOnlyIOStream(io.RawIOBase, BinaryIO)` — the read-only surface (`readable→True`, `writable→False`, `write` raises) + one canonical `readinto`/`readall` built on the subclass's `read`. `read` is left abstract (`@abc.abstractmethod` + a `NotImplementedError` body — the body is the real runtime guard, since `io.RawIOBase`'s C `__new__` ignores `__abstractmethods__`).
- `DelegatingStream(ReadOnlyIOStream)` — holds one inner `BinaryIO` and forwards `read/seek/tell/seekable/close`. `seekable()` goes through `is_seekable()`. `readinto` is a zero-copy passthrough by default, with an explicit `readinto_passthrough=False` flag for subclasses whose `read` has a side effect (byte tracking, an EOF check) that must not be bypassed.

**Migration (behavior-preserving):**
- On `DelegatingStream`: `_AcceleratorStream` (close-guard only), `_GzipTruncationCheckStream` (`readinto_passthrough=False`), `_ZstdReopenStream`.
- On `ReadOnlyIOStream` only: `SlicingStream` (transforms offsets + non-owning view), `PeekableStream`, `VerifyingStream` (sequential), `DecompressorStream`, `ArchiveStream` (per-call exception translation must wrap every call).
- `BinaryIOWrapper` kept as a distinct class (its inner is a duck-typed, possibly non-`BinaryIO` object — a different contract).
- `_SlowSeekWarningStream` **removed** — the rewind-is-slow warning is now carried by `ArchiveStream` (sourced from the codec descriptor via `RewindWarning`), since the public/codec-stream surface is always an `ArchiveStream`. The accelerator close-guard deliberately stays at the object's birth (`backend.open()` can hand out an accelerator with no `ArchiveStream` around it).

## Result
- Net **−114 lines** in `src/`.
- Behavior-preserving; gated by the existing stream tests. Full suite: **483 passed / 7 skipped**; ruff / pyrefly / ty clean.

Also adds an `IDEAS.md` entry (rapidgzip for zlib / raw-deflate streams, for the future native ZIP parser).


src/archivey/internal/streams/streamtools/base.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import TYPE_CHECKING, Any, BinaryIO
+
+if TYPE_CHECKING:
+    from _typeshed import WriteableBuffer
+
+
+class ReadOnlyIOStream(io.RawIOBase, BinaryIO):
+    """Base for a read-only ``BinaryIO``: subclasses implement ``read`` (and what they change)."""
+
+    def read(self, n: int = -1, /) -> bytes:
```
davitf, line 44: worth annotating with @abstractmethod?

src/archivey/internal/streams/streamtools/base.py
```diff
@@ -0,0 +1,95 @@
...
+
+import io
+from typing import TYPE_CHECKING, Any, BinaryIO
+
+if TYPE_CHECKING:
+    from _typeshed import WriteableBuffer
+
+
+class ReadOnlyIOStream(io.RawIOBase, BinaryIO):
+    """Base for a read-only ``BinaryIO``: subclasses implement ``read`` (and what they change)."""
```
davitf: worth explaining why this doesn't have seekable/seek/tell?

src/archivey/internal/streams/streamtools/base.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def write(self, b: Any, /) -> int:
+        raise io.UnsupportedOperation("write")
+
+
+class DelegatingStream(ReadOnlyIOStream):
+    """A read-only wrapper around one inner ``BinaryIO``, forwarding to it by default.
+
+    Subclasses override only the method whose behavior they change (e.g. just ``seek`` to add a
+    warning, or just ``close`` to add a cleanup guard).
```
davitf, line 81: Worth reminding that if any implementation needs to override read(), they also need to override readinto() so that direct calls to it don't skip the read() implementation? or should we add a mechanism here for these cases, e.g. a constructor arg that subclasses can pass, to force readinto to go through read() instead of using the passthrough? (or even a way for us to detect automatically if read() has been overridden() -- though I'm not sure if that would be a safer approach, or may end up leading to confusing bugs)

src/archivey/internal/streams/streamtools/base.py
```diff
@@ -0,0 +1,95 @@
...
+        return super().readinto(b)
+
+    def seek(self, offset: int, whence: int = io.SEEK_SET, /) -> int:
+        return self._inner.seek(offset, whence)
+
+    def tell(self, /) -> int:
+        return self._inner.tell()
+
+    def seekable(self) -> bool:
+        return self._inner.seekable()
```
davitf: would it be worth using is_seekable() here, to automatically fix the edge cases that it handles?

src/archivey/internal/streams/streamtools/slice.py
```diff
@@ -8,15 +8,13 @@
...
 
+from archivey.internal.streams.streamtools.base import ReadOnlyIOStream
 from archivey.internal.streams.streamtools.binaryio import is_seekable
 
-if TYPE_CHECKING:
-    from _typeshed import WriteableBuffer
 
-
-class SlicingStream(io.RawIOBase, BinaryIO):
+class SlicingStream(ReadOnlyIOStream):
```
davitf, line 17: Couldn't this be a subclass of DelegatingStream? or would it be dangerous? If it can't, it may be worth explaining why in a comment

src/archivey/internal/streams/codecs.py
```diff
@@ -345,14 +255,13 @@ def read(self, size: int = -1, /) -> bytes:
         return data
 
     def readinto(self, b: "WriteableBuffer", /) -> int:
-        mv = memoryview(b).cast("B")
-        data = self.read(len(mv))
-        mv[: len(data)] = data
-        return len(data)
+        # Route through this class's read() (not DelegatingStream's zero-copy passthrough) so the
+        # byte-total tracking and the EOF truncation check still run on readinto-driven reads.
+        return ReadOnlyIOStream.readinto(self, b)
```
davitf: If we add that mechanism to DelegatingStream, this one could just use it

src/archivey/internal/streams/streamtools/base.py
davitf: Added `@abc.abstractmethod` as the intent/static-check signal. Heads-up from testing it, though: it does **not** enforce instantiation here — `io.RawIOBase`'s C-level `__new__` ignores `__abstractmethods__` (I verified: `__abstractmethods__` becomes `{'read'}` but a subclass without `read()` still constructs). So the `NotImplementedError` body remains the actual runtime guard; I noted that in a comment so it's not mistaken for real enforcement.

---

src/archivey/internal/streams/streamtools/base.py
davitf: Added to the `ReadOnlyIOStream` docstring: it deliberately defines only the read-only surface and **not** `seek`/`tell`/`seekable`/`close`, because those vary per wrapper (sequential vs. seekable; owns-its-inner vs. a non-owning view) — subclasses declare them or inherit `RawIOBase`'s non-seekable defaults, and `DelegatingStream` supplies the forwarding versions for wrappers that own one inner stream.

---

src/archivey/internal/streams/streamtools/base.py
davitf: Good catch — this was a real footgun. The default zero-copy `readinto` passthrough bypasses this class's `read`, so a subclass that overrides `read` with a side effect (byte tracking, hashing, an EOF check) would silently skip it on `readinto`-driven reads.

Fixed by adding an explicit `readinto_passthrough` flag to `DelegatingStream.__init__` (default `True`). When `False`, `readinto` routes through `super().readinto` (the `read`-based `ReadOnlyIOStream` impl) so the overridden `read` always runs. I made it an explicit flag rather than auto-detecting an overridden `read`, because a plain pass-through `read` override should keep the zero-copy path — silent auto-detection would make that choice invisible and bug-prone. Documented in the class docstring, and `_GzipTruncationCheckStream` now passes `readinto_passthrough=False` (see the codecs.py thread).

---

src/archivey/internal/streams/streamtools/base.py
davitf: Done — `DelegatingStream.seekable()` now delegates to `is_seekable(self._inner)` instead of calling `self._inner.seekable()` directly. That picks up the edge cases the bare call misses (a `BufferedReader` over a non-seekable raw; a pipe that reports `seekable()=True` but cannot actually reposition).

---

src/archivey/internal/streams/streamtools/slice.py
davitf: Considered it but kept it as a `ReadOnlyIOStream`. Two reasons, now noted in the class docstring: (1) every operation is *transformed* not forwarded — `read` clamps to the slice bounds and `seek`/`tell` are relative to the slice start, not the underlying offset — so almost everything `DelegatingStream` provides would be overridden anyway; and (2) more importantly, it's a *non-owning view*: the container owns the underlying stream, so `SlicingStream` must NOT close it, whereas `DelegatingStream.close` closes its inner. So delegation would be both useless and unsafe here.

---

src/archivey/internal/streams/codecs.py
davitf: Updated — `_GzipTruncationCheckStream` now calls `super().__init__(inner, readinto_passthrough=False)` and drops its manual `readinto` override entirely. The flag routes `readinto` through `read`, so the byte-total tracking (the truncation check) runs on every read path, which is exactly what the old hand-written `readinto` was doing. Net: one fewer override and the consistency guarantee is now enforced by the base. (See the base.py thread for the flag's rationale.)

---

PR #18, opened 2026-06-30
Phase 3: TAR and ISO 9660 random-access readers + detection

davitf:
## Summary

Implements Phase 3 of the archive-reading roadmap: random-access reading of TAR (plain and compressed) and ISO 9660 images, with full detection support including inner-TAR probing over single-file compressors and ISO extended-window magic detection.

### TAR Reader (`src/archivey/formats/tar_reader.py`)
- **Random-access TAR reading** via stdlib `tarfile`, supporting both plain and compressed archives (`.tar`, `.tar.gz`, `.tar.bz2`, `.tar.xz`, etc.)
- **Codec composition**: compressed tars are decompressed through the codec layer (not `tarfile`'s native modes), enabling uniform support for all codecs archivey knows
- **Member metadata mapping**: PAX/GNU/ustar format support with proper timestamp handling (sub-second precision via PAX headers, tz-aware UTC conversion)
- **Cost model**: plain TAR reports `REQUIRES_SCANNING` listing cost and `DIRECT` access; compressed TAR reports `REQUIRES_DECOMPRESSION` and `SOLID` access
- **Error translation**: `tarfile.ReadError` and `EOFError` properly mapped to `TruncatedError` for truncated archives

### ISO 9660 Reader (`src/archivey/formats/iso_reader.py`)
- **Namespace auto-selection**: Rock Ridge > Joliet > plain ISO 9660, with choice recorded in `ArchiveInfo.extra["iso.namespace"]`
- **Metadata fidelity**: Rock Ridge carries POSIX mode/uid/gid and symlinks; Joliet and plain carry none (fields are `None`)
- **O(1) listing**: directory tree lives in header region, enabling `INDEXED` listing cost and `DIRECT` random access
- **Symlink and hardlink support**: proper member type classification and link target extraction
- **Timestamp handling**: Rock Ridge TF entries provide precise POSIX times (access/creation/modification)
- **Non-seekable rejection**: `REQUIRES_SEEK` enforced; password rejection for unsupported encryption
- **PyCdlibStream adapter**: wraps `pycdlib`'s `PyCdlibIO` to properly signal EOF and integrate with the reader interface

### Detection Enhancements (`src/archivey/internal/detection.py`)
- **Inner-TAR probe** over single-file compressors: decompresses ≥512 bytes to check for `ustar` magic at offset 257, upgrading bare compressor matches (e.g., `GZ`) to TAR combos (e.g., `TAR_GZ`) with `PROBABLE` confidence
- **Deferred probing**: when codec backend is absent, detection reports the bare compressor and defers inner-TAR determination to open time (no warning on benign tar.gz/gz mismatch)
- **ISO extended-window detection**: reads up to 32 KiB to detect ISO 9660 `CD001` signature at offset 32,769 (fallback when magic-based detection fails)
- **Conflict warning suppression**: benign mismatches (e.g., extension suggests `TAR_GZ` but magic says `GZ` due to deferred probe) do not emit warnings

### Data Model Updates
- **`ArchiveInfo.extra` field**: new `dict[str, Any]` field (default empty, excluded from equality) for format-specific archive-level metadata (e.g., ISO namespace selection)
- Mirrors the existing `ArchiveMember.extra` pattern for consistency

### Tests
- **TAR tests** (`tests/test_tar.py`): random-access read, compressed-tar cost/format properties, member metadata mapping (ustar/GNU/PAX), symlink/hardlink handling, truncation/corruption detection
- **ISO tests** (`tests/test_iso.py`): namespace auto-select and metadata fidelity, cost properties, read operations, rejection of passwords/writes/non-seekable sources, corruption handling, registry degradation (NONE support when pycdlib absent)
- **Detection tests** (`tests/test_detection.py`): inner-TAR probe over gzip

https://claude.ai/code/session_01NaewS57iiRuQFZcF5Hu5Zg


src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,395 @@
...
+    CompressionMethod,
+    MagicSignature,
+    MemberType,
+)
+
+# Errors raised while opening/reading an image: pycdlib's own exception plus OS errors from
+# the underlying handle. Built defensively so the module imports even without pycdlib.
+_ISO_ERRORS: tuple[type[Exception], ...] = (
+    (pycdlib.pycdlibexception.PyCdlibException, OSError) if pycdlib is not None else (OSError,)
+)
```
davitf: if the OSErrors are actual issues when reading the archive from the FS (e.g. file not found, physical media errors, permission errors), we should let them pass through untranslated, since they're not format-related. I believe we have that in the specs somewhere

src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,395 @@
...
+        return target.decode("utf-8", errors="surrogateescape") if target else None
+
+    # --- data ---------------------------------------------------------------------------
+
+    def _open_member(self, member: ArchiveMember) -> BinaryIO:
+        ns_path = member._raw
+        assert isinstance(ns_path, str), "ISO member is missing its namespace path"
+        try:
+            raw = self._iso.open_file_from_iso(**{self._path_kw: ns_path})
+            raw.__enter__()  # sets up the read offset; _PyCdlibStream.close() exits it
```
davitf: Could we move this __enter__() to inside the PyCdlibStream __init__ for clarity, to keep the context entering/exiting in the same logical place? we could move the PyCdlibStream(raw) construction inside this try/except, so any __enter__-time exceptions are handled

src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,395 @@
...
+            date.hour,
+            date.minute,
+            date.second,
+            tzinfo=tz,
+        )
+    except (ValueError, AttributeError, TypeError, OverflowError):
+        return None
+
+
+class _PyCdlibStream(io.RawIOBase):
```
davitf: could this be a DelegatingStream (see src/archivey/internal/streams/streamtools/base.py)?

src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,395 @@
...
+            self._path_kw = "iso_path"
+
+    def _open_error(self, exc: Exception) -> ArchiveyError:
+        translated = self._translate_exception(exc)
+        if translated is not None:
+            self._stamp_error_context(translated)
+            return translated
+        err = CorruptionError(f"Could not open ISO image: {exc!r}")
+        self._stamp_error_context(err)
+        return err
```
davitf: this is for OSError, right? if we remove wrapping for those, be sure to clean up this method

src/archivey/formats/iso_reader.py
davitf: Agreed — that's `error-handling`'s "Genuine runtime and I/O errors are not reclassified". Fixed in 0b83f89 (and the earlier b3387f9): the ISO backend now catches only `pycdlib.PyCdlibException`, so a filesystem `OSError` (missing file, permissions, media error) propagates unchanged. Same narrowing applied to the TAR open path. Regression tests assert a missing-file open raises `FileNotFoundError`, not `CorruptionError`.

---

src/archivey/formats/iso_reader.py
davitf: Done — `_PyCdlibStream.__init__` now enters the context (paired with the exit in `close()`), and the construction moved inside `_open_member`'s try/except so any enter-time `PyCdlibException` is translated. (See also 0b83f89, which reworks this onto `DelegatingStream` per your other comment.)

---

src/archivey/formats/iso_reader.py
davitf: Done in 0b83f89 — `_PyCdlibStream` now subclasses `DelegatingStream`. `PyCdlibIO.close()` is exactly its `__exit__`, so read/seek/tell/seekable/close are inherited delegation; I pass `readinto_passthrough=False` to keep the EOF workaround (route `readinto` through `read()`, which clamps to the logical length).

---

src/archivey/formats/iso_reader.py
davitf: Right. Cleaned up in 0b83f89 — `_open_error` is removed entirely. The open path now catches only `PyCdlibException` and uses the same translate/stamp/propagate shape as the listing and member-open paths, so there's no dead `OSError`→`CorruptionError` fallback left.

---

src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,369 @@
...
+    ``PyCdlibIO`` must be *entered* before use — its context manager sets up the read offset —
+    so this enters it in ``__init__`` and relies on ``DelegatingStream.close()`` (which calls
+    ``inner.close()``, the exact equivalent of ``PyCdlibIO.__exit__``) to exit it, keeping the
+    enter/exit lifecycle paired. ``readinto_passthrough=False`` routes ``readinto`` through
+    ``read``: ``PyCdlibIO.readinto`` mis-signals EOF inside a ``BufferedReader`` (it reads into
+    the ISO sector padding past the file's logical end), while ``PyCdlibIO.read`` clamps to the
+    logical length. Read/seek/tell/seekable are inherited delegation.
+    """
+
+    def __init__(self, raw: Any) -> None:
```
davitf: can we define the type properly here?
```suggestion
    def __init__(self, raw: PyCdlibIO) -> None:
```

src/archivey/formats/tar_reader.py
```diff
@@ -0,0 +1,357 @@
...
+        ".tar.xz": ArchiveFormat.TAR_XZ,
+        ".txz": ArchiveFormat.TAR_XZ,
+        ".tar.zst": ArchiveFormat.TAR_ZST,
+        ".tzst": ArchiveFormat.TAR_ZST,
+        ".tar.lz4": ArchiveFormat.TAR_LZ4,
+        ".tar.lz": ArchiveFormat(ContainerFormat.TAR, StreamFormat.LZIP),
+        ".tlz": ArchiveFormat(ContainerFormat.TAR, StreamFormat.LZIP),
+        ".tar.br": ArchiveFormat(ContainerFormat.TAR, StreamFormat.BROTLI),
+        ".tar.Z": ArchiveFormat(ContainerFormat.TAR, StreamFormat.UNIX_COMPRESS),
+    }
```
davitf: Could we define this based on FORMATS or _TAR_COMPRESSED (take the extensions from the ArchiveFormat) , and just add the extra short extensions such as .tgz?

src/archivey/formats/iso_reader.py
```diff
@@ -0,0 +1,369 @@
...
+class IsoReadBackend(ReadBackend):
+    """Backend factory for ISO 9660 images (requires the ``[iso]`` extra → ``pycdlib``)."""
+
+    FORMATS: tuple[ArchiveFormat, ...] = (ArchiveFormat.ISO,)
+    EXTENSIONS: Mapping[str, ArchiveFormat] = {".iso": ArchiveFormat.ISO}
+    # The primary volume descriptor's "CD001" magic sits at offset 32 769; detection peeks the
+    # extended 32 774-byte window on demand to find it (see internal/detection.py).
+    MAGIC: tuple[MagicSignature, ...] = (
+        MagicSignature(32769, b"CD001", ArchiveFormat.ISO),
+    )
```
davitf, line 356: archivey-dev has:

# Taken from the pycdlib code
_ISO_MAGIC_BYTES = [
    b"CD001",
    b"CDW02",
    b"BEA01",
    b"NSR02",
    b"NSR03",
    b"TEA01",
    b"BOOT2",
]

can you check and add if needed?

src/archivey/formats/iso_reader.py
davitf: You're right, and I had the cause wrong. The original lint failure wasn't inherent to a `TYPE_CHECKING` import — it was that pycdlib lived **only** in the `[iso]` extra, not the dev group (unlike zstandard/lz4/brotli, which are in dev), so the lint env (`uv sync --group dev`) couldn't resolve it. Fixed in 9306c84: added pycdlib to the dev group, imported `PyCdlibIO` under `TYPE_CHECKING`, and annotated `raw: PyCdlibIO`. One wrinkle — `PyCdlibIO` is an `io.RawIOBase` (a `BinaryIO` at runtime but not by typeshed's nominal hierarchy), so I cast at the `DelegatingStream` boundary. The runtime `_optional()` resolution stays for the core-only/no-`[iso]` case (that leg installs `--no-dev`, so pycdlib is still absent and degradation is unchanged). Bonus: the ISO tests now run in the default dev env.

---

src/archivey/formats/tar_reader.py
davitf: Done in 9306c84 — `EXTENSIONS` is now derived from the formats' `file_extension()` (`TAR → .tar`, `TAR_GZ → .tar.gz`, `(TAR, LZIP) → .tar.lz`, …), with only the short aliases (`.tgz`/`.tbz2`/`.tbz`/`.txz`/`.tzst`/`.tlz`) listed by hand. Built at module scope since a dict comprehension in the class body can't see class-level names. (It now also picks up `.tar.zz` for TAR-over-zlib.)

---

src/archivey/formats/iso_reader.py
davitf: Checked these — they're the UDF/boot volume-recognition descriptors (`BEA01`/`NSR02`/`NSR03`/`TEA01` are ECMA-167/UDF, `BOOT2` a boot record, `CDW02` a CD-i variant). The conclusion is that `CD001` alone is correct for our magic at offset 32769:

- The identifier at **offset 32769** is `CD001` for every image pycdlib can open — including UDF-bridge discs. I verified on a pycdlib-built UDF image: `CD001` sits at 32769, while the UDF descriptors (`BEA01`/`NSR02`/`TEA01`) live at later sectors (offsets 36865/38913/40961, i.e. sector 18+), not 32769. DEV's list was for a multi-sector scan, not a single fixed offset.
- pycdlib **requires** `CD001` at 32769: clobbering it to `BEA01` makes `open_fp` raise `PyCdlibInvalidISO: invalid CD isoIdentification`. So a pure-UDF image (no ISO 9660 structure) can't be opened by this backend anyway, and our reader has no UDF namespace.

So adding those identifiers at 32769 would only match images pycdlib can't read (detect-then-fail). `CD001` already covers everything in scope, so I left the magic as-is. If/when a native UDF reader lands, the multi-sector recognition scan would come with it.

---

PR #23, opened 2026-06-30
openspec: refine Phase 4 proposals after implementor review

davitf:
## Summary

Implementor-hat review of the two Phase 4 OpenSpec proposals in #21
(`phase-4-tar-streaming` + `phase-4-safe-extraction`), with fixes for conflicting,
missing, and unclear information. Targets the #21 branch so the diff is just the review
refinements. Docs/OpenSpec artifacts only — no runtime code changes.

### Conflicts resolved
- **TAR hardlink conflict** between `format-tar` ("defer to a post-pass") and
  `safe-extraction` ("source precedes link, no pending state"). Added a `format-tar`
  **MODIFIED** delta to `phase-4-safe-extraction` reconciling them: the real file always
  precedes its hardlinks, so resolution is a single forward pass. For solid/compressed
  `.tar.gz` the excluded-but-needed source is **staged as the forward pass reaches it** —
  no seek-back, no re-decompression, no deferred second pass — via a bounded
  `{source → link path}` map. Streaming (no upfront list) raises a clear `ExtractionError`
  when a selected link's source was filtered out. This also answers the review question
  about solid-tar performance without un-banning the DEV `pending_*` machine.

### Risky / underspecified
- **`REQUIRES_SEEK` relaxation** is now per-backend opt-in
  (`SUPPORTS_STREAMING_NON_SEEKABLE`, set only on `TarReadBackend`); the opener-wide
  `and not streaming` option is dropped because it would wrongly relax ZIP/ISO, which must
  keep failing fast on non-seekable sources.
- **Ordering**: `phase-4-tar-streaming` lands **first**; `phase-4-safe-extraction` consumes
  its `_iter_with_data()` override and `compressed_source_size` hook (the latter feeds the
  archive-wide bomb ratio, so safe extraction depends on it even for seekable `.tar.gz`).
  PLAN.md, both proposals, and the tasks updated.

### Clarifications
- `BombTracker` constructor gains `compressed_source_size: int | None` in the
  `safe-extraction` delta; archive-wide ratio activates on **cumulative** output
  (`_total_bytes`), not per-member.
- Stale module paths fixed (`formats/` → `internal/backends/`, `open_archive` → `core.py`,
  `reader` → `base_reader.py`) post package-layout restructure; noted that
  `StreamConfig(streaming=…)` is already wired in `tar_reader.py`.
- Removed references to the non-existent single-member `extract()` from
  `access-mode-and-cost` and `archive-reading` (the design uses `extract_all(members=…)`).
- Reflowed two requirement statements so `SHALL` is on the first line —
  `openspec validate --strict` now passes for both changes (it previously failed, including
  on a requirement #21 didn't touch).

## Test plan
- [x] `openspec validate phase-4-tar-streaming --strict` → valid
- [x] `openspec validate phase-4-safe-extraction --strict` → valid
- [x] `openspec validate archive-reading --strict --type spec` → valid
- [ ] Maintainer review of the hardlink forward-staging decision and the per-backend seek gating

---
_Generated by [Claude Code](https://claude.ai/code/session_016pts3QCc3rvScZuBVbx9mV)_

### DIFF
--- a/openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
+++ b/openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
@@ -1,0 +1,59 @@
+# TAR Format Behavior — delta (phase-4-safe-extraction)
+
+## MODIFIED Requirements
+
+### Requirement: Handle TAR hardlinks via two-pass extraction with cross-device fallback
+
+The system SHALL support hardlink extraction from TAR archives. The `linkname` field holds
+the target path. Resolution is performed by the `safe-extraction` `ExtractionCoordinator` (see
+the *Hardlink Two-Pass Extraction* requirement there), not by a separate post-pass in the TAR
+backend, and MUST NOT introduce `pending_*` link-resolution state.
+
+TAR ordering guarantees the real file (the source) precedes any hardlink that references it.
+The coordinator therefore resolves hardlinks within its single forward pass, **without
+seeking back or re-decompressing** (important for solid `.tar.gz`, where reaching an earlier
+member would require re-inflating the stream from the start):
+
+1. If the hardlink's source has already been extracted (or staged) earlier in this pass,
+   create a filesystem hardlink via `os.link()`.
+2. **Streaming mode** (forward-only, no upfront member list): if the source was filtered out
+   by the `members` selector or `filter`, the bytes are unrecoverable in a single pass —
+   raise an explicit `ExtractionError` with a clear message.
+3. **Random-access mode**: the pre-pass builds a hardlink closure map from member metadata
+   (for TAR the `linkname` lives in the header, so this reads no member payload). When the
+   forward pass reaches an **excluded-but-needed** source, the coordinator writes its content
+   to the first selected link's destination path **at that point** (the source precedes its
+   links, so its bytes are already streaming past), and `os.link()`s any further selected
+   links to it when they are reached. The excluded source is never created at its own
+   destination path. An implementation MAY instead stage the content in a hidden temp inside
+   `dest`. No seek-back, no re-decompression, and no deferred post-pass; the only state is a
+   bounded `{source → link path}` map.
+4. If `os.link()` fails with a cross-device link error, fall back to copying via
+   `shutil.copy2`.
+
+#### Scenario: Hardlink source already extracted
+
+- **WHEN** a `HARDLINK` member is encountered during extraction
+- **AND** its source has already been written to disk in this pass
+- **THEN** the system creates a filesystem hardlink from the source path to the new path via `os.link` (or `shutil.copy2` on cross-device failure)
+
+#### Scenario: Hardlink source filtered out in streaming mode
+
+- **WHEN** a selected `HARDLINK` member is encountered in streaming mode
+- **AND** its source was excluded by the `members` selector or `filter`
+- **THEN** an explicit `ExtractionError` is raised (a single forward pass cannot recover the source bytes)
+
+#### Scenario: Hardlink source filtered out in random-access mode
+
+- **WHEN** a selected `HARDLINK` member's source was excluded by the `members` selector or `filter`
+- **THEN** the source content is written to the first selected link's path as the forward pass reaches the source (further selected links are `os.link`'d to it), the excluded source is never created at its own path, and no seek-back, re-decompression, or deferred post-pass is used
+
+#### Scenario: Solid compressed tar resolves hardlinks without re-decompression
+
+- **WHEN** a `.tar.gz` (solid, no seek accelerator) is extracted and a selected hardlink's source was excluded
+- **THEN** the source is staged during the single forward decompression pass and the stream is never re-decompressed from the start to recover it
+
+#### Scenario: Cross-device hardlink falls back to copy
+
+- **WHEN** `os.link` fails with a cross-device error
+- **THEN** the system copies the source file to the link destination instead
--- a/openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
+++ b/openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
@@ -12,39 +12,40 @@ calling `reader.get_members_if_available()`, and iterating the forward pass — 
 among the algorithms below. It MUST NOT hold a push-model deferred-state machine, and MUST NOT
 force an upfront pass the run does not need.
 
-**Obtaining the member list.** Only a `members` selector or a `filter` can orphan a hardlink
-(select a link while excluding its source); an unfiltered extract-all never orphans one. Even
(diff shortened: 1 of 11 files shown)


openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,59 @@
...
+TAR ordering guarantees the real file (the source) precedes any hardlink that references it.
+The coordinator therefore resolves hardlinks within its single forward pass, **without
+seeking back or re-decompressing** (important for solid `.tar.gz`, where reaching an earlier
+member would require re-inflating the stream from the start):
+
+1. If the hardlink's source has already been extracted (or staged) earlier in this pass,
+   create a filesystem hardlink via `os.link()`.
+2. **Streaming mode** (forward-only, no upfront member list): if the source was filtered out
+   by the `members` selector or `filter`, the bytes are unrecoverable in a single pass —
+   raise an explicit `ExtractionError` with a clear message.
```
davitf: actually, I think we planned for how to handle extraction errors somewhere. it could abort immediately, or continue and just report any errors at the end. we should handle these errors the same way

openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,59 @@
...
+3. **Random-access mode**: the pre-pass builds a hardlink closure map from member metadata
+   (for TAR the `linkname` lives in the header, so this reads no member payload). When the
```
davitf: whether the linkname is in the header or data doesn't really matter, because the header is not in a central directory. I think the real question is: should we make a pre-pass when extracting tars, since that can be expensive? it can be justifiable if it gives the extractor all the info it needs for the best extraction. but, if the tar didn't actually need it and could be extracted sequentially with no errors, we're wasting a pass. 

a second pass is only actually useful in the likely rare case of a slow-seekable non-indexed source with a filter function. maybe it's worth implementing two (or three) algorithms?

openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,59 @@
...
+   forward pass reaches an **excluded-but-needed** source, the coordinator writes its content
+   to the first selected link's destination path **at that point** (the source precedes its
+   links, so its bytes are already streaming past), and `os.link()`s any further selected
+   links to it when they are reached. The excluded source is never created at its own
+   destination path. An implementation MAY instead stage the content in a hidden temp inside
+   `dest`. No seek-back, no re-decompression, and no deferred post-pass; the only state is a
+   bounded `{source → link path}` map.
+4. If `os.link()` fails with a cross-device link error, fall back to copying via
+   `shutil.copy2`.
```
davitf: but also, since there may be multiple hardlinks to the same file, it may be worth linking between them. e.g. if B links to A but on a different device, we need to copy. but then if we get another link C -> A that will be written to the same device as B, we can't link C to A but could link C to B. worth addressing? how does tarfile handle these linking cases?

openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,59 @@
...
+
+#### Scenario: Hardlink source filtered out in random-access mode
+
+- **WHEN** a selected `HARDLINK` member's source was excluded by the `members` selector or `filter`
+- **THEN** the source content is written to the first selected link's path as the forward pass reaches the source (further selected links are `os.link`'d to it), the excluded source is never created at its own path, and no seek-back, re-decompression, or deferred post-pass is used
+
+#### Scenario: Solid compressed tar resolves hardlinks without re-decompression
+
+- **WHEN** a `.tar.gz` (solid, no seek accelerator) is extracted and a selected hardlink's source was excluded
+- **THEN** the source is staged during the single forward decompression pass and the stream is never re-decompressed from the start to recover it
```
davitf: I didn't understand this. What is "staged" here? if the stream is never re-decompressed, when will the source actually be decompressed?

openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,98 @@
+# TAR Format Behavior — delta (phase-4-safe-extraction)
+
+## MODIFIED Requirements
+
+### Requirement: Handle TAR hardlinks via two-pass extraction with cross-device fallback
+
+The system SHALL support hardlink extraction from TAR archives; the `linkname` field holds
+the source path, and TAR ordering guarantees the real file (the source) precedes any hardlink
+that references it. Resolution is performed by the `safe-extraction` `ExtractionCoordinator`,
+which acts as a **pull-based sink**: it drives the `ArchiveReader` — inspecting `reader.cost`,
```
davitf: no need for inspecting the cost now, right?

openspec/changes/phase-4-safe-extraction/specs/format-tar/spec.md
```diff
@@ -0,0 +1,98 @@
...
+among the algorithms below. It MUST NOT hold a push-model deferred-state machine, and MUST NOT
+force an upfront pass the run does not need.
+
+Only a `members` selector or a `filter` can **orphan** a hardlink — select a link while
+excluding its source; an unfiltered extract-all never orphans one, because the source is
+always selected and precedes the link. The coordinator uses **one core algorithm**, with an
+**optional optimization** when a free member list is available:
+
+**Core — sequential pass with a conditional second pass.** One forward pass: write each
+selected member; record written FILEs in a per-source `{device → on-disk path}` map; a
```
davitf: is it easy to get the device for a written file? or would we know it only when trying to link? if it's hard, we could just keep a list of extracted paths per source, and the hardlink handler tries them sequentially. but fine to keep if it's easy

openspec/changes/phase-4-safe-extraction/specs/safe-extraction/spec.md
```diff
@@ -4,19 +4,51 @@
 
 ### Requirement: Archive-wide decompression ratio for solid containers
 
-When a member's `compressed_size` is unknown or zero but the reader exposes a known outer
-`compressed_source_size` (the byte length of the compressed container stream — e.g. a
-`.tar.gz` file's size on disk), the system SHALL evaluate the decompression ratio during
-`extract()` / `extract_all()` as:
+The system SHALL evaluate an archive-wide decompression ratio during `extract()` /
```
davitf, line 7: would it be worth also checking the file count in the bomb tracker? e.g. if a tar archive includes a huge number of small files, it can also overwhelm the fs, right? do bomb trackers usually handle this case as well?

PR #25, opened 2026-07-01
Phase 4a: TAR forward-only streaming + strict_eof

davitf:
## Summary
- Adds forward-only TAR reading on non-seekable sources (`streaming=True`) for plain and compressed archives via progressive `tarfile` iteration and a per-backend `SUPPORTS_STREAMING_NON_SEEKABLE` opt-in (ZIP/ISO still fail fast).
- Wires `strict_eof` on `open_archive()` with end-of-archive null-block verification (warn by default, `TruncatedError` when strict).
- Exposes `compressed_source_size` on readers for the upcoming `phase-4-safe-extraction` change.

## Test plan
- [x] `uv run pytest` (588 passed)
- [x] `uv run pyrefly check`
- [x] `uv run ty check`
- [x] `uv run ruff check`
- [ ] Review non-seekable `.tar` / `.tar.gz` / `.tar.bz2` streaming scenarios in `tests/test_tar.py`
- [ ] Confirm ZIP/ISO still reject non-seekable sources under `streaming=True`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/archivey/core.py
+++ b/src/archivey/core.py
@@ -56,6 +56,7 @@ def open_archive(
     streaming: bool = False,
     password: bytes | str | None = None,
     encoding: str | None = None,
+    strict_eof: bool = False,
 ) -> ArchiveReader:
     """Open an archive for reading.
 
@@ -99,9 +100,11 @@ def open_archive(
     backend_cls = registry.reader_for_format(format)
 
     # Fail fast for a seek-requiring backend on a non-seekable source (the access-mode
-    # contract: streaming=False does not implicitly buffer).
+    # contract: streaming=False does not implicitly buffer). Per-backend opt-in only:
+    # TAR may open non-seekable sources under streaming=True; ZIP/ISO still fail fast.
     if (
         backend_cls.REQUIRES_SEEK
+        and not (streaming and backend_cls.SUPPORTS_STREAMING_NON_SEEKABLE)
         and is_stream(open_source)
         and not is_seekable(open_source)
     ):
@@ -126,4 +129,5 @@ def open_archive(
         password=password,
         encoding=effective_encoding,
         archive_name=archive_name,
+        strict_eof=strict_eof,
     )
--- a/src/archivey/core.py
+++ b/src/archivey/core.py
@@ -64,6 +64,12 @@ def open_archive(
     time on a non-seekable source. ``streaming=True`` promises forward-only, single-pass
     access (works on any source, but disables random-access methods).
 
+    ``strict_eof`` controls TAR end-of-archive verification: after all members are read,
+    the TAR backend checks for two null-filled 512-byte EOF blocks. When the check
+    fails, ``strict_eof=False`` (the default) emits a warning; ``strict_eof=True``
+    raises :class:`~archivey.TruncatedError`. This keyword is a Phase 4 stopgap — Phase 5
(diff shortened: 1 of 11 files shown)


src/archivey/internal/backends/tar_reader.py
```diff
@@ -168,23 +165,27 @@ def _open_tarfile(
...
-        self, *, name: str | None = None, fileobj: BinaryIO | None = None
+        self,
+        *,
+        name: str | None = None,
+        fileobj: BinaryIO | None = None,
+        streaming: bool = False,
     ) -> tarfile.TarFile:
-        # mode="r:" reads an *uncompressed* tar stream — we feed it either the raw file (plain
-        # tar) or our own decompressor (compressed tar), never tarfile's native r:gz/r:bz2.
-        # errorlevel=1 makes tarfile raise on fatal read errors (so truncation/corruption
```
davitf, line 181: if you've removed this comment here, it's worth documenting what errorlevel=1 does below in the call itself

src/archivey/internal/backends/tar_reader.py
```diff
@@ -168,23 +165,27 @@ def _open_tarfile(
...
-        # surfaces rather than being silently skipped); we translate those below. encoding=None
-        # is accepted (tarfile then applies its own utf-8 default).
```
davitf, line 183: also worth keeping the encoding=None case as a comment below?

src/archivey/internal/streams/streamtools/binaryio.py
```diff
@@ -98,7 +98,14 @@ def is_seekable(stream: Any) -> bool:
...
         return False
-    if not seekable():
+    try:
+        if not seekable():
+            return False
+    except AttributeError:
+        logger.debug(
+            "Stream %r seekable() raised AttributeError; treating as non-seekable",
+            stream,
+        )
```
davitf, line 115: nice catch! can you document which streams have this issue? (some internal tarfile ones, IIRC?) and can you confirm that they're actually non-seekable?

src/archivey/core.py
```diff
@@ -56,6 +56,7 @@ def open_archive(
     streaming: bool = False,
     password: bytes | str | None = None,
     encoding: str | None = None,
+    strict_eof: bool = False,
```
davitf, line 59: Please document what this arg does. if this is intended to be removed when phase 5 (the config object) is implemented, can you register that please?

tests/test_tar.py
```diff
@@ -273,6 +286,144 @@ def test_non_seekable_tar_fails_fast_via_detection() -> None:
...
+    with open_archive(NonSeekableBytesIO(data), streaming=True) as ar:
+        assert ar.format == ArchiveFormat.TAR_GZ
+        assert ar.cost.stream_capability == StreamCapability.FORWARD_ONLY
+        assert ar.cost.listing_cost == ListingCost.REQUIRES_DECOMPRESSION
+        assert ar.cost.access_cost == AccessCost.SOLID
+        collected = {}
+        for member, stream in ar.stream_members():
+            if stream is not None:
+                collected[member.name] = stream.read()
+        assert collected["hello.txt"] == b"hello world"
```
davitf, line 367: check all members or at least the length, just to ensure the reader doesn't stop scanning for some reason

PR #27, opened 2026-07-01
Review fixes: availability gating, error translation, and robustness

davitf:
Fixes from a full spec-vs-code review, grown through three review rounds into six commits (plus one maintainer commit). Every fix has a regression test; the full suite (628 passed / 15 skipped), Pyrefly, ty, and ruff are clean, and `openspec validate` failures are identical before/after (all pre-existing; the new `phase-5-public-api` change validates strict).

## Commit 1 — straightforward bug fixes

1. **Registry: compressed-tar availability ignored the outer stream codec.** `format_availability(TAR_ZST)` reported `FULL` with the zstd backend absent. Per the `backend-registry` single-codec rule, `tar.` with its sole codec missing is now `NONE` with the codec's install hint, and selection raises `UnsupportedFormatError` up front.
2. **ZIP: raw `RuntimeError` escaped when listing an archive with an encrypted symlink.** The symlink-target read is now routed through the translator: a password failure logs a warning and leaves `link_target` unset (listing stays usable); other failures surface as stamped `ArchiveyError`s.
3. **ZIP: `encoding` parameter was accepted but ignored** — now wired to `zipfile.ZipFile(..., metadata_encoding=...)`, with `raw_name` re-encoding to match.
4. **Gzip accelerator ISIZE backstop: spurious `TruncatedError` on `read(0)`** mid-stream; an explicit `read(0)` is now recognized as not-EOF.
5. **TAR: a crafted PAX mtime beyond `datetime`'s range crashed the listing**; now degrades to `modified=None` with a warning, like `_pax_time` already did.
6. **Password on unencrypted formats was inconsistent** — TAR and directory now raise `UnsupportedOperationError` like single-file and ISO already did.
7. **Streaming `__iter__` yielded members with no `member_id`/`archive_id`** on every backend except TAR; the base now stamps ids progressively.
8. **Cleanup:** unreachable ZIP comment-decoding fallbacks dropped (cp437 maps every byte).

## Commit 2 — review follow-ups

1. **Relative link-target resolution.** A symlink target is now resolved against the link's *own directory* (`dir/link → file` means `dir/file`; `../` walks up; absolute targets stay unresolved and raise `LinkTargetNotFoundError` on open); hardlink targets remain archive-relative (the TAR model). Directory targets match their trailing-slash names. The `archive-reading` spec documents target-name resolution with scenarios.
2. **Streaming passes resolve backward-pointing links progressively.** Hardlinks always point at an earlier member, so a single forward pass resolves them; shared between the base streaming `__iter__` and TAR's streaming `stream_members` via a new `_register_progressively` helper (which also absorbed TAR's duplicated id-stamping).
3. **Stream-position contract.** Detection and the TAR/single-file backends treat a seekable stream's current position as the archive origin (embedded archives work) instead of rewinding to absolute 0; `open_codec_stream` gives mid-positioned streams a clean origin via `fix_stream_start_position` so absolute-offset codec indexes (XZ/lzip, stdlib gzip rewind) stay correct. ZIP tolerates prefix data natively; ISO's position-0 requirement (pycdlib seeks absolutely) is documented. `format-detection` spec updated.
4. **ZIP timestamps.** The NTFS extra field (`0x000A`) FILETIMEs are now parsed; precedence is Extended Timestamp (`0x5455`) > NTFS > DOS `date_time`, each layer overriding only the times it carries. The `format-zip` spec — which previously described a "ZIP64 NT timestamp" matching neither the old code nor reality — is rewritten to match.
5. **`format-zip` spec: Phase 3 → 7 gap documented** — stdlib zipfile cannot decode deflate64/ppmd/zstd members even with the extras installed, so `format_availability(ZIP)` describes the post-Phase-7 composition; mirrored on `_CONTAINER_OPTIONAL_CODECS`.
6. **`MemberSelector`:** documented that the spec's `Collection[ArchiveMember | str]` form is a Phase 5 item; only the predicate form exists today.
7. **Smaller fixes:** the gzip ISIZE check is no longer disarmed by no-op seeks; `ArchiveStream.close()` marks the wrapper closed even when the inner close raises, and a retried `close()` no-ops; `source_name` now lives once in streamtools.

## Commit 3 — drop the dict-style reader protocol (maintainer decision)

- **`__getitem__` and `__len__` removed.** Name lookup is the explicit `get(name, default=None)` (with duplicate names it returns the last — what a sequential extraction leaves on disk); `open()`/`read()` keep accepting a name directly and raise `KeyError` when absent. `len(reader)` raises Python's own `TypeError` in every mode; `list(reader)` simply iterates — which also retires commit 2's `__len__`-raises-`TypeError` workaround.
- **`__contains__` redefined rather than removed** — without it, `in` silently falls back to iterating `__iter__`, which would consume a streaming reader's single forward pass and compare members by value. `member in reader` is now **identity membership** (was this `ArchiveMember` yielded by this reader?): O(1), scan-free, valid in either access mode, useful for disambiguating members across readers. A string operand raises `TypeError` pointing at `get()`.
- Specs updated (`archive-reading` iteration + new "Name lookup and member identity" requirement, `access-mode-and-cost` gate/table, `error-handling`, `format-zip`), plus the SPEC.md/ARCHITECTURE.md prose sketches.

## Commit 4 — review round 3: PR feedback, SlicingStream SEEK_END, roadmap reconciliation

- **`SUPPORTS_PASSWORD` is now `ReadBackend` data**, checked centrally by `open_archive`; the per-backend password rejections in tar/directory/single-file/ISO are gone (ZIP declares `True`; the native 7z/RAR readers will too).
- **Stream-origin normalization moved to `open_archive`:** after detection (which peeks and restores), a seekable stream positioned mid-file is wrapped once in a zero-origin `SlicingStream`, so every backend — including pycdlib's absolute-offset ISO reader — sees the archive begin at `tell() == 0`. Retires the single-file reader's `_source_start` bookkeeping and the ISO position-0 caveat. `ReadOnlyIOStream` gains `mode = "rb"` (typing.IO's abstract `mode` returns `None` at runtime, crashing pycdlib's `'b' not in fp.mode` duck-typing).
- **`SlicingStream`: `SEEK_END` with a non-zero offset and no declared length now works** (probes the underlying end on demand) instead of raising `UnsupportedOperation` for no structural reason.
- **Roadmap/spec reconciliation from the plan review:** PLAN Phase 5 "hashable" → deliberately unhashable, "depth limit" → cycle detection; Phase 8 rescoped to the native seekable-zstd frame-index reader + blocked-gzip; `format-rar` pins the decompressor to RARLAB unrar only (version-verified; unrar-free = missing tool, with a scenario); `format-7z` folder cache bounded per the no-growing-RAM rule; `safe-extraction` lstat-semantics existence checks and unlink-then-create REPLACE; `cli`/`archive-writing` stale wording fixed; Phase 6 pending writing decisions and Phase 7 shared-source plumbing recorded in PLAN; IDEAS gains the fsspec-URL and parallel-extraction sketches, seekable-zstd marked scheduled.

## Commit 5 — `phase-5-public-api` OpenSpec change (proposal, design, spec deltas, tasks)

Encodes the maintainer decisions from the roadmap review, ready for `/opsx:apply`:

- **Passwords:** `password` accepts a single value, an ordered candidate sequence, or a **provider callable** receiving the member being decrypted (`None` for header-level); successful passwords join a per-archive known-good list so streaming passes stay single-pass on multi-password archives. No per-call `open(member, password=)` (the `format-7z` delta is rewritten to the candidate model).
- **Multi-source:** `Sequence[...]` accepted + single-path volume-set sibling discovery (path-only by design); Phase 5 lands signature/plumbing with an explicit `UnsupportedFeatureError` until the Phase 7 joining readers.
- **Config:** frozen `ArchiveyConfig` dataclass (accelerator modes, `strict_archive_eof`, nested `ExtractionLimits`) passed explicitly — no contextvars/ambient config (deliberate departure from DEV); the `strict_eof` kwarg folds into `config.strict_archive_eof`, default `False`.
- **`MemberSelector` collection form:** `str` entries match all duplicates by name; `ArchiveMember` entries match by identity (members are unhashable by design).
- `extract()` results list stays unconditional in v1 (a no-tracking mode needs a no-member-cache reader mode; deferred, recorded in the design).

## Commit 6 — size discovery is cheap-only; streams advertise fsspec-style `.size`

- **`source_byte_size()` no longer `SEEK_END`-probes arbitrary seekable streams** — on a decompressor an end-seek means decompressing the entire payload (and `GzipFile` forwards `fileno()` to the *compressed* file, so fstat duck-checks are unsafe too). New probe order: path stat → integer `.size` attribute (the fsspec convention) → `try_get_size()` (our decompressors' compute-if-cheap hook; its presence makes its answer final) → `SEEK_END` only for a whitelist of provably-O(1) types (`FileIO`, `BufferedReader`/`BufferedRandom` over one, `BytesIO`, `mmap`).
- **`ArchiveStream` and `SlicingStream` gain a `.size` property** — member streams carry `member.size` from archive metadata, codec streams consult the inner `try_get_size()`/`.size` once opened. Net effect: `open_archive(reader.open("inner.zip"))` gets `compressed_source_size` from metadata with zero extra I/O, so the extraction bomb tracker works on nested archives.
- Regression tests pin the trap: `source_byte_size(gzip.open(...))` returns `None` with `tell()` still at 0 — nothing decompressed.

*(Commit 7, `fix: hide stub name on stream wrappers for pycdlib on Windows`, is the maintainer's follow-up to the same typing.IO-stub family of issues as commit 4's `mode` fix.)*

https://claude.ai/code/session_018hQi3WhKr3LnvViH9smD3j

### DIFF
--- a/openspec/specs/format-detection/spec.md
+++ b/openspec/specs/format-detection/spec.md
@@ -277,10 +277,12 @@ and the backend rather than detection consuming bytes the caller can no longer r
 
 - For **paths and seekable streams**: detection reads via `peek`/`read` and restores
   the stream's **starting position** (its `tell()` on entry) afterwards; no wrapper is
-  needed. The archive is taken to begin wherever the caller positioned the stream, so
-  an archive embedded mid-file detects against the right bytes and the backend then
-  reads from the same origin (backends that address the source absolutely — ISO via
-  `pycdlib` — still require position 0; see `format-iso`).
+  needed for detection itself. The archive is taken to begin wherever the caller
+  positioned the stream, so an archive embedded mid-file detects against the right
+  bytes. `open_archive()` then normalizes the origin for the backend: a seekable
+  stream positioned mid-file is wrapped in a zero-origin view (`SlicingStream`) so
+  every backend — including those that address the source with absolute offsets, like
+  ISO via `pycdlib` — sees the archive begin at `tell() == 0`.
 - For **non-seekable streams**: `open_archive()` SHALL wrap the source in a
   `PeekableStream` **before** running detection and pass that same `PeekableStream`
   to both detection and the backend. Detection inspects bytes through
--- a/openspec/specs/format-detection/spec.md
+++ b/openspec/specs/format-detection/spec.md
@@ -276,7 +276,11 @@ responsibility, not `detect_format()`'s, so that one wrapper is shared by detect
 and the backend rather than detection consuming bytes the caller can no longer reach:
 
 - For **paths and seekable streams**: detection reads via `peek`/`read` and restores
-  the position with `seek(0)` afterwards; no wrapper is needed.
+  the stream's **starting position** (its `tell()` on entry) afterwards; no wrapper is
+  needed. The archive is taken to begin wherever the caller positioned the stream, so
+  an archive embedded mid-file detects against the right bytes and the backend then
+  reads from the same origin (backends that address the source absolutely — ISO via
+  `pycdlib` — still require position 0; see `format-iso`).
 - For **non-seekable streams**: `open_archive()` SHALL wrap the source in a
   `PeekableStream` **before** running detection and pass that same `PeekableStream`
   to both detection and the backend. Detection inspects bytes through
@@ -308,11 +312,11 @@ this wrapping internally, so callers of the high-level API never wrap by hand.
 └──────────────────────────────────────────────────────────────┘
 ```
 
-#### Scenario: seekable stream is rewound
+#### Scenario: seekable stream position is preserved
 
-- **WHEN** `detect_format()` is called with a seekable `BinaryIO` at position 0
-- **THEN** after detection completes the stream position SHALL be 0
-- **AND** the backend can read the full stream from the start without any data loss
+- **WHEN** `detect_format()` is called with a seekable `BinaryIO` at any position N
+- **THEN** after detection completes the stream position SHALL again be N
+- **AND** the backend can read the full archive from that origin without any data loss
 
 #### Scenario: non-seekable source wrapped once by the opener and shared
 
--- a/openspec/specs/format-zip/spec.md
+++ b/openspec/specs/format-zip/spec.md
@@ -33,10 +33,28 @@ The system SHALL expose the following cost and capability properties for every o
 The system SHALL map each `ZipInfo` entry to a `ArchiveMember` dataclass using the following field rules:
 
 - `mode`: parsed from `external_attr >> 16`. If `external_attr == 0` and `create_system != 3` (Unix), `mode` is set to `None`.
-- `modified`: from the `date_time` tuple, constructed as a naive `datetime` (no TZ; DOS format has 2-second granularity). If the ZIP64 extra field contains an NT timestamp, use that as a timezone-aware UTC `datetime` instead.
+- `modified`/`accessed`/`created`: layered by precedence, each layer overriding only the
+  times it actually carries. Base: the DOS `date_time` tuple as a naive `datetime` (no TZ;
+  local wall-clock, 2-second granularity; `None` for the year-1980 "no timestamp"
+  sentinel). Above it: the NTFS extra field (`0x000A`) — three 64-bit FILETIMEs
+  (modification/access/creation, 100 ns UTC ticks since 1601, zero = "not set"; written
+  by Windows tools such as 7-Zip) — as timezone-aware UTC `datetime`s. Highest: the
+  Extended Timestamp extra field (`0x5455`) — signed 32-bit Unix times, its flags byte
+  signaling which of modification/access/creation are present — as timezone-aware UTC
+  `datetime`s.
 - `type`: inferred from `mode` if Unix, otherwise from `is_dir()` and symlink detection via extra field `0x000A` (NTFS) or `0x7875` (Unix UID/GID).
 - `compression`: map `compress_type` integer to `CompressionMethod`.
 - `is_encrypted`: set to `True` when `flag_bits & 0x1` is non-zero.
+
+> **Phase 3 → 7 gap (member decode via stdlib zipfile).** Until the
+> `compressed-streams` codec layer is wired into ZIP member reads (Phase 7, alongside
+> the native 7z reader's container codecs), member *data* decompression goes through
+> stdlib `zipfile`, which cannot decode deflate64/PPMd (or zstd before Python 3.14)
+> even when the corresponding codec packages are installed — reading such a member
+> raises `UnsupportedFeatureError`. `format_availability(ZIP)`'s FULL/PARTIAL result
+> (see `backend-registry`) describes the intended post-Phase-7 composition over those
+> codecs, so until then it can report FULL while these rare member codecs still fail
+> at read time. Listing is unaffected.
 
 #### Scenario: Unix mode from external_attr
 
@@ -48,10 +66,15 @@ The system SHALL map each `ZipInfo` entry to a `ArchiveMember` dataclass using t
 - **WHEN** a ZIP entry has `external_attr == 0` or `create_system != 3`
 - **THEN** `member.mode` is set to `None`
 
-#### Scenario: NT timestamp takes precedence over DOS date_time
+#### Scenario: Extended Timestamp takes precedence over DOS date_time
 
-- **WHEN** a ZIP entry carries a ZIP64 extra field containing an NT timestamp
-- **THEN** `member.modified` is a timezone-aware UTC `datetime` derived from that NT timestamp, overriding the value from `date_time`
+- **WHEN** a ZIP entry carries an Extended Timestamp extra field (`0x5455`) with a modification time
+- **THEN** `member.modified` is a timezone-aware UTC `datetime` derived from that Unix time, overriding the value from `date_time`
+
+#### Scenario: NTFS timestamps used when no Extended Timestamp is present
+
+- **WHEN** a ZIP entry carries an NTFS extra field (`0x000A`) with non-zero FILETIMEs and no `0x5455` field
+- **THEN** `member.modified`/`accessed`/`created` are timezone-aware UTC `datetime`s derived from those FILETIMEs, overriding the value from `date_time`
 
 #### Scenario: Encrypted entry detection
 
--- a/openspec/specs/format-zip/spec.md
+++ b/openspec/specs/format-zip/spec.md
@@ -26,7 +26,7 @@ The system SHALL expose the following cost and capability properties for every o
 #### Scenario: Central directory lookup is O(1)
 
 - **WHEN** `reader.get("some/member.txt")` is called on a ZIP reader
-- **THEN** the lookup is satisfied via the in-memory `NameToInfo` dict with no additional I/O
+- **THEN** the lookup is satisfied via the reader's in-memory name→member map (built from the central directory) with no additional I/O
 
 ### Requirement: Map ZIP member metadata to the unified ArchiveMember model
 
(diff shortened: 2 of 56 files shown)


src/archivey/internal/backends/iso_reader.py
```diff
@@ -12,6 +12,11 @@
 out of scope (``UnsupportedOperationError``). The image is read uncompressed in place — a
 compressed ``.iso.xz`` is a single-file compressor wrapping the image, not mounted here
 (see the seek-heavy-container note in the proposal).
+
+Stream-position caveat: unlike the TAR/single-file backends, an ISO stream source must
+start at position 0 — ``pycdlib`` addresses the image with absolute offsets (the PVD at
+32 KiB etc.), so a mid-positioned stream is not supported (wrap the region in a
+``SlicingStream`` first if that is ever needed).
```
davitf: why don't we do it in the constructor here? we could use fix_stream_start_position() to wrap the incoming stream if needed before passing to pycdlib

src/archivey/internal/backends/directory_reader.py
```diff
@@ -225,6 +226,11 @@ def open_read(
         # uniform ReadBackend signature.
         if not isinstance(source, Path):
             raise TypeError("Directory backend requires a Path source")
+        if password is not None:
```
davitf: do you think it's worth making SUPPORTS_PASSWORD as a ReadBackend field, so it can be checked at the super level?

openspec/specs/format-detection/spec.md
```diff
@@ -276,7 +276,11 @@ responsibility, not `detect_format()`'s, so that one wrapper is shared by detect
 and the backend rather than detection consuming bytes the caller can no longer reach:
 
 - For **paths and seekable streams**: detection reads via `peek`/`read` and restores
-  the position with `seek(0)` afterwards; no wrapper is needed.
+  the stream's **starting position** (its `tell()` on entry) afterwards; no wrapper is
+  needed. The archive is taken to begin wherever the caller positioned the stream, so
+  an archive embedded mid-file detects against the right bytes and the backend then
+  reads from the same origin (backends that address the source absolutely — ISO via
+  `pycdlib` — still require position 0; see `format-iso`).
```
davitf: as mentioned in another comment, can we support it via wrapping the stream, for consistency?

openspec/specs/format-zip/spec.md
```diff
@@ -25,19 +25,37 @@ The system SHALL expose the following cost and capability properties for every o
 
 #### Scenario: Central directory lookup is O(1)
 
-- **WHEN** `reader["some/member.txt"]` is called on a ZIP reader
+- **WHEN** `reader.get("some/member.txt")` is called on a ZIP reader
 - **THEN** the lookup is satisfied via the in-memory `NameToInfo` dict with no additional I/O
```
davitf: `NameToInfo` is a zipfile internal dict, right? looking at BaseArchiveReader, it keeps its own mapping of name to member in `_members_by_name`, which is what get() uses. (and that's the correct behavior even for zips, just this line seems incorrect)

openspec/specs/format-zip/spec.md
```diff
@@ -25,19 +25,37 @@ The system SHALL expose the following cost and capability properties for every o
...
+  Extended Timestamp extra field (`0x5455`) — signed 32-bit Unix times, its flags byte
+  signaling which of modification/access/creation are present — as timezone-aware UTC
+  `datetime`s.
 - `type`: inferred from `mode` if Unix, otherwise from `is_dir()` and symlink detection via extra field `0x000A` (NTFS) or `0x7875` (Unix UID/GID).
 - `compression`: map `compress_type` integer to `CompressionMethod`.
 - `is_encrypted`: set to `True` when `flag_bits & 0x1` is non-zero.
 
+> **Phase 3 → 7 gap (member decode via stdlib zipfile).** Until the
+> `compressed-streams` codec layer is wired into ZIP member reads (Phase 7, alongside
+> the native 7z reader's container codecs), member *data* decompression goes through
```
davitf: do you mean the internal zip file reader replacing zipfile? that's in the ideas log, but not specified in phase 7 yet, right? or does phase 7 specify replacing the zipfile decompressors while still using it (is that even possible)?

src/archivey/internal/backends/single_file_reader.py
```diff
@@ -91,6 +91,11 @@ def __init__(
         self._stream_codec = stream_codec_for_format(format.stream)
         self._codec = self._stream_codec.codec
         self._seekable = not is_stream(source) or is_seekable(source)
+        # The compressed stream starts wherever the caller positioned the source (the
+        # stream-position contract): remember it so re-opens rewind *here*, not to 0.
+        self._source_start = (
+            source.tell() if is_stream(source) and is_seekable(source) else 0
+        )
```
davitf: would it be easier to also use fix_stream_start_position here? (and now that we expect this behavior in all backends, maybe we should just wrap any non-0 stream in it at open_archive time. what do you think?)

src/archivey/internal/backends/tar_reader.py
```diff
@@ -132,6 +133,11 @@ def __init__(
         archive_name: str | None,
         strict_eof: bool = False,
     ) -> None:
+        if password is not None:
+            raise UnsupportedOperationError(
+                "TAR archives do not support passwords (they carry no encryption).",
+                archive_name=archive_name,
+            )
```
davitf: also here, worth adding a ReadBackend field instead?

src/archivey/internal/backends/iso_reader.py
davitf: Done, via the open_archive-level wrap you suggested in the single-file thread: the opener now normalizes any mid-positioned seekable stream once (after detection), so ISO gets this for free and the caveat is gone. One wrinkle this surfaced: pycdlib duck-types `fp.mode` (`'b' not in fp.mode`), and `typing.IO`'s abstract `mode` property returns `None` at runtime on our wrappers — `ReadOnlyIOStream` now exposes `mode = "rb"`. Covered by a new mid-positioned-ISO test.

---

src/archivey/internal/backends/directory_reader.py
davitf: Yes — done. `ReadBackend.SUPPORTS_PASSWORD` (default `False`; ZIP declares `True`, the native 7z/RAR readers will too), checked centrally in `open_archive` before the backend is even constructed, so backends never see a password they can't use. The four per-backend checks (tar, directory, single-file, ISO) are removed.

---

openspec/specs/format-detection/spec.md
davitf: Done — `open_archive` now normalizes the origin for every backend (one `fix_stream_start_position` wrap after detection), and this spec section states that rule instead of the ISO exception. Detection itself still just peeks-and-restores, so standalone `detect_format()` on a mid-positioned stream also behaves.

---

openspec/specs/format-zip/spec.md
davitf: Correct on both counts — `NameToInfo` is zipfile's internal dict, and `get()` is served by the base reader's `_members_by_name`. Fixed the scenario to say "the reader's in-memory name→member map (built from the central directory)".

---

openspec/specs/format-zip/spec.md
davitf: Neither, exactly — it's a middle path that wasn't written down anywhere yet, so I've now spelled it out in the note. Replacing zipfile's decompressors in place isn't possible (no plug-in point; `ZipExtFile` hardcodes its codec dispatch). The intended Phase 7 mechanism: keep zipfile for the central directory, but for member *data* slice the raw compressed bytes ourselves (local-header offset + `SlicingStream`) and decode through the shared codec layer — same pattern the native 7z reader uses for its folders, and a first concrete step toward the full native ZIP reader in IDEAS (which additionally replaces the header parsing for non-seekable/multi-volume/AES support). project.md already assigned "ZIP container codecs" to Phase 7; the note now says how.

---

src/archivey/internal/backends/single_file_reader.py
davitf: Agreed — implemented the open_archive-level version: after detection (which peeks from and restores the caller's position), any seekable stream not at position 0 is wrapped once in `fix_stream_start_position`. Every backend now sees the archive at `tell() == 0`, so the single-file reader's `_source_start` bookkeeping is deleted, ISO works on mid-positioned streams for free, and the codec-layer wrap becomes a no-op belt-and-braces for standalone codec users. format-detection spec updated to state the opener-normalizes rule.

---

src/archivey/internal/backends/tar_reader.py
davitf: Done — same `SUPPORTS_PASSWORD` backend field, checked centrally in `open_archive` (see the directory-reader thread); this per-backend check is removed.

---

PR #28, opened 2026-07-02
Phase 4b: safe extraction (ExtractionCoordinator + bomb limits)

davitf:
Implements the **`safe-extraction`** capability (openspec change `phase-4-safe-extraction`): members are written to disk safely behind path-safety filters, policy transforms, and decompression-bomb limits.

## What's here

**New internal modules**
- `internal/progress.py` — `ExtractionPolicy` / `OverwritePolicy` / `OnError` / `ExtractionStatus` enums, `ExtractionProgress` / `ExtractionResult` dataclasses, shared selector/filter aliases.
- `internal/filters.py` — `check_universal()` (path traversal, absolute paths, null bytes, symlink/hardlink escape, special files) + `STRICT`/`STANDARD`/`TRUSTED` policy transforms.
- `internal/extraction.py` — `BombTracker` (cumulative bytes, per-member ratio, archive-wide ratio, entry count) and `ExtractionCoordinator`, a **pull-based sink** (no push-model `ExtractionHelper`): FILE/DIR/SYMLINK/HARDLINK writes, overwrite policy with lstat + unlink-then-create symlink safety, `OnError`, and the core hardlink algorithm (single forward pass + conditional seekable second pass + cross-device sibling linking).

**Wiring**
- `ArchiveReader.extract_all(...)` full signature (was a `NotImplementedError` stub) → delegates to the coordinator.
- Top-level `archivey.extract(...)` one-shot API.
- New public extraction types exported from `archivey`.

## Decisions & deviations (please review)

- **Interim path-traversal handling.** Read-time `normalize_member_name` *silently collapses* traversal (`../evil` → `evil`), so `check_universal` can't see the escape on `member.name`. As a stopgap it also inspects `member.raw_name` and rejects a hostile **stored** name, satisfying `testing-contract` today. The principled fix — minimal, meaning-preserving normalization so `member.name` stays truthful and the check runs on it alone — is proposed as a **separate change** (`minimal-name-normalization`, also in this PR's openspec dir) per maintainer direction to keep this change scoped.
- **Archive-wide ratio = always-stop.** The spec explicitly classifies cumulative bytes + entry-count as always-stop and the per-member ratio as skippable, but leaves the archive-wide ratio unclassified. Treated as always-stop (implemented via a private `ExtractionError` subclass so `except ExtractionError` still catches it).
- **STANDARD keeps uid/gid** (per the transforms table + DEV's `tar` filter; the prose bullet's "remove uid/gid" is the outlier). No on-disk effect — the coordinator only `chown`s under `TRUSTED` as root.
- **`members=[...]` collection selector** is normalized to a predicate inside the coordinator, leaving `stream_members`' Phase-5 duplicate-name deferral untouched.
- **Deferred (task 4.2):** the optional planned-single-pass optimization. The core algorithm handles every case correctly and the current TAR backend rarely exposes a free member list, so it's left for a follow-up.

## Tests

`tests/test_extraction.py` (50 tests): universal filters + policy transforms, `BombTracker` guards, ZIP/TAR/directory extraction, overwrite policies (incl. planted-symlink no-write-through, dangling-symlink-as-existing), `OnError` STOP/CONTINUE, symlink escape + dangling-to-filtered-target, hardlink inode sharing + seekable orphan recovery + forward-only `OnError` + cross-device chained linking, seekable/non-seekable `tar.gz`, and the `testing-contract` adversarial scenarios (path traversal, zip bomb, entry-count bomb).

Full suite: **680 passed, 15 skipped**. `pyrefly` + `ty` + `ruff` clean. `openspec validate phase-4-safe-extraction --strict` passes.

https://claude.ai/code/session_01FqR2Cxqm8k4aRhnKfkErty

---
_Generated by [Claude Code](https://claude.ai/code/session_01FqR2Cxqm8k4aRhnKfkErty)_

### DIFF
--- a/src/archivey/internal/extraction.py
+++ b/src/archivey/internal/extraction.py
@@ -618,11 +618,12 @@ class ExtractionCoordinator:
         truncates or removes an existing destination (only the temp is discarded), and the
         target name never appears half-written. The temp lives in the destination directory
         so the rename stays on one filesystem."""
+        # mkstemp hands back an already-open fd; write straight into it (no close+reopen).
         fd, tmp_name = tempfile.mkstemp(dir=dest_path.parent, prefix=".archivey-tmp-")
-        os.close(fd)
         tmp = Path(tmp_name)
         try:
-            self._copy_stream_to(stream, tmp, tracker)
+            with os.fdopen(fd, "wb") as dst:
+                self._copy_to_fileobj(stream, dst, tracker)
             self._apply_metadata(tmp, member)
             os.replace(tmp, dest_path)
         except BaseException:
@@ -632,19 +633,28 @@ class ExtractionCoordinator:
                 os.unlink(tmp)
             raise
 
+    @staticmethod
+    def _copy_to_fileobj(
+        stream: BinaryIO | None, dst: BinaryIO, tracker: BombTracker | None
+    ) -> None:
+        """Copy ``stream`` into the already-open ``dst`` in chunks, counting each toward the
+        bomb tracker. Cleanup of a partial ``dst`` on failure is the caller's job."""
+        if stream is None:
+            return
+        while True:
+            chunk = stream.read(_CHUNK)
+            if not chunk:
+                break
+            if tracker is not None:
+                tracker.count(len(chunk))
+            dst.write(chunk)
+
     def _copy_stream_to(
         self, stream: BinaryIO | None, path: Path, tracker: BombTracker | None
     ) -> None:
         try:
             with open(path, "wb") as dst:
-                if stream is not None:
-                    while True:
-                        chunk = stream.read(_CHUNK)
-                        if not chunk:
-                            break
-                        if tracker is not None:
-                            tracker.count(len(chunk))
-                        dst.write(chunk)
+                self._copy_to_fileobj(stream, dst, tracker)
         except BaseException:
             # Remove this member's partial output on any failure (bomb guard, write
             # error, KeyboardInterrupt) and re-raise. No log here on purpose: the failure
--- a/src/archivey/internal/extraction.py
+++ b/src/archivey/internal/extraction.py
@@ -20,10 +20,12 @@ optimization over this core, not a correctness requirement.
 
 from __future__ import annotations
 
+import contextlib
 import errno
 import logging
 import os
 import shutil
+import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, BinaryIO, Callable, Collection, cast
@@ -356,12 +358,11 @@ class ExtractionCoordinator:
         tracker: BombTracker,
         source_paths: dict[int, list[Path]],
     ) -> ExtractionResult:
-        if not self._prepare_destination(transformed, dest_path):
+        if not self._prepare_destination(transformed, dest_path, atomic=True):
             return ExtractionResult(original, None, ExtractionStatus.SKIPPED, None)
 
         os.makedirs(dest_path.parent, exist_ok=True)
-        self._copy_stream_to(stream, dest_path, tracker)
-        self._apply_metadata(dest_path, transformed)
+        self._write_file_atomic(stream, dest_path, transformed, tracker)
 
         # Record this FILE's path under the ORIGINAL member id so later hardlinks whose
         # link_target_member is this member can os.link against it.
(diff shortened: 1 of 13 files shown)


src/archivey/internal/extraction_types.py
davitf, line 1: this is not just progress-related; rename the module?

src/archivey/internal/progress.py
```diff
@@ -0,0 +1,92 @@
...
+
+    STRICT = "strict"  # default; untrusted archives
+    STANDARD = "standard"  # moderate trust; e.g. your own older archives
+    TRUSTED = "trusted"  # apply stored mode (and uid/gid as root); path safety still enforced
+
+
+class OverwritePolicy(Enum):
+    """What to do when a destination entry already exists where a member would be written."""
+
+    ERROR = "error"  # raise ExtractionError if a destination entry exists
```
davitf: "treat as an error" instead? it's handled according to the OnError policy, right? if not, it should

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,695 @@
...
+                        if not chunk:
+                            break
+                        if tracker is not None:
+                            tracker.count(len(chunk))
+                        dst.write(chunk)
+        except BaseException:
+            # Remove this member's partial output on any failure (bomb guard, write
+            # error, KeyboardInterrupt) before re-raising.
+            try:
+                os.unlink(path)
```
davitf: is a warning log warranted here?

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,695 @@
...
+            )
+        if self._overwrite is OverwritePolicy.SKIP:
+            return False
+
+        # REPLACE: unlink-then-create, never write-through. Remove a symlink/file with
+        # unlink (so bytes never follow the link to its target); remove a real dir tree.
+        if dest_path.is_symlink() or not dest_path.is_dir():
+            dest_path.unlink()
+        else:
+            shutil.rmtree(dest_path)
```
davitf: Given that _copy_stream_to deletes files in case of an extraction failure, should we avoid deleting the old file until the new one is ready? Maybe we can extract to a temporary file, and move it on top of the old one after a successful extraction.

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,695 @@
...
+        os.makedirs(dest_path.parent, exist_ok=True)
+        # A symlink is target-independent: create it even if the target was filtered out,
+        # appears later, or lies outside the archive — it may dangle. Only the escape
+        # check below constrains it. An os.symlink failure (unsupported FS) propagates as
+        # a per-member OnError failure; no copy-the-target fallback.
+        os.symlink(target, dest_path)
+
+        # Post-creation re-resolution (defense-in-depth layer 3): the created link's
+        # resolved target must stay within dest, catching chained-symlink TOCTOU. An
+        # unresolvable link (ELOOP / RuntimeError) is treated as an escape.
```
davitf: is "layer 3" a common security concept, or something we've defined in our specs? can you explain in simpler terms why this is needed here, and we can't just rely on the before-linking check (which is done in check_universal, right? worth pointing to it here)?

If there's a specific concern (chained symlinks? what does it actually mean?), couldn't we re-check right before linking, instead of checking and unlinking afterwards? or is this a last-resort check to be super super safe? (we may keep it, but should justify it properly)

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,752 @@
...
+        """Write a FILE by streaming into a temp sibling, applying metadata, then
+        ``os.replace()``-ing it onto ``dest_path`` — atomic, so a mid-stream failure never
+        truncates or removes an existing destination (only the temp is discarded), and the
+        target name never appears half-written. The temp lives in the destination directory
+        so the rename stays on one filesystem."""
+        fd, tmp_name = tempfile.mkstemp(dir=dest_path.parent, prefix=".archivey-tmp-")
+        os.close(fd)
+        tmp = Path(tmp_name)
+        try:
+            self._copy_stream_to(stream, tmp, tracker)
```
davitf: could we decompress to the opened file directly, instead of closing and re-opening?

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,762 @@
...
+        results: list[ExtractionResult],
+    ) -> None:
+        # Count the recovered source bytes toward the cumulative/ratio guards too.
+        tracker.start_member(source_member)
+        first = group[0]
+        # The excluded source is written to the first selected link's destination path,
+        # never to its own name.
+        if self._prepare_destination(first.original, first.dest_path):
+            os.makedirs(first.dest_path.parent, exist_ok=True)
+            self._copy_stream_to(stream, first.dest_path, tracker)
```
davitf: do we need the strategy of writing to temp then renaming here as well?

src/archivey/internal/extraction.py
```diff
@@ -0,0 +1,762 @@
...
+            source_paths.setdefault(source_member.member_id, []).append(first.dest_path)
+            results[first.result_index] = ExtractionResult(
+                first.original, first.dest_path, ExtractionStatus.EXTRACTED, None
+            )
+        else:
+            results[first.result_index] = ExtractionResult(
+                first.original, None, ExtractionStatus.SKIPPED, None
+            )
+
+        for orphan in group[1:]:
```
davitf, line 551: do we need that strategy of trying to link to all written paths, then copying only if none worked? or does this already do it? if we need it, should we put it into a common helper so both places can use it, or are they too different?

PR #31, opened 2026-07-03
live-decompression-ratio-guard (streaming bomb protection for unknown-size pipes)

davitf:
**Spec-only** proposal (no code yet) capturing davitf's idea for closing a decompression-bomb gap.

## The gap
The ratio guard is **static-denominator only**: the per-member ratio needs `member.compressed_size` (`None` for TAR members) and the archive-wide ratio needs `compressed_source_size` (`None` for a pipe). So a `streaming=True` `.tar.gz` from a **non-seekable pipe** has *no* ratio protection — only the absolute `max_extracted_bytes` cap, which still lets ~2 GiB out of a few KiB of input.

## The idea
Measure the ratio **live**: `uncompressed_written / compressed_bytes_consumed`, using a running count of compressed bytes pulled from the source, and trip once it exceeds `max_ratio` past the activation floor. No total size needed — a pipe yields the running consumed count even though it never yields the total. It's the general form of the existing archive-wide ratio (a static `compressed_size` is just the *final* value of that counter).

## Shape
- **`compressed-streams` (ADDED):** the decompression layer exposes a running `compressed bytes consumed` count (a counting reader around the raw source), surfaced on the reader as `compressed_bytes_consumed` — cheap, pipe-safe.
- **`safe-extraction` (ADDED):** `BombTracker` computes the live ratio when both static denominators are absent; **cumulative/archive-wide** (solid streams can't attribute compressed bytes per member), an always-stop guard, and it **complements** (doesn't replace) the static checks — the static path is used when a total is known, so no double-counting.

## Design notes captured (design.md)
Counter placement across backends, accelerator (`rapidgzip`/`indexed_bzip2`) gating, engagement condition, and why it's cumulative-only. Safe on uncompressed sources (~1:1 never trips).

## Sequencing
Layers on the `BombTracker` from **#28** (phase-4b safe extraction) — land after it merges, then `/openspec-apply` this.

`openspec validate live-decompression-ratio-guard --strict` passes.

https://claude.ai/code/session_01FqR2Cxqm8k4aRhnKfkErty

---
_Generated by [Claude Code](https://claude.ai/code/session_01FqR2Cxqm8k4aRhnKfkErty)_


src/archivey/internal/detection.py
```diff
@@ -134,7 +135,14 @@ def _probe_inner_tar(prefix: bytes, stream_format: StreamFormat) -> bool:
...
         return False
     try:
-        with open_codec_stream(codec, io.BytesIO(prefix)) as stream:
+        # A bounded prefix is often shorter than the full compressed stream (detection
+        # peeks at most DETECTION_LIMIT bytes). The sequential backend must be used here:
+        # the rapidgzip accelerator expects a complete/seekable source and fails on a
+        # truncated prefix, which would defer a .tar.gz to bare .gz and break streaming
+        # live-ratio guards downstream.
+        with open_codec_stream(
+            codec, io.BytesIO(prefix), config=StreamConfig(streaming=True)
```
davitf: the other agent fixed the issue by setting streaming=True to avoid rapidgzip. is it still needed? does it make sense in general, or should we remove it now that we've fixed the issue some other way?

src/archivey/internal/streams/counting.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import BinaryIO
+
+from archivey.internal.streams.streamtools import DelegatingStream
+
+
+class CountingReader(DelegatingStream):
+    """Wrap ``inner`` and count the total number of bytes returned by reads.
+
+    ``readinto_passthrough=False`` routes ``readinto`` through ``read`` so the count is
+    updated on every read path (see :class:`DelegatingStream`). The count reflects bytes
```
davitf: this stream is so simple that I think it's worth overriding readinto here instead of disallowing passthrough. what do you think?

src/archivey/internal/base_reader.py
```diff
@@ -396,6 +405,35 @@ def compressed_source_size(self) -> int | None:
...
+        (see ``safe-extraction``), used when ``compressed_source_size`` is ``None`` — i.e. a
+        compressed archive read from a non-seekable pipe, whose total size is not cheaply
+        knowable. A backend that decompresses a *stream* source wraps it in a
+        ``CountingReader`` and records it here; readers with a knowable source size (a path,
+        a seekable stream) leave it ``None`` and rely on the cheaper static ratio instead.
+        """
+        c = self._compressed_input_counter
+        return c.bytes_read if c is not None else None
+
+    def _count_compressed_input(self, source: Path | BinaryIO) -> Path | BinaryIO:
```
davitf: this method name is weird, it sounds like it would do the counting. maybe a "wrap" something? also, it's slightly weird that each backend needs to call it directly; is there a simple way to make the base reader itself do the wrapping (without introducing worse patterns)?

src/archivey/internal/extraction.py
```diff
@@ -137,16 +141,29 @@ def count(self, chunk_bytes: int) -> None:
...
+                        f"{self._max_ratio:.0f}:1"
+                    )
+            elif self._compressed_consumed is not None:
+                consumed = self._compressed_consumed()
+                if consumed and consumed > 0 and self._total_bytes / consumed > self._max_ratio:
+                    raise _AlwaysStopExtractionError(
+                        f"Live decompression ratio "
+                        f"{self._total_bytes / consumed:.0f}:1 exceeds limit "
+                        f"{self._max_ratio:.0f}:1"
+                    )
```
davitf, line 170: shouldn't this detection live inside the bomb tracker, and the extractor just delegate to it?

src/archivey/internal/detection.py
davitf: Dropped from this PR entirely. That `streaming=True` change was the other agent's inner-tar fix (commit `8464404`), unrelated to the ratio guard — it only rode along because this branch was cut on top of it. I rebased it out, so `detection.py` here is now identical to `main`. The inner-tar fix (which does still need `streaming=True`, plus the full-block read) lives solely in #32 now, so there's no longer an overlap to reconcile between the two PRs.

---

src/archivey/internal/streams/counting.py
davitf: Done. `CountingReader` now overrides `readinto` to count directly (zero-copy inner `readinto` when available, plus a fallback that routes through `read` — which already counts — with no double-count), and I dropped `readinto_passthrough=False`. Both read paths keep the count current without forcing the extra copy.

---

src/archivey/internal/base_reader.py
davitf: Renamed to `_wrap_compressed_input` (it wraps the source in a `CountingReader`; it doesn't do the counting).

On making the base reader wrap automatically: I kept the explicit call. The two callers (`single_file_reader`, `tar_reader`) wrap the source at different points, each right before handing it to `open_codec_stream`, and only the two stream-decompressing backends need it at all (zip/7z/etc. address their source by seeking, never wrapping it). A base-level auto-wrap would need a template-method seam threaded through those distinct open paths — more machinery than the one-line explicit call at each site. Happy to revisit if you'd prefer the seam.

---

src/archivey/internal/extraction.py
davitf: Went with the "pass the reader" option. `BombTracker` now takes `source=reader` and fetches both denominators itself — the static `compressed_source_size` (captured once) and the live `compressed_bytes_consumed` (sampled fresh each `count()`). `ExtractionCoordinator.run()` just hands over the source; the wiring lambda is gone, and the ratio calculation/raising stays entirely inside the tracker. Added a unit test for the live (pipe) denominator path too.

---

PR #32, opened 2026-07-04
detection: inner-TAR probe reads a full block so large-block .tar.bz2…

davitf:
… is detected

The inner-TAR probe only decompressed the 4 KiB detection prefix. bzip2 is block-transform (BWT) based and emits no output until a whole block (up to 900 KB) is read, so a .tar.bz2 whose first block compresses to more than the prefix (any leading member with ~5 KB+ of incompressible data) decoded to zero bytes and was mis-reported as bare .bz2 — with no open-time re-probe to recover it. Stream-oriented codecs (gzip/xz/zstd/...) were unaffected.

The probe now reads from the actual source: it decodes the peeked prefix first and, on a TruncatedError (the block-codec 'need more input' signal), reads up to one maximum block (1 MiB, covering a filled bzip2 level-9 block) and retries. The extra read is bounded and, like the prefix peek, does not consume the source (seekable restored, PeekableStream buffered). The decode forces the sequential backend so rapidgzip isn't engaged on a bounded prefix. Applies uniformly to all single-file compressors; no public API change and no format newly rejected.

Claude-Session: https://claude.ai/code/session_01FqR2Cxqm8k4aRhnKfkErty


src/archivey/internal/detection.py
```diff
@@ -133,29 +152,51 @@ def _probe_inner_tar(prefix: bytes, stream_format: StreamFormat) -> bool:
...
+            return None  # the compressed prefix ended mid-block; more input may complete it
+        except (ArchiveyError, OSError, ValueError):
+            return b""  # not decodable as this codec at all -> not an inner tar
+
+    head = _decode(prefix)
+    if head is None:
+        # A block-transform codec (bzip2) buffered a first block larger than the peeked
+        # prefix; read up to one maximum block from the source and retry once. A genuinely
+        # truncated stream simply yields no more bytes and still resolves to "no inner tar".
+        head = _decode(peek_more(_INNER_TAR_MAX_PROBE_BYTES))
```
davitf: why don´t we pass the original stream to _decode (or a wrapper over it that just limits the max read amount, e.g. a SlicingStream), so it reads as much as needed for decoding instead of doing a fallback? I don't know if it's possible cleanly, but think about what results in the cleaner code overall, rather than minimizing diffs

davitf: Done in e4604d0. The probe now decodes through a bounded, non-consuming view of the source (`_BoundedPeekReader` over the existing `peek_more` callable) handed straight to `open_codec_stream` — one decode path, the codec pulls exactly what it needs (a few KiB for gzip/xz/…, up to one max block for bzip2), and the `TruncatedError`-driven re-decode is gone.

One note: I wrapped `peek_more` rather than the raw stream, because the source here can be a *path* (opened per-peek), not only a stream, and detection must stay non-consuming — `peek_more` already unifies path / seekable / `PeekableStream` behind that contract, whereas a `SlicingStream` over the raw stream would consume it. The reader caches its last peek so growth stays linear.

---

PR #35, opened 2026-07-04
fix(zip): translate open-time errors (overlap zip-bomb guard, bad password) to ArchiveyError

davitf:
## What & why

While reviewing the library's zip-bomb resistance against Fifield's overlapping-entries construction (https://www.bamsoftware.com/hacks/zipbomb/), I found that the ZIP backend was the **only** reader that called its underlying library's `open()` without the translate+stamp wrapper every other backend (tar/iso/single-file) uses.

stdlib `zipfile.ZipFile.open()` validates the local file header **at open time** — before any byte flows through archivey's read-time `ArchiveStream` translator. That means several stdlib exceptions were leaking to callers as raw, untranslated exceptions instead of `ArchiveyError`s, breaking the exception-translation contract:

- **`BadZipFile("Overlapped entries ... (possible zip bomb)")`** — CPython's overlap guard, the direct defense against the paper's bomb. Leaked as a raw `zipfile.BadZipFile` from `open()`/`extract()`.
- `BadZipFile("File name in directory ... differ")` — the name-mismatch check.
- `RuntimeError("... password required ...")` / `"Bad password"` — encrypted member opened without / wrong password.
- `NotImplementedError("... compression method ...")` — unsupported compression.

## The fix

`zip_reader._open_member` now wraps `self._archive.open(...)` in the same translate+stamp path the other backends use, catching `(zipfile.BadZipFile, RuntimeError, io.UnsupportedOperation)` and routing them through the existing `_translate_exception` (which already maps these to `CorruptionError` / `EncryptionError` / `UnsupportedFeatureError` / `StreamNotSeekableError`). A genuine `OSError` from the underlying source is deliberately **not** in the caught set, so it still propagates unchanged.

Now an overlap bomb surfaces as a `CorruptionError` stamped with `archive=/member=/format=` context, from both `open()` and `extract()`.

## Note on defense-in-depth

The overlap guard itself lives in stdlib (present in all supported CPython ≥ 3.11); this change is about surfacing its verdict as a proper typed error. The **format-agnostic** backstop — the extraction `BombTracker` (per-member ratio, archive-wide static/live ratio, cumulative `max_extracted_bytes`) — remains the primary defense and already stops a plain non-overlapping deep-ratio bomb regardless of format.

## Tests

`tests/test_zip.py` gains `test_overlapping_entries_bomb_translated_to_corruption`, which crafts a Fifield-style overlap archive (adjacent tiny local headers over one shared compressed kernel) and asserts every overlapping member raises a translated `CorruptionError` carrying the stdlib `BadZipFile` as its `__cause__`. It runs in the zero-dep core leg too (stdlib `zipfile` only).

Green in all three dependency configurations (`[all]`, `[all-lowest]`, `[core-only]`); Pyrefly, ty, and ruff clean.

https://claude.ai/code/session_018hQi3WhKr3LnvViH9smD3j

---
_Generated by [Claude Code](https://claude.ai/code/session_018hQi3WhKr3LnvViH9smD3j)_

### DIFF
--- a/src/archivey/internal/backends/zip_reader.py
+++ b/src/archivey/internal/backends/zip_reader.py
@@ -344,10 +344,23 @@ class ZipReader(BaseArchiveReader):
         # — and a duplicate member name can't resolve to the wrong entry.
         info = member._raw
         assert isinstance(info, zipfile.ZipInfo), "ZIP member is missing its ZipInfo handle"
-        raw = cast(
-            "BinaryIO",
-            self._archive.open(info, pwd=self._password),
-        )
+        # zipfile validates the local header when opening a member — the "Overlapped
+        # entries" zip-bomb guard, the name-mismatch check, a missing-password
+        # RuntimeError, and unsupported-compression NotImplementedError all fire here,
+        # before any byte flows through the ArchiveStream translator. Translate them at
+        # the source so callers only ever see ArchiveyErrors (a genuine OSError from the
+        # underlying source is not in the caught set, so it propagates unchanged).
+        try:
+            raw = cast(
+                "BinaryIO",
+                self._archive.open(info, pwd=self._password),
+            )
+        except (zipfile.BadZipFile, RuntimeError, io.UnsupportedOperation) as exc:
+            translated = self._translate_exception(exc)
+            if translated is not None:
+                self._stamp_error_context(translated, member.name)
+                raise translated from exc
+            raise
         return self._wrap_member_stream(raw, member.name, size=member.size)
 
     def _get_archive_info(self) -> ArchiveInfo:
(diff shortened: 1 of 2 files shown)


src/archivey/internal/backends/zip_reader.py
```diff
@@ -344,10 +344,23 @@ def _open_member(self, member: ArchiveMember) -> BinaryIO:
...
+        # RuntimeError, and unsupported-compression NotImplementedError all fire here,
+        # before any byte flows through the ArchiveStream translator. Translate them at
+        # the source so callers only ever see ArchiveyErrors (a genuine OSError from the
+        # underlying source is not in the caught set, so it propagates unchanged).
+        try:
+            raw = cast(
+                "BinaryIO",
+                self._archive.open(info, pwd=self._password),
+            )
+        except (zipfile.BadZipFile, RuntimeError, io.UnsupportedOperation) as exc:
```
davitf, line 358: in this case would it make sense to catch Exception here? (I know it's bad in general, but the behavior here is translate-or-raise, and the translator will already ignore any other unknown exception type)

PR #36, opened 2026-07-04
openspec: fold 2026-07 pre-Phase-5 review outcomes into the phase-5 change

davitf:
Folds the accepted outcomes of the 2026-07 pre-Phase-5 model reviews into the `phase-5-public-api` change, plus five mechanical corrections to main specs that contradicted each other or the implementation. Docs/spec-only — no code changes.

## Phase-5 change (`openspec/changes/phase-5-public-api/`)

- **Per-call limits override**: `extract()` / `extract_all()` gain `limits: ExtractionLimits | None = None` (precedence: per-call > `config.extraction_limits` > default), replacing the four loose Phase-4b bomb-limit kwargs. `ExtractionLimits.UNLIMITED` preset (fields widen to `| None`; `None` disables a guard). Presets are deliberately not trust-named and not coupled to `ExtractionPolicy` — rationale in design.md. This resolves the reviewer conflict ("one knob location" vs. "per-call override for the security knob") with one structured type and two supply points.
- **`PasswordRequest` context object**: the provider callable becomes `Callable[[PasswordRequest], str | bytes | None]` with `member: ArchiveMember | None` and `attempt: int`, so an interactive UI can distinguish a first ask from a wrong-password retry, and the shape can grow without breaking a bare callable signature.
- **`extract()` parity with `open_archive()`**: `source` widens to the multi-volume `Sequence[...]` form and `encoding` is added (MODIFIED "One-Shot Extraction API" delta).
- **Link-resolution sweep** (MODIFIED "Transparent link following" delta): hardlinks resolve positionally (latest occurrence strictly before the link; forward fallback only for malformed link-before-source archives, which extraction's orphan second pass already recovers); symlinks keep last-wins-overall; cycle detection tracks member ids, not names; streaming and random-access modes must agree on duplicate names. Includes the concrete duplicate-name scenarios.
- **New `format-tar` delta**: the truncation requirement's `strict_eof` references become `config.strict_archive_eof` (the rename previously had no format-tar delta).
- **Tasks added**: per-call limits wiring/tests, `PasswordRequest` attempt-count tests, positional-resolution + false-cycle tests, porting the chained-symlink attack test from `_dev_oracle`, a cyclic-symlink `open()` test, a defensive raise for FILE members with `stream=None`, and top-level exports (`MemberSelector`, `MemberFilter`, `ArchiveyConfig`, `ExtractionLimits`, `PasswordRequest`, `PasswordProvider`).
- **Two open maintainer decisions recorded** in proposal.md rather than silently resolved:
  1. `max_entries` counting semantics — selector-skipped members don't count today while filter-skipped ones do; recommendation: count only members actually written.
  2. ZIP `format_availability` reports the intended post-Phase-7 composition, not current read capability (deflate64/PPMd fail at read even when installed); recommendation: report current truth (`PARTIAL` + note) until the Phase 7 codec bypass.

## Main-spec corrections (stale/contradictory text; no behavior change)

- `testing-contract`: the non-seekable ZIP scenario expected a successful read, contradicting `format-zip` / `access-mode-and-cost`; it now expects the `StreamNotSeekableError` fail-fast, and seek-requiring backends are explicitly tested to fail fast.
- `format-7z` / `format-rar`: `is_solid` attributed to `ArchiveInfo`, not `CostReceipt` (matches `access-mode-and-cost` and `cost.py`).
- `archive-reading` + `ARCHITECTURE.md`: link-following pseudocode drops the nonexistent `self[member]` indexing in favor of `get()` (there is deliberately no `__getitem__`).
- `safe-extraction`: the `SANITIZE` policy is relabeled post-v1 (Phase 5 explicitly defers it).
- `backend-registry`: `MagicSignature.weak` removed (content probes replaced weak magic); `CONTENT_PROBES` aligned with the implemented `(format, probe)` shape.

## Validation

`openspec validate --all`: 17 passed / 9 failed — the identical 9 pre-existing failures as on `main` (verified by stashing; they're missing-SHALL issues in requirements untouched here). `change/phase-5-public-api` and every spec edited in this PR validate clean. Markdown-only diff, so the three-configuration test matrix does not apply.

https://claude.ai/code/session_013uqPYhAnGMi6XoovksoLNT

---
_Generated by [Claude Code](https://claude.ai/code/session_013uqPYhAnGMi6XoovksoLNT)_

### DIFF
--- a/openspec/changes/phase-5-public-api/specs/archive-reading/spec.md
+++ b/openspec/changes/phase-5-public-api/specs/archive-reading/spec.md
@@ -49,6 +49,119 @@ Phase 4 `strict_eof` keyword is removed — end-of-archive strictness lives at
 - **WHEN** `archivey.open_archive(source, password="secret")` is called
 - **THEN** the returned `ArchiveReader` uses the provided password for encrypted members
 
+### Requirement: Transparent link following
+
+The system SHALL transparently follow symlinks and hardlinks in `open()` and `read()`. If `member.type` is `SYMLINK` or `HARDLINK`, the call is redirected to the target member. This behavior is format-independent and is implemented once in the `ArchiveReader` ABC. The resolved target, when known, is also exposed as `member.link_target_member`.
+
+**Hardlinks resolve positionally**: the target is the most recent occurrence of the
+target name **strictly before** the link in archive order (this is the TAR model — every
+real tar writer stores the data-bearing entry before the link entries that reference it,
+because hardlinks are detected by inode during archiving; RAR5's redirect model is the
+same). With duplicate names this matches what a sequential extraction would link
+against on disk at the moment the link is written. An archive whose hardlink source
+appears **only later** in archive order is malformed but tolerated: in random-access
+mode resolution falls back to the later member (and extraction recovers it — see
+`format-tar`'s orphan second pass); in streaming mode a forward-pointing hardlink
+cannot be resolved in a single pass and fails per `OnError`.
+
+**Symlinks resolve to the last occurrence overall** of the target name (random-access
+mode): a symlink is a *name*, resolved at use time, and the final on-disk state of a
+duplicated name after sequential extraction is its last occurrence. In streaming mode a
+symlink can only resolve to the latest occurrence seen so far; a forward-pointing
+symlink stays unresolved (`link_target_member` is `None`). The two modes SHALL agree on
+hardlink resolution for the same archive; the symlink forward-visibility difference is
+inherent to a single pass and is documented.
+
+**Target-name resolution.** The stored target string is resolved to an archive-namespace
+member name before lookup, because the two link kinds store targets in different
+namespaces: a **hardlink** target is archive-relative from the root (the linkname is the
+source member's own stored path) and is normalized as-is, while a **symlink** target is
+a filesystem path relative to the link's *own directory* (`dir/link -> file` means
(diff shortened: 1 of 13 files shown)


openspec/changes/phase-5-public-api/specs/archive-reading/spec.md
```diff
@@ -49,28 +49,154 @@ Phase 4 `strict_eof` keyword is removed — end-of-archive strictness lives at
...
+def open(self, member: str | ArchiveMember, _seen: frozenset[int] = frozenset()) -> BinaryIO:
+    if isinstance(member, str):
+        found = self.get(member)  # name lookup — there is no __getitem__ on the reader
+        if found is None:
+            raise KeyError(f"Member {member!r} not found")
+        member = found
+    if member.type in (MemberType.SYMLINK, MemberType.HARDLINK) and member.link_target:
+        if member.member_id in _seen:
+            raise ReadError(f"Link cycle detected at '{member.name}'")
+        target = member.link_target_member or self.get(member.link_target)
```
davitf: does this code follow the spec properly? 

shouldn't we rely on the existing plumbing (e.g. BaseArchiveReader._resolve_link) here? (although reading it, I didn't clearly see where it ensures that hardlinks get an earlier member id)

otherwise, don't we need different behaviors for hard and sym links in get()? isn't link_target relative for symlinks and so we need to resolve it, and for hardlinks we need an exact name match + earlier ID?

PR #37, opened 2026-07-04
Hardlink resolution, streaming forward-pass contract, and members() on pipes

davitf:
## Summary

- Implement positional hardlink resolution and unified link-target lookup in `BaseArchiveReader` (`_lookup_hardlink_target`, `_resolve_link`, `_register_progressively`), with eager registration reusing the progressive path plus a full link-resolution pass.
- Enforce a single shared forward pass on `streaming=True` readers across `__iter__()`, `stream_members()`, `extract_all()`, and `members()`; abandoned or duplicate passes raise `UnsupportedOperationError`.
- Allow `members()` on streaming readers as a list-only scan that always consumes the pass (even on indexed backends); repeat `members()` returns from cache, but blocks subsequent iteration. `get_members_if_available()` remains a no-scan peek that does not consume the pass.
- Update authoritative specs (`archive-reading`, `access-mode-and-cost`, `archive-data-model`), `ARCHITECTURE.md`, `SPEC.md`, and add contract tests for interrupted/partial iteration scenarios.

## Test plan

- [x] `uv run --no-sync pytest tests/test_reader_contract.py`
- [x] `uv run --no-sync pytest tests/test_tar.py -k "link or symlink or hardlink or streaming"`
- [ ] Full three-config push gate per `CONTRIBUTING.md` before merge

### DIFF
--- a/tests/test_tar.py
+++ b/tests/test_tar.py
@@ -24,6 +24,7 @@ from archivey import (
 from archivey.cost import AccessCost, ListingCost, StreamCapability
 from archivey.exceptions import (
     CorruptionError,
+    ReadError,
     StreamNotSeekableError,
     TruncatedError,
 )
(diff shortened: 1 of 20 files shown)


tests/test_zip.py
```diff
@@ -172,13 +172,58 @@ def test_symlink_member(tmp_path: Path) -> None:
...
+    with open_archive(path) as ar:
+        archive = ar._archive  # type: ignore[attr-defined]
+        opens: list[str] = []
+        original_open = archive.open
+
+        def counting_open(info, pwd=None):
+            opens.append(info.filename)
+            return original_open(info, pwd=pwd)
+
+        monkeypatch.setattr(archive, "open", counting_open)
```
davitf, line 209: do we need this `open`-patching thing? is it relevant in this test?

tests/test_tar.py
```diff
@@ -662,3 +664,285 @@ def test_streaming_pass_resolves_backward_links() -> None:
...
+        assert collected[0].link_target_member.name == "target.txt"
+
+
+def test_scan_members_finishes_interrupted_pass(tmp_path: Path) -> None:
+    source = NonSeekableBytesIO(_build_forward_symlink_tar())
+    with open_archive(source, format=ArchiveFormat.TAR, streaming=True) as ar:
+        for member in ar:
+            if member.name == "forward_link":
+                break
+        members = ar.scan_members()
```
davitf, line 805: I'd check that members includes the two expected members exactly, just in case

PR #51, opened 2026-07-09
Add SharedSource primitive + concurrent-open contract (shared-source-streams)

davitf:
## Summary

Implements the OpenSpec change `shared-source-streams` — the Phase 6 entry gate for concurrent-open member streams.

- **`SharedSource`** in `streamtools`: locked, per-view-position byte-range views over one seekable source (stdlib-shaped errors; dormant `independent_handles` seam for later parallel I/O). Views are deliberately unbuffered (consumers already read in large chunks).
- **single-file**: seekable *stream* sources route through `SharedSource.view(...)`; path sources keep independent codec FDs (rapidgzip truncation backstop intact). Removes `_first_stream` so `_open_member` is reentrant. `_close_archive` deliberately does **not** close the non-owning `SharedSource`: member streams stay readable after reader close (consistent with ZIP/path sources), and killing the source under a live rapidgzip stream **aborts the process** — rapidgzip 0.16 raises C++ `std::invalid_argument` through `terminate()` when a Python-source callback fails, on read/close/GC-guard alike (documented as known-issues **Bug 3**; the caller-closes-own-source exposure predates this change).
- **ZIP**: no code change for either source kind — stdlib `zipfile` coordinates a passed-in stream exactly as it does its own path handle (`_SharedFile` keeps a per-open position, re-seeks under `ZipFile._lock`), so a SharedSource layer would duplicate that coordination. Concurrent-open tests lock both legs in.
- View seeks match `BytesIO` exactly: `SEEK_CUR`/`SEEK_END` underflow clamps to the origin (ZipFile-style EOCD probes on truncated sources get typed `BadZipFile`/`CorruptionError`, not a raw `ValueError`); a negative `SEEK_SET` raises.
- **Typed read-after-close** (the delta's "fail loudly" SHALL, its own scenario wording: reader *and* underlying source closed): an inner stream failing with the stdlib closed-handle `ValueError` is mapped by `ArchiveStream` to `UnsupportedOperationError` *before* the per-library translator, so ZIP's generic `ValueError→CorruptionError` rule can't mislabel it and single-file doesn't leak the primitive's raw `ValueError`. Covered by reader-surface tests (single-file and ZIP) plus an `ArchiveStream` precedence unit test.
- Tests: interleaved/thread/misuse primitive coverage; concurrent-open for ZIP (path + stream) and single-file; crash-probe-verified that reader close with live member streams no longer aborts under the accelerator.

TAR-RA remains carved out (single shared decoder); ISO untouched (pycdlib-owned addressing). No `packaging-and-extras` delta.

## Test plan

- [x] `openspec validate --strict shared-source-streams`
- [x] Pyrefly + ty + ruff clean
- [x] Full suite `[all]`: 1079 passed, 99 skipped
- [x] Full suite `[all-lowest]`: 1079 passed, 99 skipped
- [x] Full suite `[core-only]`: 967 passed, 211 skipped (+ zero-dep core check)
- [x] Subprocess crash probes: explicit-close / read-then-close / cycle-GC of a live member stream after `reader.close()` all exit 0 (previously exit 134 / SIGABRT with rapidgzip)

<a href="https://cursor.com/agents/bc-88eb73c4-8801-470d-8a0a-027ef18b1c06" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Open in Web" width="114" height="28" src="https://cursor.com/assets/images/open-in-web-dark.png"&gt;``</a> <a href="https://cursor.com/background-agent?bcId=bc-88eb73c4-8801-470d-8a0a-027ef18b1c06" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Open in Cursor" width="131" height="28" src="https://cursor.com/assets/images/open-in-cursor-dark.png"&gt;``</a>

### DIFF
--- a/src/archivey/internal/streams/streamtools/shared.py
+++ b/src/archivey/internal/streams/streamtools/shared.py
@@ -1,0 +1,232 @@
+"""``SharedSource`` — concurrent-safe byte-range views over one underlying source.
+
+A source (an OS file handle or a seekable ``BinaryIO``) has exactly one file position.
+Two consumers that each ``seek``+``read`` the same handle will clobber each other's
+offset — even on a single thread. :class:`SharedSource` mints independent, seekable,
+**non-owning** views over ``[start, start+length)``; each view keeps its own ``_pos``,
+and every read re-seeks the underlying to that absolute position under a shared lock so
+the seek+read pair is atomic.
+
+This is the streamtools analogue of stdlib ``zipfile._SharedFile``. It is deliberately
+archivey-dependency-free: it raises stdlib-shaped errors (``ValueError`` / ``OSError`` /
+``io.UnsupportedOperation``), never ``archivey.exceptions``.
+
+**Path-source independent handles (dormant).** For a path source, ``view()`` *can* mint a
+fresh ``open(path, 'rb')`` handle per view for true parallel I/O (see
+``independent_handles``). That seam is **default off** for now: every view shares one
+handle + lock. Engaging live per-view handles belongs with parallel extraction.
+"""
+
+from __future__ import annotations
+
+import io
+import os
+import threading
+from pathlib import Path
+from typing import BinaryIO
+
+from archivey.internal.streams.streamtools.base import ReadOnlyIOStream
+from archivey.internal.streams.streamtools.binaryio import (
+    is_filename,
+    is_seekable,
+    source_byte_size,
+)
+
+
+class SharedSource:
+    """Factory for locked, per-view-position slices over one seekable source.
+
+    Construct from a :class:`~pathlib.Path` (opens and owns the handle) or an already-open
+    seekable ``BinaryIO`` (does **not** take ownership — the caller closes it).
+    """
+
+    def __init__(
+        self,
+        source: Path | BinaryIO,
+        *,
+        independent_handles: bool = False,
+    ) -> None:
+        # Seam for true parallel I/O on path sources (fresh FD per view). Dormant:
+        # ignored for now; every view shares ``_handle`` + ``_lock``. See module docstring.
+        self._independent_handles = independent_handles
+
+        self._lock = threading.Lock()
+        self._closed = False
+        self._path: Path | None = None
+        self._owns_handle = False
+
+        if isinstance(source, Path) or is_filename(source):
+            # ``is_filename`` admits bytes; Path wants str/PathLike[str], so fsdecode.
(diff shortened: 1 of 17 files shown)


src/archivey/internal/streams/streamtools/shared.py
```diff
@@ -0,0 +1,240 @@
...
+        size = self._size
+        if size is not None:
+            if start > size:
+                raise ValueError(
+                    f"view start {start} is past the end of the source (size {size})"
+                )
+            if length is not None and start + length > size:
+                raise ValueError(
+                    f"view [{start}, {start + length}) exceeds source size {size}"
+                )
```
davitf: I'm wondering if we should actually have these conditions. If a backend uses these to open the compressed stream for a member based on offset+len from the central directory, but the archive happens to be truncated, these would fail fast instead of letting the backend read what's available.

Should we instead set length to the actual available size if it's less than what was requested? that's how a real stream behaves, right?

src/archivey/internal/streams/streamtools/shared.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __exit__(self, *args: object) -> None:
+        self.close()
+
+    def _check_open(self) -> None:
+        if self._closed:
+            raise ValueError("I/O operation on closed file.")
+
+
+class _SharedSourceView(ReadOnlyIOStream):
```
davitf: This is essentially a clone of SlicingStream but with locks, right? could we merge them somehow? (e.g. by making the lock optional and using nullcontext if it's None) or are there more subtle changes?

src/archivey/internal/streams/streamtools/slice.py
```diff
@@ -61,6 +86,15 @@ def __init__(
...
         self._pos = 0  # position relative to the start of the slice
 
+    def _check_open(self) -> None:
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
+        if self._check_open_fn is not None:
+            self._check_open_fn()
+
+    def _io_guard(self) -> ContextManager[object]:
+        return self._lock if self._lock is not None else nullcontext()
```
davitf: Could this be a field computed at init time (it reuses a single nullcontext if there's no lock)? We wouldn't even need a _lock field. Although we need to know whether to seek, so maybe another var for that will make it more explicit (and easier to separate later if we might want this behavior even without a lock):

```
    self._io_guard = lock if lock is not None else nullcontext()
    self._seek_before_read = lock is not None
```

PR #52, opened 2026-07-10
Declare member-stream capabilities (concurrent + seekable) for readers

davitf:
## Summary

Archives the parallel-reader exploration and proposes **declared member-stream capabilities** — the third and final revision of the concurrent-streams contract, decided in maintainer review (superseding both the `allow_multiple_open_streams` opt-in draft and the no-flag unconditional draft):

```python
open_archive(src)                                               # forward-only, single stream
open_archive(src, member_streams=MemberStreams.SEEKABLE)        # e.g. parquet-in-zip
open_archive(src, member_streams=MemberStreams.CONCURRENT)      # worker fan-out
open_archive(src, member_streams=MemberStreams.CONCURRENT | MemberStreams.SEEKABLE)
```

- **Default contract, uniform on every format (directory included):** forward-only streams, at most one live member stream, no locks, no seek machinery. A second overlapping open raises `ConcurrentAccessError` naming the `open_archive()` call site; `seek()` raises `io.UnsupportedOperation`. Fail-fast in development regardless of the developer's test corpus; pre-1.0, the strict default is the reversible choice.
- **New usage-error hierarchy:** `ArchiveyUsageError` is deliberately *not* an `ArchiveyError`, so `except ArchiveyError` around archive handling never swallows caller bugs. `ConcurrentAccessError` is its first subclass; detected misuse (single-owner overlap, post-close use, provider reentry, early-closed caller source) moves there.
- **`SEEKABLE` is opt-in everywhere**, including the single-stream `open_stream()` API (`seekable: bool = False`): one rule — no archivey stream is seekable unless asked. XZ/lzip index parsing and the rapidgzip `AUTO` resolution become demand-driven; undeclared streams build no seek-point tables.
- **Scope honesty:** the gate covers stream *capabilities*; what a caller's member open *order* costs on a solid archive remains governed by `AccessCost`/`stream_members()`, stated explicitly.
- **All correctness machinery survives under the gate:** post-materialization worker seam, operation-owner tokens/child scopes, stream leases and close/finalizer behavior, password/provider synchronization, callback lock rules, TAR/ISO comprehensive shared-handle locking (`tar-concurrent-open`, instantiated only for `CONCURRENT` readers), and required free-threaded `3.13t` coverage. The undeclared default path takes none of those locks.
- The directory reader's deliberate uniform strictness is recorded as a standing principle (`format-directory` delta).

## Scope

This PR contains exploration archive, contracts, and implementation plans. The runtime implementation remains a follow-up; the API shape lands whole, with the `SEEKABLE` machinery flip sequenced after the `CONCURRENT` gate.

`openspec validate --strict` clean on both changes.

<a href="https://cursor.com/agents/bc-88eb73c4-8801-470d-8a0a-027ef18b1c06" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Open in Web" width="114" height="28" src="https://cursor.com/assets/images/open-in-web-dark.png"&gt;``</a> <a href="https://cursor.com/background-agent?bcId=bc-88eb73c4-8801-470d-8a0a-027ef18b1c06" rel="nofollow noreferrer noopener" target="_blank">``&lt;img alt="Open in Cursor" width="131" height="28" src="https://cursor.com/assets/images/open-in-cursor-dark.png"&gt;``</a>


docs/parallel-reader.md
```diff
@@ -0,0 +1,178 @@
...
+| **ZIP** | RA, independent | **Yes** | `_open_zip_entry` → `ZipFile.open(info)`; stdlib `_SharedFile` coordinates the handle (path *and* stream sources). No per-open scratch on `ZipReader`. Outside archivey SharedSource retrofit (library-owned addressing). |
+| **single-file** | RA when seekable | **Yes** (post-`shared-source-streams`) | Seekable stream: `SharedSource.view(0)` + fresh codec per open. Path: independent codec FD per open. `_first_stream` scratch **removed**. Non-seekable: one-shot `_pending_stream` (streaming / single-pass — out of invariant scope). |
+| **TAR-RA** | RA, **single decoder** | **Exempt** | One `self._tar` (`tarfile.TarFile`); `extractfile` is not safe for interleaved concurrent opens. Carved out by the concurrent-open SHALL and by this invariant. |
+| **ISO** | RA, library-owned | **Yes** (archivey state) | `pycdlib.open_file_from_iso` owns seeking on the shared ISO handle. No per-open scratch on `IsoReader`. Outside SharedSource retrofit (same shape as ZIP / stdlib); not listed as non-compliant. |
```
davitf: I've checked the pycdlib code. It seems to ensure the seek-before-read behavior but doesn't have a lock. Is that the same behavior as tar?

If so: we could have a simple LockedReadStream (better names possible) wrapper that just wraps the opened stream and locks before any read operation that should be atomic, then we use that with a per-archive lock. Would it solve the issue in both tar and iso? We could use it only if the archive was opened in the concurrent-opens mode.

openspec/changes/concurrent-open-opt-in/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,230 @@
...
+**solid** format (7z folder, RAR block, compressed TAR's single stream) satisfies the
+requirement by giving each `open()` its own decompressor / logical view over the shared source,
+re-decoding or re-seeking as the *random open on a solid member* scenario already permits —
+which is where the interleaving expense lives, and why the capability is opt-in. Backends whose
+member addressing is owned by an external library that already coordinates the shared handle
+(ISO via `pycdlib`, ZIP via stdlib `zipfile` through its `_SharedFile`) satisfy interleaving
+through that library and are not listed as non-compliant.
+
+**Streaming / non-seekable carve-out.** Forward-only streaming readers (`streaming=True`) are a
+single forward pass and are out of scope (the flag is meaningful only in random-access mode; the
```
davitf: open_archive should raise an error on an invalid combination of arguments (streaming=True with allow_multiple_open_streams=True)

openspec/changes/concurrent-open-opt-in/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,230 @@
...
+"previous (member, stream) is invalid after the iterator advances" rule of `stream_members` is a
+separate contract). A random-access TAR whose uncompressed stream is **not** seekable does not
+support interleaved opens until a seekable layer exists; that is a documented limitation, not a
+license to skip the shared-source view when the stream *is* seekable.
+
+**Single-reader guarantee only.** The reader object itself remains **not thread-safe**: it MUST
+NOT be driven (concurrent `open()`, iteration, or `close()`) from multiple threads. That misuse
+is **unsupported and undefined** — the reader holds no lock for it. The interleaving guarantee
+above is confined to already-opened member streams used from one thread.
```
davitf: why can't we make it safe? this non-guarantee makes it hard to provide a single reader to multiple workers to let them open members at will. how hard would it be to make the reader thread-safe, especially if we already do the member listing upfront?

openspec/changes/concurrent-open-opt-in/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,230 @@
...
+- **WHEN** a member stream is read after its reader (and the underlying source) has been closed
+- **THEN** a typed error is raised at the reader surface rather than returning arbitrary or empty bytes
+
+### Requirement: Random-access member-open is reentrant and reader-state-free
+
+The system SHALL require that, for a **random-access backend that advertises independent member
+open** (`streaming=False`, serving members by byte-range access — e.g. the native 7z/RAR
+readers, ZIP, single-file, and random-access TAR when the uncompressed stream is seekable), the
+member-open implementation (`_open_member`) is a function of the member and the archive's shared
+source only: it MUST NOT mutate shared reader state, and it MUST NOT keep per-open scratch on the
```
davitf: this restriction seems too narrow. what we actually want is no *unsafe* mutations; any mutation (e.g. the lifecycle bookkeeping mentioned below) must be properly protected. do you agree? what is this designed to prevent?

openspec/changes/concurrent-open-opt-in/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,230 @@
...
+re-decoding or re-seeking as the *random open on a solid member* scenario already permits —
+which is where the interleaving expense lives, and why the capability is opt-in. Backends whose
+member addressing is owned by an external library that already coordinates the shared handle
+(ISO via `pycdlib`, ZIP via stdlib `zipfile` through its `_SharedFile`) satisfy interleaving
+through that library and are not listed as non-compliant.
+
+**Streaming / non-seekable carve-out.** Forward-only streaming readers (`streaming=True`) are a
+single forward pass and are out of scope (the flag is meaningful only in random-access mode; the
+"previous (member, stream) is invalid after the iterator advances" rule of `stream_members` is a
+separate contract). A random-access TAR whose uncompressed stream is **not** seekable does not
```
davitf: Should we change that behavior to be the same as this one? My first instinct is that it seems like a safer approach (no read-time errors), and would be good for consistency. But on the other hand, it makes simple code like this dangerous:
```
for member, stream in reader.stream_members():
    # ignore stream completely
```

if the GC doesn't reclaim and auto-close the stream before the next member is iterated, that would crash.

So maybe the current stream_members() stream behavior is the actual best, but we should document the rationale. WDYT?

PR #53, opened 2026-07-10
Fix multi-password ZIP validation without duplicate decompression

davitf:
## Problem

Traditional **ZipCrypto** validates only a single verification byte at `open()` time (~1/256 wrong passwords pass). The authoritative check is the member CRC (and, for compressed members, the decompressor), which `zipfile` defers to read time.

With several candidate passwords, accepting on the weak byte alone let a wrong candidate shadow the right one and surface a spurious `CorruptionError` — the root cause of flaky CI on the `encrypted-multi` corpus entry.

## Fix

Bounded multi-candidate confirmation (OpenSpec change `zip-multipassword-disambiguation`, archived):

- **Compressed (DEFLATE / BZIP2 / LZMA):** confirm with a bounded decompressed prefix (**64 KiB** via `read_exact`), discard, re-open fresh for the caller.
- **STORED:** one **shared CRC pass** over a `SlicingStream` of the ciphertext body for weak-check survivors (parallel ZipCrypto keystreams, constant memory); winner by central-directory CRC; ties by candidate order; fresh re-open. Compressibility / magic probes were investigated and **rejected**.
- Helpers: `internal/zipcrypto.py`, `internal/password_confirm.py`.
- Specs synced into main `archive-reading` and `format-zip`; change at `openspec/changes/archive/2026-07-11-zip-multipassword-disambiguation/`.
- `open_archive` documents the rare full STORED scan when multiple passwords collide on the weak byte.

Single-candidate fast path unchanged. Candidate-rejection exception scoping unchanged (CRC-message `BadZipFile`, `zlib.error`, `LZMAError`, bz2’s exact `OSError`; structural failures stay `CorruptionError`).

## Tests

- `tests/zipcrypto.py` — minimal ZipCrypto writer + verification-byte collision finder
- `tests/test_zip_multipassword.py` — false-accept scenarios for DEFLATED and STORED
- Exploration scripts under `scripts/exploration/`

Folds #56. Rebased onto current `main`.


src/archivey/internal/backends/zip_reader.py
```diff
@@ -367,6 +413,92 @@ def _to_member(self, info: zipfile.ZipInfo) -> ArchiveMember:
...
+        return getattr(self._archive, "_lock")
+
+    @contextmanager
+    def _ciphertext_chunks(
+        self,
+        info: zipfile.ZipInfo,
+        *,
+        body_start: int = 0,
+        body_limit: int | None = None,
+    ) -> IterType[IterType[bytes]]:
```
davitf: Instead of this push-based approach for the test, why don't you creae a SlicingStream over the bytes of the encrypted member, and pass that to parallel_plaintext_crc32()? then there, we can just read bytes from the stream in chunks. seems cleaner than passing an iterator of byte chunks

src/archivey/internal/password_confirm.py
```diff
@@ -0,0 +1,57 @@
...
+            return item
+    return None
+
+
+def crc32_over_chunks(chunks: Iterable[bytes], *, start: int = 0) -> int:
+    """Accumulate CRC-32 over ``chunks`` starting from ``start`` (zlib convention)."""
+    crc = start
+    for chunk in chunks:
+        crc = zlib.crc32(chunk, crc)
+    return crc & 0xFFFFFFFF
```
davitf: is this unused? if so, please remove

src/archivey/internal/password_confirm.py
```diff
@@ -0,0 +1,57 @@
...
+
+import zlib
+from collections.abc import Iterable, Sequence
+from typing import BinaryIO, TypeVar
+
+# Decompressed plaintext budget for compressed-member confirmation. Empirically
+# stdlib DEFLATE/BZIP2/LZMA reject wrong-key garbage within tens of bytes (see
+# openspec/changes/zip-multipassword-disambiguation/design.md); 1 MiB leaves a
+# wide margin and covers typical members exactly (EOF → CRC).
+CONFIRM_PREFIX_BYTES = 1 << 20
```
davitf: I think we can reduce to 64k then. didn't we agree on that?

src/archivey/internal/password_confirm.py
```diff
@@ -0,0 +1,57 @@
...
+
+_T = TypeVar("_T")
+
+
+def read_and_discard(stream: BinaryIO, bound: int, *, chunk_size: int = 65536) -> int:
```
davitf: could we just use read_exact() in binaryio.py instead of adding this function? for 64k, the accumulation cost is negligible

PR #66, opened 2026-07-12
Implement native 7z reader (OpenSpec native-7z-reader)

davitf:
## Summary

Implements OpenSpec change `native-7z-reader`: a zero-dependency native 7z reader (header parse, codecs, AES, multi-volume, anti-items).

## CI fix (this push)

`test_7z_cli_multi_password_archive_roundtrip` failed on Ubuntu/py3.12 when a wrong known-good password decompressed to the expected length and was accepted because folder digests were absent and per-member CRCs were not checked during password confirmation. `archive.read()` also skipped `VerifyingStream` checks (verification only ran on an empty EOF read).

- Confirm encrypted 7z folders against folder or per-member CRC before caching a password
- Verify digests on `VerifyingStream.close()` when the inner stream is already at EOF
- Regression coverage for wrong-candidate CRC rejection and close-after-full-read verification

## Test plan

- [x] `pytest tests/test_sevenzip_reader.py tests/test_codecs.py`
- [ ] Full CI matrix (`[all]`, `[all-lowest]`, `[core-only]`)

<div><a href="https://cursor.com/agents/bc-03a970fe-6c7a-4a85-892c-fb56f9999ed1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-03a970fe-6c7a-4a85-892c-fb56f9999ed1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/archivey/internal/backends/sevenzip_parser.py
+++ b/src/archivey/internal/backends/sevenzip_parser.py
@@ -8,6 +8,7 @@ not import or delegate to ``py7zr``.
 from __future__ import annotations
 
 import io
+import struct
 import zlib
 from collections.abc import Callable
 from dataclasses import dataclass
@@ -114,7 +115,6 @@ class SevenZipFileRecord:
 class SevenZipArchive:
     major_version: int
     minor_version: int
-    after_header: int
     pack_pos: int
     pack_sizes: list[int]
     pack_positions: list[int]
@@ -200,12 +200,13 @@ def parse_sevenzip_archive(
     if _crc32(start_header) != start_header_crc:
         raise CorruptionError("7z signature header CRC mismatch")
 
-    next_header_offset = int.from_bytes(start_header[0:8], "little")
-    next_header_size = int.from_bytes(start_header[8:16], "little")
-    next_header_crc = int.from_bytes(start_header[16:20], "little")
-    after_header = _SIGNATURE_HEADER_SIZE
+    # start_header is 20 bytes: nextHeaderOffset (u64), nextHeaderSize (u64), CRC (u32).
+    next_header_offset, next_header_size, next_header_crc = struct.unpack(
+        "<QQI", start_header
+    )
 
-    fp.seek(after_header + next_header_offset)
+    # Offsets in the header are relative to the end of the 32-byte signature header.
+    fp.seek(_SIGNATURE_HEADER_SIZE + next_header_offset)
     header_data = _read_exact(fp, next_header_size, "7z next header")
     if _crc32(header_data) != next_header_crc:
         raise CorruptionError("7z next header CRC mismatch")
@@ -213,7 +214,6 @@ def parse_sevenzip_archive(
     parsed = _parse_header(
         fp,
         io.BytesIO(header_data),
-        after_header=after_header,
         decode_folder=decode_folder,
     )
 
@@ -224,7 +224,7 @@ def parse_sevenzip_archive(
     num_unpackstreams_folders = streams.num_unpackstreams_folders or []
     unpack_sizes = streams.unpack_sizes or []
     digests = streams.digests or []
-    pack_pos = after_header + streams.pack_pos
+    pack_pos = _SIGNATURE_HEADER_SIZE + streams.pack_pos
 
     files = _map_files_to_folders(
         parsed.files,
@@ -238,7 +238,6 @@ def parse_sevenzip_archive(
     return SevenZipArchive(
         major_version=major_version,
         minor_version=minor_version,
-        after_header=after_header,
         pack_pos=pack_pos,
         pack_sizes=pack_sizes,
         pack_positions=pack_positions,
@@ -288,7 +287,6 @@ def _parse_header(
     archive_fp: BinaryIO,
     buffer: BinaryIO,
     *,
-    after_header: int,
     decode_folder: DecodeFolder,
 ) -> _ParsedHeader:
     prop = _read_property(buffer, "7z header")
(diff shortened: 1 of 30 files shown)


src/archivey/internal/base_reader.py
```diff
@@ -835,22 +825,17 @@ def __iter__(self) -> Iterator[ArchiveMember]:
...
         try:
             # Return a shallow copy so callers cannot mutate the published cache container.
             return list(self._get_members_registered())
         finally:
             self._state.release_pass(token)
 
+    def get_members(self) -> list[ArchiveMember]:
+        """Compatibility alias for callers expecting the v1-style member-list method."""
+        return self.members()
+
```
davitf: why has this been added to this change? we're not aiming for v1 compatibility (that's was only ever in alpha stage)

src/archivey/internal/backends/sevenzip_parser.py
```diff
@@ -0,0 +1,976 @@
...
+    minor_version = signature[7]
+    start_header_crc = int.from_bytes(signature[8:12], "little")
+    start_header = signature[12:32]
+    if _crc32(start_header) != start_header_crc:
+        raise CorruptionError("7z signature header CRC mismatch")
+
+    next_header_offset = int.from_bytes(start_header[0:8], "little")
+    next_header_size = int.from_bytes(start_header[8:16], "little")
+    next_header_crc = int.from_bytes(start_header[16:20], "little")
+    after_header = _SIGNATURE_HEADER_SIZE
```
davitf: this is never changed, should we just use the constant where it's needed? it might be clearer

src/archivey/internal/backends/sevenzip_parser.py
```diff
@@ -0,0 +1,976 @@
...
+    major_version = signature[6]
+    minor_version = signature[7]
+    start_header_crc = int.from_bytes(signature[8:12], "little")
+    start_header = signature[12:32]
+    if _crc32(start_header) != start_header_crc:
+        raise CorruptionError("7z signature header CRC mismatch")
+
+    next_header_offset = int.from_bytes(start_header[0:8], "little")
+    next_header_size = int.from_bytes(start_header[8:16], "little")
+    next_header_crc = int.from_bytes(start_header[16:20], "little")
```
davitf: would it be cleaner to parse this using struct?

src/archivey/internal/backends/sevenzip_parser.py
```diff
@@ -0,0 +1,976 @@
...
+) -> _ParsedHeader:
+    prop = _read_property(buffer, "7z header")
+    if prop == _Property.END:
+        return _ParsedHeader(_StreamsInfo(), [], None, False)
+    if prop == _Property.HEADER:
+        parsed = _parse_plain_header(buffer)
+        parsed.is_header_encrypted = False
+        return parsed
+    if prop != _Property.ENCODED_HEADER:
+        raise CorruptionError(f"Expected 7z HEADER or ENCODED_HEADER, got 0x{prop:02x}")
```
davitf, line 300: In this function I think this early raise made it harder to read. I looked at the _parse_header call below and got the impression it might be a multi-level recursion. if it were scoped inside a "prop == _Property.ENCODED_HEADER" block, it would be easier to understand the intent. Move the exception to a final block instead, like in _parse_plain_header below

src/archivey/internal/backends/sevenzip_reader.py
```diff
@@ -0,0 +1,877 @@
...
+
+
+@dataclass(frozen=True)
+class _MemberRaw:
+    record: SevenZipFileRecord
+    folder_index: int | None
+    file_in_folder: int | None
+
+
+class _LimitedFolderReader(ReadOnlyIOStream):
```
davitf: I'm wondering if it would be possible and worthwhile to use SlicingStream instead of this custom reader. The extra features this class has are:
- optional source closing - it makes sense as a SlicingStream arg
- draining on end. possible as an arg also, but maybe it's not even needed. this is only needed for continuous streaming: I'm wondering how the iterator could skip the remaining bytes in the underlying stream (if it's seekable, just seek to the new position; otherwise, the problem is knowing how many bytes were effectively read from the member stream; maybe SlicingStream could have a tell()-like method that works even if non-seekable, so we can skip the remaining; or if the underlying one has one as well, that also works). One advantage is that we can only drain when actually needed; if the user closes a member stream, we only actually need to skip remaining bytes in the underlying stream when they start to read from a following one.

Unclear how complicated this really is. What's your take?

src/archivey/internal/backends/sevenzip_reader.py
```diff
@@ -0,0 +1,877 @@
...
+
+    def close(self) -> None:
+        if self.closed:
+            return
+        try:
+            self.drain()
+        finally:
+            if self._close_source:
+                self._source.close()
+            super().close()
```
davitf: no need to drain if we're closing the underlying source, right?

src/archivey/internal/backends/sevenzip_reader.py
```diff
@@ -0,0 +1,877 @@
...
+def _filetime_to_datetime(value: int | None) -> datetime | None:
+    if value is None or value == 0:
+        return None
+    try:
+        return datetime.fromtimestamp(
+            value / 10_000_000 - _NTFS_EPOCH_OFFSET,
+            tz=timezone.utc,
+        )
+    except (ValueError, OverflowError, OSError):
+        return None
```
davitf: worth an invalid timestamp diagnostic instead of ignoring silently? and how could this code raise an OSError?

PR #73, opened 2026-07-12
Deep codebase review: reports, questions, and two docstring fixes

davitf:
## What this is

A full-tree deep review. Every source module was read; tests, both type-checkers, the linter, and
coverage were run (baseline is green: **1348 passed / 70 skipped**, pyrefly + ty clean, ruff clean,
87% coverage). Findings are written up under `review/`, grounded in `file:line` citations and marked
**VERIFIED** / **SUSPECTED**.

**Headline:** the codebase is in very good shape — disciplined concurrency model, rigorous
error-translation contract, cleanly factored codec layer, and a safety posture that matches the
marketing. The reports flag the exceptions, most of which are edge cases or gaps rather than live
bugs.

## Code changes (behavioral scope: none beyond docstrings)

Two doc-only fixes where the code was clearly right, one commit each:
- `open_stream` docstring said it "returns the decompressed bytes" — it returns an `ArchiveStream`.
- `ArchiveReader.open` docstring said a foreign-member open raises `ValueError` — it raises
  `ArchiveyUsageError` (confirmed by `test_reader_contract.py`).

Everything heavier (a public-API/behavior change, a concurrency mechanism, or a hostile-input parser)
was **not** applied — it's written up with a recommendation in `review/QUESTIONS.md` instead.

## Top findings

1. **High** — Native 7z parser pre-allocates `num_files` `_FileProps` with no bound → OOM on a
   crafted (valid-CRC) header. Undercuts the "memory-safe hostile-input parsing" claim; fuzzers miss
   it. (`latent-bugs` L1, `unknown-unknowns` U1)
2. **Med** — `BaseException` (Ctrl-C / MemoryError) during materialization wedges the reader:
   misleading error / CV deadlock. (`concurrency` C1)
3. **Med** — ZIP `member.hashes` never populated despite the spec mandating `"crc32"` — breaks the
   founding dedupe use case for the commonest format. (`specs-docs` S1)
4. **Med** — Directory `get_members_if_available()` does a full uncached filesystem walk every call
   (spec says "without scanning") and races its uid/gid caches under free-threading. (`concurrency`
   C2/C3, `specs-docs` S2/S3)
5. **Med** — No benchmark gate, so the solid-block O(n²) re-decode trap is documented but unenforced.
   (`roadmap`)

Plus test-scenario gaps (`tests`), duplication debt (`complexity`), and filesystem-semantics
assumptions (`unknown-unknowns`).

## Decisions requested

`review/QUESTIONS.md` has five numbered questions, each with a recommendation. The four that gate a
follow-up change: broaden materialization cleanup to `BaseException` (Q1), directory cost honesty
(Q2), populate ZIP CRC32 (Q3), bound the 7z parser (Q4).

## Reading order

`review/00-recon-map.md` → `review/SUMMARY.md` → `review/QUESTIONS.md` → the theme files →
`review/FIXES.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_015ZmdhYAvw1X7q4suXLvmWn)_

### DIFF
--- a/review/QUESTIONS.md
+++ b/review/QUESTIONS.md
@@ -1,0 +1,68 @@
+# Questions for the maintainer
+
+Decisions I need before applying the corresponding change. Each has my recommendation and reasoning
+so you can answer in one pass. Numbered for easy reply.
+
+## Q1 — Broaden the materialization state-cleanup to `BaseException`? (concurrency C1 / latent-bugs L2)
+
+`base_reader.py:503` uses `except Exception` around materialization. A `KeyboardInterrupt` /
+`MemoryError` during `_iter_members()` leaves `cache_state = MATERIALIZING` forever — non-concurrent
+readers then raise a misleading error, CONCURRENT waiters deadlock on the CV.
+
+**Recommendation: yes.** Change the *state-cleanup* handler to `except BaseException: self._state.fail_materialization(); raise` (keep re-raising, so the interrupt still propagates). `mark_reader_closed` already handles `BaseException` this way, so this makes the two consistent. It touches a concurrency mechanism, which is why I'm asking rather than applying. I'd pair it with the red test in tests.md T2.
+
+## Q2 — Directory backend: is a filesystem walk an "index" or a "scan"? (concurrency C2/C3, specs-docs S2/S3)
+
+`DirectoryReader` sets `_MEMBER_LIST_UPFRONT = True` and reports `ListingCost.INDEXED`, so
+`get_members_if_available()` does a full uncached `os.scandir` recursion every call — which
+`archive-reading` spec:181 says it must *not* do ("without scanning"). The two specs disagree.
+
+**Recommendation: treat the directory as `REQUIRES_SCANNING` and set `_MEMBER_LIST_UPFRONT = False`.**
+Rationale: (a) it makes `get_members_if_available()` honest (returns `None`, no walk); (b) it removes
+the free-threaded cache race (C2) for free, since the walk then only happens under the materialization
+election; (c) it aligns directory with plain-tar, which already reports `REQUIRES_SCANNING` for the
+same "no O(1) index, walk to enumerate" situation. The cost is that callers lose a cheap directory
+peek — but there was never a cheap peek; it was an O(n) walk mislabeled. Alternatively, keep INDEXED
+but *cache* the first walk and add a lock, and soften the `cost.py` `INDEXED` docstring — more code,
+same honesty problem. I prefer the first option.
+
+## Q3 — Populate ZIP `member.hashes["crc32"]`, and does it trigger VerifyingStream? (specs-docs S1, latent-bugs L3)
+
+archive-data-model spec:193 mandates ZIP CRC32 under `"crc32"`; the backend never populates it,
+breaking the founding dedupe use case for the most common format.
+
+**Recommendation: populate it (`hashes={"crc32": info.CRC}` in `_to_member`), and keep relying on
+zipfile's own CRC check — do NOT route ZIP reads through `VerifyingStream`.** Rationale: zipfile
+already verifies CRC at EOF on every member read, so wrapping in `VerifyingStream` would double-hash
+for no benefit and change the error path. Surfacing the datum (for dedupe/inspection) and keeping the
+existing verification are separable; do the former. Needs a test asserting `reader.get(x).hashes["crc32"]`
+matches, and one confirming read behavior is unchanged. (If you'd rather unify verification across
+backends later, that's a separate change.)
+
+## Q4 — What bound for the 7z parser's count fields? (latent-bugs L1, unknown-unknowns U1, threat-model O1)
+
+`_read_files_info` pre-allocates `num_files` `_FileProps` with no bound → OOM on a crafted header.
+This is threat-model O1 made concrete and undercuts VISION claim #2 in practice.
+
+**Recommendation: bound every count field parsed from the header against the remaining header size,
+plus a hard ceiling.** Concretely: `num_files` (and `num_folders`, timestamp/attribute counts)
+cannot exceed `next_header_size` (each real entry consumes ≥1 byte of header somewhere), so a cheap
+`if num_files > len(header_data): raise CorruptionError(...)` catches the pathological case without a
+magic constant. Add a hard ceiling (e.g. a configurable `max_members`, tying into the "listing limits"
+config the roadmap is missing) for defense in depth. I'd want this before the public release and the
+Atheris gate. The exact ceiling / whether it's config-driven is your call — hence the question.
+
+## Q5 — Is the strict-EOF-under-IGNORE-policy precedence intended? (other O-4)
+
+`diagnostics_collector.py` raises `escalate_as` (TAR strict-EOF `TruncatedError`) even when the
+per-code policy is `IGNORE`, with no diagnostic recorded. Two knobs, undefined interaction.
+
+**Recommendation: keep the current precedence (strict-EOF wins), but spec it and add a test.**
+`strict_archive_eof=True` is an explicit "make this fatal" that should outrank a per-code IGNORE.
+Just make it intentional in the diagnostics spec rather than emergent. Low priority.
+
+---
+
+If you answer Q1/Q3/Q4 with "apply my recommendation," I can implement all three (each with the
+red-green test named in tests.md) in a follow-up. Q2 is the only one that changes a cost value users
+might read, so it's the one I most want your explicit call on.
--- a/review/complexity.md
+++ b/review/complexity.md
@@ -1,0 +1,110 @@
+# Theme 2 — Complexity & duplication
+
+The codebase is, on the whole, well-factored: the `streamtools` base hierarchy
+(`ReadOnlyIOStream`/`DelegatingStream`), the codec-descriptor table, and the
+`SegmentedDecompressorStream` scaffolding all pull real weight and read cleanly. The
+duplication that exists is concentrated in the backends. Concrete consolidations below,
+ranked by payoff.
+
+## X1 — The handle-lock branch is copy-pasted ~15× (medium payoff)
+
+Every CONCURRENT-aware backend repeats this shape:
+
+```python
+if self._handle_lock is not None:
+    with self._handle_lock:
+        result = do_thing()
+else:
+    result = do_thing()
+```
+
+Occurrences: `tar_reader.py` (`_open_tarfile` init, `_iter_members`, `_iter_members_progressive`,
+`_iter_with_data`, `_verify_tar_eof`, `_open_member`, `_close_archive` — 7), `zip_reader.py`
+(`_zip_open_raw`, `_zip_close_raw`, `_close_archive` — 3), `iso_reader.py` (init, `_open_member`,
+`_close_archive` — 3). Each is a place a future edit can forget the lock on one branch.
+
+Sketch — a single helper on `BaseArchiveReader` (or a per-backend mixin), since the lock is
+optional-or-`nullcontext`:
+
+```python
+# base_reader.py
+from contextlib import nullcontext
+
+def _handle_guard(self) -> ContextManager[object]:
+    return self._handle_lock if self._handle_lock is not None else nullcontext()
+```
+
+Then every site collapses to `with self._handle_guard(): ...`. The one case that can't (the TAR
+progressive walk holds the lock only around `next()`, not the yield) stays explicit. This removes
+~12 branch pairs and the "forgot one branch" failure mode entirely.
+
+## X2 — NTFS FILETIME conversion duplicated between ZIP and 7z (medium payoff)
+
+`_NTFS_EPOCH_OFFSET = 11_644_473_600`, `_filetime_to_datetime(...)`, and the `_TimestampIssue`
+dataclass exist in near-identical form in `zip_reader.py:143-174` and `sevenzip_reader.py:140-184`
+(7z's variant takes `int | None`, ZIP's takes `int`; both return `(datetime|None, issue|None)`).
+The FILETIME epoch math and the out-of-range guard are byte-for-byte the same.
+
+Sketch — lift to a shared internal helper (e.g. `internal/timestamps.py`):
+
+```python
+def filetime_to_datetime(value: int | None, filename: str, *, field: str
+    ) -> tuple[datetime | None, TimestampIssue | None]: ...
+```
+
+and have both backends import it. The `_TimestampIssue` shapes differ slightly (ZIP carries
+`source`, 7z hardcodes `"ntfs"`); unify on the ZIP shape with a default. ~40 lines removed,
+one place to fix a timestamp edge case instead of two.
+
+## X3 — ZIP repeats its member-open exception tuple 3× (low-medium payoff)
+
+The tuple `(zipfile.BadZipFile, RuntimeError, io.UnsupportedOperation, NotImplementedError,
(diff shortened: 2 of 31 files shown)


review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+Decisions I need before applying the corresponding change. Each has my recommendation and reasoning
+so you can answer in one pass. Numbered for easy reply.
+
+## Q1 — Broaden the materialization state-cleanup to `BaseException`? (concurrency C1 / latent-bugs L2)
+
+`base_reader.py:503` uses `except Exception` around materialization. A `KeyboardInterrupt` /
+`MemoryError` during `_iter_members()` leaves `cache_state = MATERIALIZING` forever — non-concurrent
+readers then raise a misleading error, CONCURRENT waiters deadlock on the CV.
+
+**Recommendation: yes.** Change the *state-cleanup* handler to `except BaseException: self._state.fail_materialization(); raise` (keep re-raising, so the interrupt still propagates). `mark_reader_closed` already handles `BaseException` this way, so this makes the two consistent. It touches a concurrency mechanism, which is why I'm asking rather than applying. I'd pair it with the red test in tests.md T2.
```
davitf, line 12: Yes, do it. Please comment on the code why we need BaseException there, to avoid someone inadvertently changing it back.

review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+`archive-reading` spec:181 says it must *not* do ("without scanning"). The two specs disagree.
+
+**Recommendation: treat the directory as `REQUIRES_SCANNING` and set `_MEMBER_LIST_UPFRONT = False`.**
+Rationale: (a) it makes `get_members_if_available()` honest (returns `None`, no walk); (b) it removes
+the free-threaded cache race (C2) for free, since the walk then only happens under the materialization
+election; (c) it aligns directory with plain-tar, which already reports `REQUIRES_SCANNING` for the
+same "no O(1) index, walk to enumerate" situation. The cost is that callers lose a cheap directory
+peek — but there was never a cheap peek; it was an O(n) walk mislabeled. Alternatively, keep INDEXED
+but *cache* the first walk and add a lock, and soften the `cost.py` `INDEXED` docstring — more code,
+same honesty problem. I prefer the first option.
```
davitf, line 27: Agree, please do it

review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Q3 — Populate ZIP `member.hashes["crc32"]`, and does it trigger VerifyingStream? (specs-docs S1, latent-bugs L3)
+
+archive-data-model spec:193 mandates ZIP CRC32 under `"crc32"`; the backend never populates it,
+breaking the founding dedupe use case for the most common format.
+
+**Recommendation: populate it (`hashes={"crc32": info.CRC}` in `_to_member`), and keep relying on
```
davitf, line 34: Agree, please do it

review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+`_read_files_info` pre-allocates `num_files` `_FileProps` with no bound → OOM on a crafted header.
+This is threat-model O1 made concrete and undercuts VISION claim #2 in practice.
+
+**Recommendation: bound every count field parsed from the header against the remaining header size,
+plus a hard ceiling.** Concretely: `num_files` (and `num_folders`, timestamp/attribute counts)
+cannot exceed `next_header_size` (each real entry consumes ≥1 byte of header somewhere), so a cheap
+`if num_files > len(header_data): raise CorruptionError(...)` catches the pathological case without a
+magic constant. Add a hard ceiling (e.g. a configurable `max_members`, tying into the "listing limits"
+config the roadmap is missing) for defense in depth. I'd want this before the public release and the
+Atheris gate. The exact ceiling / whether it's config-driven is your call — hence the question.
```
davitf, line 53: Agree. Does it make sense to use the config ExtractionLimits.max_entries as the ceiling?

review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Q5 — Is the strict-EOF-under-IGNORE-policy precedence intended? (other O-4)
+
+`diagnostics_collector.py` raises `escalate_as` (TAR strict-EOF `TruncatedError`) even when the
+per-code policy is `IGNORE`, with no diagnostic recorded. Two knobs, undefined interaction.
+
+**Recommendation: keep the current precedence (strict-EOF wins), but spec it and add a test.**
+`strict_archive_eof=True` is an explicit "make this fatal" that should outrank a per-code IGNORE.
+Just make it intentional in the diagnostics spec rather than emergent. Low priority.
```
davitf, line 62: Agree

review/QUESTIONS.md
```diff
@@ -0,0 +1,68 @@
...
+
+---
+
+If you answer Q1/Q3/Q4 with "apply my recommendation," I can implement all three (each with the
+red-green test named in tests.md) in a follow-up. Q2 is the only one that changes a cost value users
+might read, so it's the one I most want your explicit call on.
```
davitf, line 68: I've agreed with all, including Q2. Please apply them

review/unknown-unknowns.md
```diff
@@ -0,0 +1,89 @@
...
+
+`info.CRC` is on every `ZipInfo` and is thrown away. Populating `member.hashes["crc32"]` (S1) turns
+on cheap dedupe for the most common format at zero read cost — the founding use case. This is the
+single highest-value "we're not exploiting what's already there."
+
+### E2 — Reflink/`copy_file_range` for cross-device hardlink fallback (low)
+
+`_place_link` falls back to `shutil.copy2` on EXDEV (`extraction.py:955`). On Linux with a
+CoW filesystem (btrfs/XFS), `os.copy_file_range` or a reflink (`FICLONE`) would make that fallback
+near-free and space-shared. Minor, but the hardlink-heavy backup use case would benefit.
```
davitf, line 79: Can you make that change please?

review/latent-bugs.md
```diff
@@ -0,0 +1,94 @@
...
+
+### D1 — Private stdlib API dependency: `lzma._decode_filter_properties` (VERIFIED)
+
+`sevenzip_reader.py:198-199` calls `getattr(lzma, "_decode_filter_properties")` — a **private**
+CPython function with no stability guarantee. It's wrapped in `except (AttributeError, ...)` that
+maps failure to `CorruptionError`. So if a future Python renames/removes it, **every** LZMA/LZMA2 7z
+member would be reported as corrupt rather than failing loudly with "archivey needs updating." This
+is the one place the native reader leans on a private stdlib internal; it deserves a
+`hasattr`-at-import assertion (fail loud) rather than a silent per-member CorruptionError, and a
+tracking test that would break on the Python that removes it.
```
davitf, line 73: Agree, please make this change. How does py7zr handle this? If it's easy to replace with a non-private function, it may be worth doing

review/latent-bugs.md
```diff
@@ -0,0 +1,94 @@
...
+
+### D2 — `capture_open_site` retains a full stack snapshot per reader (VERIFIED, minor)
+
+`open_site.py:27` does `traceback.extract_stack()` on **every** `open_archive` and retains the full
+`tuple[FrameSummary, ...]` for the reader's lifetime (for a possible `ConcurrentAccessError`
+breadcrumb that most readers never hit). For the founding use case (opening millions of archives in
+a dedupe sweep) that's a measurable per-open cost and retained memory for a rarely-used diagnostic.
+Consider capturing lazily (only the file:line eagerly, the full stack on demand) or gating it behind
+a config flag.
```
davitf, line 82: Agreed, please propose this change

review/complexity.md
```diff
@@ -0,0 +1,110 @@
+# Theme 2 — Complexity & duplication
+
+The codebase is, on the whole, well-factored: the `streamtools` base hierarchy
+(`ReadOnlyIOStream`/`DelegatingStream`), the codec-descriptor table, and the
+`SegmentedDecompressorStream` scaffolding all pull real weight and read cleanly. The
+duplication that exists is concentrated in the backends. Concrete consolidations below,
+ranked by payoff.
+
+## X1 — The handle-lock branch is copy-pasted ~15× (medium payoff)
```
davitf, line 11: Agreed, can you make this change please?

review/complexity.md
```diff
@@ -0,0 +1,110 @@
...
+
+def _handle_guard(self) -> ContextManager[object]:
+    return self._handle_lock if self._handle_lock is not None else nullcontext()
+```
+
+Then every site collapses to `with self._handle_guard(): ...`. The one case that can't (the TAR
+progressive walk holds the lock only around `next()`, not the yield) stays explicit. This removes
+~12 branch pairs and the "forgot one branch" failure mode entirely.
+
+## X2 — NTFS FILETIME conversion duplicated between ZIP and 7z (medium payoff)
```
davitf, line 43: Agreed, please do it. (especially as rar support might need it as well, right?)

review/complexity.md
```diff
@@ -0,0 +1,110 @@
...
+
+The tuple `(zipfile.BadZipFile, RuntimeError, io.UnsupportedOperation, NotImplementedError,
+zlib.error, lzma.LZMAError, UnicodeDecodeError, ValueError, OSError)` appears verbatim in
+`_open_zipfile_member` (626), `_open_compressed_confirmed` (688-698), and `_ensure_link_target`
+(878-887). A drifting edit to one (adding a newly-observed exception type) silently leaves the
+others narrower.
+
+Sketch — a module constant `_ZIP_MEMBER_READ_ERRORS = (...)` and a small
+`_translate_and_raise(exc, member_name)` helper wrapping the shared "translate → stamp → raise /
+else re-raise" body, which is *also* duplicated at each of those sites.
```
davitf, line 71: Agree, please do it

review/complexity.md
```diff
@@ -0,0 +1,110 @@
...
+iter_candidates()`, then `while has_provider(): ask_provider(...)`, `record_success`, exhaustion
+messaging. It duplicates this because it needs a *batched* disambiguation (one shared CRC pass
+over all weak-check survivors), which `attempt()`'s one-candidate-at-a-time shape can't express.
+
+That's a legitimate reason to diverge, but the result is the single hardest-to-follow method in
+the ZIP backend (I re-read it three times). It would benefit from either (a) a `attempt_batch`
+primitive on `_PasswordCandidates` that takes a `weak_ok` predicate + a `disambiguate(survivors)`
+callback, so the loop lives in one place, or (b) at minimum a comment block naming the three
+phases (collect-survivors / disambiguate / provider-fallback). This is a readability refactor, not
+a bug — flagging for the "clean as you go" backlog, not urgent.
```
davitf, line 86: Please do what you think will make it the easiest to understand!

review/complexity.md
```diff
@@ -0,0 +1,110 @@
...
+
+```python
+if previous is not None:
+    # Do not close here: the caller still holds the last yielded stream ...
+    pass
+```
+
+The `if`/`pass` does nothing; only the comment matters. It could be a bare comment at end of
+function. I left it — it's harmless and the `if` arguably documents the invariant (there *is* a
+dangling `previous`). Noting so it's a conscious choice, not an oversight.
```
davitf, line 100: I think it's worth it replacing it with a simple comment

review/concurrency.md
```diff
@@ -0,0 +1,133 @@
...
+
+Under CONCURRENT, two threads may call `get_members_if_available()` simultaneously (nothing
+serializes it — it takes no pass/worker token, only `require_open`). Both then mutate the same
+`dict` concurrently. Under the GIL this is benign; under free-threading (`3.13t`) concurrent
+`dict.__setitem__` on the same dict is a data race with no lock. The `3.13t` CI job exercises
+concurrent `open`/`read`, but I don't see it exercising concurrent `get_members_if_available()`
+on a directory (see tests.md T3).
+
+This is entangled with a design smell (C3) — the real question is whether directory's
+`get_members_if_available()` should be doing a full walk at all.
```
davitf, line 76: We're changing this behavior as you suggested, so this is now moot, right?

review/other.md
```diff
@@ -0,0 +1,64 @@
...
+
+`iso_reader.py` installs `_install_pycdlib_directory_cycle_guard()` **at module import**
+(`:169`), which does `setattr(pcd_module, "collections", _DequeGuardedCollections(...))`. This is
+process-global and permanent: any *other* code in the same process that uses pycdlib directly now
+gets archivey's guarded `deque` too. It's well-documented and confined to pycdlib's namespace (not a
+global `collections.deque` swap), and it fixes a real infinite-loop-on-hostile-ISO bug. But it's a
+side effect of `import archivey.internal.backends.iso_reader` — which `import archivey` triggers
+eagerly (`__init__.py:195`). A consumer embedding archivey alongside their own pycdlib usage should
+know their pycdlib is being patched. Worth one line in the ISO docs / a note in the threat model.
+Not a bug; a "the maintainer should be aware" item.
```
davitf, line 15: Yes please

review/other.md
```diff
@@ -0,0 +1,64 @@
...
+
+`ArchiveMember` is mutable and `__hash__` raises (types.py:347). The spec says "callers must treat
+as read-only." Late-bound fields (`link_target_member`, `_diagnostics`, `link_target`) are mutated
+by the library *after* the member is handed to the caller (during lazy link resolution / streaming
+finalization). A caller who snapshots `member.link_target` early and one who reads it after a later
+`scan_members()` can see different values for the *same object*. This is documented, but it's the
+kind of thing that bites a user building a cache keyed on member identity. A `member.freeze()` /
+immutable-view helper, or at least a doc example of the "read it when you need it, don't cache
+across passes" rule, would help. (Related: `member_id`/`archive_id` raise `AttributeError` before
+registration — a caller touching them on a hand-built member gets a surprise.)
```
davitf, line 36: What would the proper behavior here? Being unhashable means they can't be added to sets or used as dict keys, right? Maybe we should allow hashing via the immutable fields? Or could that cause nasty surprises for users? I'm honestly in doubt

review/roadmap.md
```diff
@@ -0,0 +1,67 @@
...
+
+- **A resource-limits config for *listing*, not just extraction.** `ExtractionLimits` guards
+  extraction bombs, but there's no equivalent for the *parse/list* phase (max members, max header
+  size, max name length). Threat-model O1 names this; the roadmap doesn't have a concrete change for
+  it. It's the natural home for the L1 fix and should be a named OpenSpec change.
+
+- **A documented policy for `member.hashes` population parity across backends.** 7z populates
+  crc32; ZIP doesn't (specs-docs.md S1); directory/tar don't (no stored digest to surface, fine).
+  "Which backends surface which stored digests, and the recipe for cheap dedupe" deserves to be a
+  first-class doc + a conformance-sweep assertion, since it's the founding use case.
```
davitf, line 63: Agree it should be added. We should also check if single file reader correctly exposes the hashes included in each format

review/roadmap.md
```diff
@@ -0,0 +1,67 @@
...
+
+- **An explicit free-threading support statement.** The `3.13t` job exists but runs core-only.
+  Either commit to "free-threaded support = the core backends, ISO/accelerators GIL-only" as a
+  documented matrix, or extend the job. Right now the support boundary is implicit in a CI flag.
```
davitf, line 67: Should we try to support free-threading with the external backends, or is that impossible until they're also changed to be safe? Is it worth testing what already works? I imagine accelerators might not work as they're too low-level, but ISO and other python libs might work as it's python-only and we already use proper locking

PR #100, opened 2026-07-14
Implement benchmark-gate: structural CI gate + nightly wall drift

davitf:
## Summary

Implements **benchmark-gate**: opt-in measurement counters + a **structural** CI gate (seek baselines / solid decode-once). Wall-time / VISION ≤1.3× is a **manual + nightly non-blocking** drift tool — shared-runner noise makes ratio regression gates flake.

## Review follow-ups

Addressed the review comments:

1. **Deleted `wall_time.json`** — cold-pass ci ratios disagreed with RESULTS; nothing automated read it.
2. **Dropped no-op `unrar` from the benchmark job** — RAR decode-once is unit-tested against committed fixtures (`unrar` only) in `test_measurement.py`.
3. **Added `benchmark-wall`** — nightly / `workflow_dispatch`, `--mode full --scale realistic`, JSON artifact, `continue-on-error`.
4. **Labeled RESULTS** with host (4-core Xeon KVM) and “speedup scales with cores.”
5. **Softened** the common-path “no silent re-decode” claim — seek axis carries that; byte `== unpacked` is tautological at member output.
6. **Noted ISO/directory** out of harness scope (measurement still wired).
7. **Parametrized decode-once unit test** — 7z + RAR5/RAR4 solid fixtures, ×1.1 bound.

## Investigation (ZIP / TAR / accelerators)

Full write-up: [`benchmarks/RESULTS.md`](../blob/cursor/benchmark-gate-06c9/benchmarks/RESULTS.md).

| Case | ratio (4-core host) | note |
| --- | ---: | --- |
| zip_read_all | ~1.18× | wrapper overhead; drifts on quieter runners |
| tar_read_all (plain) | ~1.70× | per-member overhead, not codecs |
| gzip_read_all | ~1.02× | within VISION |
| targz/tarbz2 accel off | ~1.18× / ~1.02× | tracks stdlib |
| targz/tarbz2 accel on | ~0.45× / ~0.35× | core-count dependent |

## What landed

- Opt-in `bytes_decompressed` + `source_seek_count`
- Harness `--scale {ci,realistic}`; wall timing unmeasured + interleaved
- PR: structural gate + JSON artifact; nightly non-blocking wall job
- Decode-once correctness tests for solid 7z + RAR

<div><a href="https://cursor.com/agents/bc-9cf492bd-002a-488f-a8e1-b65d9bfe06c9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9cf492bd-002a-488f-a8e1-b65d9bfe06c9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


benchmarks/baselines/wall_time.json
davitf, line 1: **This committed baseline is both misleading and effectively dead.**

The numbers here (zip `4.7×`, gzip `5.9×`, tar `2.8×`) were produced at **ci scale with a single cold pass** — `--update-baselines` forces `scale=ci`, and `warmup` is only enabled for `realistic` scale, so `pair_wall` took one un-warmed sample per side. That's exactly the methodology `RESULTS.md` debunks (the source of the bogus "faster than zipfile" reading), so the committed ratios don't match the write-up's fair numbers (zip `1.18×`, gzip `1.02×`) at all.

It's also never consulted by automation: `wall_time.json` is only read in `--mode full` (`_wall_checks` → `WALL_BASELINE`), and no workflow runs full mode. So it sits in the repo enshrining the discredited method while gating nothing.

Suggest either deleting it, or regenerating it from a fair (`realistic` + interleaved) run and actually wiring a job to check it — see my note on the missing scheduled full-mode job.

---

.github/workflows/ci.yml
```diff
@@ -186,3 +186,28 @@ jobs:
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+      - name: Set up Python
+        run: uv python install 3.11
+      - name: Sync (dev + all extras)
+        run: uv sync --group dev --extra all
+      - name: Install unrar (solid RAR fixtures when rar builder absent still need reader)
```
davitf: **This `unrar` install is a no-op for RAR coverage.** `unrar` is the *reader*; the solid-RAR fixture is built by `build_solid_rar`, which needs the `rar` **writer** binary (`shutil.which("rar")`). This job never installs `rar`, so `build_solid_rar` returns `None`, no RAR fixture is produced, and the reader is never exercised here — the RAR benchmark cases silently don't run in CI.

Two clean options:
- install the `rar` writer here so the solid-RAR cases actually gate, or
- drop this step and instead cover the RAR decode-once invariant with the **committed** `tests/fixtures/rar/basic_solid__*.rar` fixtures (need only `unrar`) as a plain unit test — see my comment on `tests/test_measurement.py`.

The comment ("solid RAR fixtures when rar builder absent still need reader") describes a path that can't happen: absent the builder there's no fixture to read.

---

.github/workflows/ci.yml
```diff
@@ -186,3 +186,28 @@ jobs:
           uv run --python 3.13t --no-sync
           --with pytest --with pytest-timeout
           pytest -m concurrent_reader -q --timeout=60 -o addopts=
+
+  # Performance budget gate (VISION): structural invariants on every PR (bytes
+  # decompressed / seek counts). Wall-time ratios vs stdlib are noisier on shared
+  # runners — run with --mode full on a schedule or manually.
```
davitf: **The wall-time / VISION budget is never actually gated by automation.** This comment ("run with `--mode full` on a schedule or manually") describes a schedule that doesn't exist — there's no `schedule:` trigger anywhere, and this job runs `--mode structural` only. So the whole point of a "performance budget gate" (the VISION ≤1.3× wall ratio) is manual-only; a wall-time regression is never caught.

That's a defensible choice — I reproduced how noisy these ratios are on a shared runner (zip drifted `1.18×`→`1.55×`, and the accelerator speedups halved vs the recorded figures), which is a good reason to keep wall-time out of a blocking gate. But then the PR title/comment overclaim. I'd either:
- add a real non-blocking `schedule:` (nightly) job that runs `--mode full --scale realistic` and uploads the JSON (`--json-out`) for drift tracking, or
- reword this to say the automated gate covers **structural invariants only** and wall-time is a manual tool.

Minor: consider adding `--json-out` here too and uploading it as an artifact so per-PR structural drift is inspectable.

---

benchmarks/RESULTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+The bzip2 accelerator is **`rapidgzip.IndexedBzip2File`** (via
+`use_indexed_bzip2`), not the separate `indexed_bzip2` package.
+
+| Case | ay | stdlib | vs stdlib | vs accel_off |
+| --- | ---: | ---: | ---: | ---: |
+| targz accel **off** | 21.3 ms | 18.0 ms | 1.18× | — |
+| targz accel **on** | 8.3 ms | 18.2 ms | **0.45×** | **2.57×** faster |
+| tarbz2 accel **off** | 375 ms | 366 ms | 1.02× | — |
+| tarbz2 accel **on** | 126 ms | 365 ms | **0.35×** | **2.97×** faster |
```
davitf, line 96: **These accelerator headline numbers are core-count-dependent and won't reproduce.** rapidgzip parallelises across cores, so the `2.57×` / `2.97×` speedups (and the `0.45×` / `0.35×` vs stdlib) reflect the author's core count. Re-running `--mode full --scale realistic` on a 2-core box I got:

| case | RESULTS | reproduced |
| --- | ---: | ---: |
| targz accel_on vs stdlib | 0.45× | **0.88×** |
| tarbz2 accel_on vs stdlib | 0.35× | **0.50×** |
| zip_read_all | 1.18× | 1.55× |

The *direction* holds everywhere (accel_on beats stdlib, accel_off tracks it), but the specific figures are hardware-specific. Suggest labeling this table with the machine/core count it was measured on and noting "speedup scales with cores," so nobody treats `2.57×` as a portable claim.

---

benchmarks/harness.py
```diff
@@ -0,0 +1,716 @@
...
+        if r.operation == "read_all" and r.unpacked_bytes is not None:
+            # Common-path formats: decompressed bytes should match what was read.
+            if (
+                r.format in ("zip", "tar", "gzip", "tar.gz", "tar.bz2")
+                and r.bytes_decompressed < r.unpacked_bytes
+            ):
+                failures.append(
+                    f"{r.case}: bytes_decompressed={r.bytes_decompressed} "
+                    f"< unpacked={r.unpacked_bytes}"
+                )
```
davitf, line 545: **For common formats this check only catches *under*-decode, so the "no silent re-decode" claim is carried by the seek axis, not this one.** `bytes_decompressed` is counted at the member-output layer (`OutputCountingStream` in `_wrap_member_stream`), so for zip/tar/gzip it equals `unpacked` **by construction** — an internal re-decode of the *source* wouldn't move it. This `< unpacked` guard therefore only fires if a member delivers fewer bytes than it read, never on over-reading.

That's fine as far as it goes, but `RESULTS.md`'s "exact match — no silent re-decode" for ZIP/TAR/gzip is really proven by the **seek-count** invariant (and by `bytes==unpacked` being tautological), not by a byte-over-read check. Where the byte axis does real work is the solid path above (folder-level counting, where re-decode genuinely inflates the total). Worth softening the wording so the two axes' roles are clear. No code change strictly required.

---

benchmarks/harness.py
```diff
@@ -0,0 +1,716 @@
...
+def _rapidgzip_available() -> bool:
+    try:
+        import rapidgzip  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+def run_cases(
```
davitf, line 257: **Coverage gap: ISO is instrumented but never benchmarked.** `iso_reader.py` gained `_track_source_seeks` wiring in this PR, but `run_cases` has no ISO case, so that instrumentation is exercised only by whatever unit tests touch it — not by the gate. If ISO is in scope for the perf budget, add a small ISO fixture + `iso_read_all` case here; if it's deliberately out of scope, a one-line note in the module docstring would save a future reader the hunt. (Same, to a lesser degree, for the directory backend.)

---

tests/test_measurement.py
```diff
@@ -0,0 +1,114 @@
...
+        with open_archive(path) as reader:
+            assert isinstance(reader, BaseArchiveReader)
+            for _member, stream in reader.stream_members():
+                if stream is not None:
+                    stream.read()
+            assert reader.bytes_decompressed == len(payload_a) + len(payload_b)
+            assert reader.source_seek_count > 0
+
+
+def test_solid_7z_sequential_decodes_once(tmp_path: Path) -> None:
```
davitf: **Yes — the no-re-read checks should be first-class unit tests, and this test is the proof they can be.** Decode-once-on-sequential / re-decode-on-random is a deterministic *correctness* property of each backend, not a perf measurement — it doesn't depend on wall time, so it belongs in the unit suite where it runs across the whole `py3.11–3.14 × os × install` matrix (the `test` job already runs `pytest tests/`, so this file already gets that coverage). The benchmark harness should keep the *noisy/version-sensitive* parts (wall ratios, seek-count baselines); the decode-once contract should be pinned here per backend.

Two concrete extensions to this test:

**1. Cover RAR — no builder needed.** The 7z case builds a fixture with `py7zr`, but committed solid RAR fixtures already exist (`tests/fixtures/rar/basic_solid__.rar`, `basic_solid__rar4.rar`) and need only the `unrar` reader. Parametrize so both backends assert the same invariant:

```python
import pytest
from tests.sample_archives import ...  # or point directly at the committed fixture paths

SOLID_CASES = [
    pytest.param("7z", None, id="7z"),          # built below via py7zr
    pytest.param("rar", "tests/fixtures/rar/basic_solid__.rar", id="rar5"),
    pytest.param("rar", "tests/fixtures/rar/basic_solid__rar4.rar", id="rar4"),
]

@pytest.mark.parametrize("fmt,fixture", SOLID_CASES)
def test_solid_sequential_decodes_once(fmt, fixture, tmp_path):
    archive, unpacked = _solid_fixture(fmt, fixture, tmp_path)  # builds 7z / returns committed rar
    with enable_measurement(), open_archive(archive) as reader:
        assert reader.info.is_solid is True
        for _m, s in reader.stream_members():
            if s is not None:
                s.read()
        sequential = reader.bytes_decompressed
        assert sequential <= unpacked * 1.1        # decode-once (tighter than the harness's ×2)
    with enable_measurement(), open_archive(archive) as reader:
        for name in reversed([m.name for m in reader.members() if m.is_file]):
            reader.read(name)
        assert reader.bytes_decompressed > sequential   # re-decode is visible
```

Gate the RAR params with the repo's existing `requires`/`importorskip` markers (as the `data` tests do) so core-only legs skip cleanly.

**2. Tighten the bound.** The harness uses `unpacked * 2` slack to tolerate arbitrary fixtures; for a controlled fixture you can assert `* 1.1` (or exact), which actually catches a *partial* re-decode that `×2` sails past. That's the difference between "documented" and "regression-proof."

Net: keep `test_benchmark_gate.py` + the harness for the perf/seek-baseline story, but promote the decode-once behavior into a small parametrized correctness test here (7z + RAR, tight bound). It'll fail loudly the moment any solid backend regresses, on every interpreter — independent of whether the benchmark job even runs.

---

davitf (commented):
Reviewed by reading the code, reproducing the gate at both scales, and running the tests. The measurement infrastructure and the structural gate are correct, deterministic, and reproduce exactly (I got identical byte counts at ci scale, and the realistic wall ratios line up directionally). Opt-in counters off the public ABC with zero hot-path overhead is a clean design.

The main theme of the comments: the PR **measures** the perf budget well but doesn't **gate** it — the wall-time / VISION ≤1.3× budget isn't run by any automation, so the "performance budget gate" is really a *structural-invariant* gate plus a manual wall-time tool. That's a reasonable response to how noisy the ratios are on shared runners (I reproduced zip drifting 1.18×→1.55× and the accelerator speedups halving), but the framing and the committed `wall_time.json` oversell it.

Inline items, roughly in priority order:
1. `wall_time.json` — enshrines the single-cold-pass method `RESULTS.md` debunks, and nothing reads it. Delete or regenerate + wire a job.
2. Benchmark job's `unrar` install is a no-op — no `rar` writer means no RAR fixture is built, so RAR is never gated in CI.
3. No scheduled full-mode run despite the comment/write-up referencing one — wall-time regressions aren't caught. Reword or add a nightly non-blocking job.
4. Accelerator headline numbers are core-count-specific and don't reproduce off the author's box — label with the machine.
5. Byte axis is tautological for common formats — the "no silent re-decode" claim there rests on the seek axis; soften the wording.
6. ISO instrumented but not benchmarked.
7. **Decode-once should be a first-class parametrized unit test** (it already is for 7z in `test_measurement.py`) — extend to RAR via the committed solid fixtures and tighten the bound; details inline.

Nothing here blocks the core instrumentation, which I'd happily merge. Items 1–3 are the ones I'd resolve before calling it a "gate."

---

PR #101, opened 2026-07-14
docs: investigate RAR unrar piping vs temp-file trade-offs

davitf:
## Summary

Explore-mode investigation prompted by: *"the RAR reader writes stream data to a temp file that it then passes to `unrar` — is that bad? Could we pipe a small synthetic header + compressed block instead, like `rarfile`'s hack but over a pipe?"*

Adds `docs/internal/rar-unrar-piping-investigation.md` (and links it from the internal index). **No code changes** — this captures measured findings, not an implementation.

## The load-bearing finding

**RARLAB `unrar` cannot read a non-seekable archive.** Measured against unrar 7.0.7 on Linux with the repo's fixtures:

| Invocation | Result |
| --- | --- |
| `cat a.rar \| unrar p -si` / `unrar p -` | rc 7 (cmdline error — `-si` is a *compressor* switch) |
| pipe / FIFO behind `/dev/stdin` | rc 2 (fatal, non-seekable) |
| seekable `/dev/stdin < a.rar`, or inherited **memfd** via `/proc/self/fd/N` | rc 0 ✅ |

This rules out the entire "pipe the header, then the compressed block" family of ideas and reframes the temp-file question as **disk vs RAM** and **whole-archive vs one-member** — never "stream it through."

## What else the doc establishes (all measured)

- `rarfile`'s "hack" is a **small temp file, not a pipe**; it's disabled for solid/split/encrypted/>20 MB members, and falls back to a whole-archive temp file otherwise — i.e. exactly what archivey does today. The small-member optimization is already spec-deferred + benchmark-gated.
- **memfd** (`memfd_create` + `pass_fds`) is a viable anonymous, seekable, in-RAM alternative to `mkstemp` for **single-file** archives — but trades disk for RAM and breaks multivolume (see below).
- **Solid**: a member can't be extracted in isolation (rc 3 — LZ window from prior members required), and members can't be merged into "one giant member" by byte-concatenation (rc 3 — needs real RAR5 header re-encoding).
- **Solid demux fragility (new finding):** the boundary tracks `unrar`'s *semantic* emission policy, which is not derivable from any single stored field — RAR5 symlinks are `csz=0/usz>0` (target in header) while RAR3 symlinks are `csz>0/usz>0` (target in LZ data), yet both emit 0 stdout bytes. Flagged as a candidate future hardening change (same class as the already-forbidden mixed-password ALL-pipe demux).
- **Multivolume**: whole-volume `cat` concatenation fails (rc 3); `unrar` discovers later volumes **by filename on disk**, so memfd can't serve them — validating the current materialize-named-`.partN.rar` approach.

## Recommendation (in the doc)

Drop the pipe idea permanently; the only clean win is the already-deferred small-member optimization (prefer memfd over mkstemp when it graduates), for seekable non-path sources only. The solid-demux emission-policy coupling is the one piece that could become its own OpenSpec change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GSfJHYQUP9XZ1XBhCmmUtv)_

### DIFF
--- a/docs/internal/index.md
+++ b/docs/internal/index.md
@@ -7,6 +7,7 @@ Maintainer / contributor material. Not the primary user guide.
 | [Threat model](threat-model.md) | Trust boundaries, enforced guarantees, open security/compat gaps |
 | [Compression-library analysis](library-analysis.md) | Per-codec backend choice and rationale |
 | [Known issues](known-issues.md) | Accelerator lifecycle / macOS coexistence notes |
+| [RAR ↔ unrar piping investigation](rar-unrar-piping-investigation.md) | Why `unrar` can't stream a piped archive; temp-file vs memfd vs synthetic-member trade-offs; solid-demux fragility |
 
 Normative behavior remains in `openspec/specs/`. Product framing: `VISION.md` at the
 repository root. Decision summaries: [decision log](../decisions/index.md).
--- a/docs/internal/rar-unrar-piping-investigation.md
+++ b/docs/internal/rar-unrar-piping-investigation.md
@@ -1,0 +1,259 @@
+# RAR ↔ unrar: can we replace the temp file with pipes?
+
+> Investigation note (2026-07). Origin: "the RAR reader writes stream data to a temp
+> file that it then passes to `unrar` — is that bad? `unrar` seems to support stdin;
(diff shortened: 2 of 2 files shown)


davitf:
Outdated

PR #115, opened 2026-07-15
review: crypto correctness deep dive (Brief 2)

davitf:
Deep review commissioned by `review/next/02-crypto.md` — every native decryption / key-derivation / integrity-verification path added since #73. Docs only, no source changes. Deliverables under `review/next/02-crypto/`.

## Baseline
`uv run pytest` → **1540 passed, 120 skipped**; `pyrefly` 0 errors, `ty` all-pass, `ruff` all-pass. Config `[all]` (cryptography 49, py7zr 1.1.3, rarfile 4.3); no `unrar`/`7z` binary in-container. Oracle diffs run against py7zr / rarfile / stdlib as libraries.

## Headline
The KDFs and ciphers are **correct** — bit-exact vs the project's own oracles (7z KDF vs py7zr, RAR3/RAR5 s2k vs rarfile, ZipCrypto vs stdlib `_ZipDecrypter`, BLAKE2sp vs official KATs, 0 mismatches). The bugs are in the **glue that decides whether an integrity check runs at all**. None is caught by the current suite (RAR data path needs `unrar`; the 7z/KDF triggers need format-legal-but-crafted archives py7zr never emits).

## Findings

| # | Sev | Where | One-liner |
|---|-----|-------|-----------|
| F1 | **High** | `rar_reader.py:119` | RAR5 tweaked-checksum members keep the key-tweaked BLAKE2sp in `member.hashes`, so a correctly-decrypted `-htb`-encrypted member fails `VerifyingStream` with a spurious `CorruptionError`. `_member_hashes` already drops crc32 when tweaked but not blake2sp; rarfile skips blake2sp when tweaked. |
| F2 | **Medium** | `sevenzip_reader.py:119` | An encrypted 7z folder with no folder digest and CRC-less members (both format-legal) confirms *any* password and returns garbage — no CRC gate, no `VerifyingStream`. Reachable for an AES+COPY (stored) folder where no codec rejects the garbage. |
| F3 | **Medium** | `crypto.py:176` | 7z `NumCyclesPower` is uncapped below the `0x3F` sentinel; a hostile `0x3E` forces ~2⁶² SHA-256 rounds (CPU DoS) during password confirmation / header decode. RAR5 caps its shift at 24; 7z has no equivalent. |
| F4 | Low | `rar_unrar.py:53` | Password passed to `unrar` as `-p<password>` in argv → visible via `ps`/`/proc` (matches rarfile; threat-model note). |
| F5 | Low | `rar_parser.py:1406` | RAR5 key-derived password-check compared with `!=` rather than `hmac.compare_digest` (WinZip AES correctly uses `compare_digest`). |

Each finding has a reproduction (unit-level + oracle diff) in the theme files. `QUESTIONS.md` collects three maintainer decisions (tweaked-blake2sp skip-vs-untweak; 7z fail-closed on checksum-less encrypted folders; 7z KDF cap + the `0x3F` sentinel). A "**what is actually fine**" section (verified-good: WinZip AES HMAC gating, non-seekable AES members, CTR boundary, `[crypto]`-absent contract, `cryptography>=45` floor) is in `verification.md` / `availability-and-contract.md`.

https://claude.ai/code/session_01VLyDWN8P1somMdMiq8pPi9

---
_Generated by [Claude Code](https://claude.ai/code/session_01VLyDWN8P1somMdMiq8pPi9)_

### DIFF
--- a/review/next/02-crypto/QUESTIONS.md
+++ b/review/next/02-crypto/QUESTIONS.md
@@ -1,0 +1,41 @@
+# Maintainer decisions — Brief 2 (crypto)
+
+These are the calls I did not want to make silently (CLAUDE.md: pause and ask on
+spec/behaviour discrepancies rather than resolving them).
+
+## Q1 — RAR5 tweaked-checksum BLAKE2sp (F1)
+
+When a RAR5 member is encrypted with tweaked checksums (`RAR5_XENC_TWEAKED`), the stored
+BLAKE2sp is key-tweaked and cannot be checked against the plaintext hash without reproducing
+the inverse tweak. Two options:
+
+- **(recommended) Skip it**, matching rarfile and matching what `_member_hashes` already does
+  for crc32 — drop blake2sp from `member.hashes` when tweaked, optionally emitting a
+  `DIGEST_UNVERIFIABLE` diagnostic. Minimal, restores correct reads immediately.
+- **Implement the inverse tweak** so archivey actually verifies tweaked BLAKE2sp (and could
+  re-enable the tweaked crc32 too). More faithful to WinRAR, but adds crypto surface and
+  needs a `-htb`-encrypted fixture to test.
+
+Either way the current behaviour (raise `CorruptionError` on good data) is wrong. Which
+direction?
+
+## Q2 — 7z folders with no integrity anchor (F2)
+
+An encrypted 7z folder with no folder digest and CRC-less members currently confirms *any*
+password and returns garbage. The safe fix is to **fail closed** — refuse to confirm a
+password when there is no checksum to validate against. Does that conflict with any planned
+"recover what you can from a damaged/odd archive" behaviour? My read of VISION #3 is that a
+*wrong-password-not-detected* is strictly worse than a refusal, so fail-closed is right — but
+confirm you want an encrypted-but-checksum-less member to hard-error rather than hand back
+unverified bytes.
+
+## Q3 — 7z KDF cap and the `0x3F` sentinel (F3)
+
+1. **Cap `NumCyclesPower`?** RAR5 caps its KDF shift at 24; 7z is uncapped below the `0x3F`
+   sentinel, so `0x3E` = ~2⁶² rounds. I recommend capping at 24 (raise
+   `UnsupportedFeatureError` above it). Agree on 24, or a different ceiling?
+2. **Is `0x3F` really a "no-hash" sentinel?** archivey follows py7zr (copy `salt+password`,
+   no hashing). If the 7-Zip C++ reference instead treats `0x3F` as `2^63` iterations, both
+   archivey and py7zr are wrong for that one value — but no creator emits `0x3F`, so it is a
+   crafted-only concern. Worth a one-line confirmation against the reference, or explicitly
+   accept "bit-exact with py7zr (our oracle)" as the contract and move on?
--- a/review/next/02-crypto/QUESTIONS.md
+++ b/review/next/02-crypto/QUESTIONS.md
@@ -1,90 +1,86 @@
-# Maintainer decisions — Brief 2 (crypto) — RESOLVED
+# Maintainer decisions — Brief 2 (crypto) — RESOLVED (source-confirmed)
 
-Answers below fold in davitf's PR #115 review (2026-07-15) plus a check of the `archivey-dev`
-reference at `730275b`. Open source-verification items are collected in
-`7z-source-questions.md`.
+Answers below fold in davitf's PR #115 review (2026-07-15), a check of the `archivey-dev`
+reference at `730275b`, and the **7-Zip + UnRAR source answers** (2026-07-16, see
+`7z-source-questions.md` for the raw citations). All questions are now closed; the "plan"
+lines are the agreed implementation direction.
 
-## Q1 — RAR5 tweaked-checksum BLAKE2sp (F1) — **decided: untweak-and-verify**
+## Q1 — RAR5 tweaked-checksum BLAKE2sp (F1) — **untweak-and-verify (both transforms confirmed)**
 
-davitf: *"DEV had special handling for tweaked rar checksums (maybe only crc32). If we have
-password(s) at open time, could we quickly check which one is correct using the check field,
-then derive the correct checksums from the tweaked ones … if we don't know the password,
-leave the hashes empty (maybe store the tweaked ones in extras)."*
+The tweak is RAR's `ConvertHashToMAC` — a **one-way forward transform**, so we compute the
+real hash from decrypted data, transform it, and compare to the stored value (we cannot
+recover the real checksum from the tweaked one). Both transforms are now confirmed against
+UnRAR `crypt5.cpp`:
 
-Findings from `archivey-dev/src/archivey/formats/rar_reader.py`:
+- **HashKey** = `PBKDF2-HMAC-SHA256(pw_utf8, salt, (1 << kdf_count) + 16)` — the AES key is at
+  `1 << kdf_count`, HashKey at `+16`, PswCheck at `+32` (single PBKDF2 chain, three cutpoints;
+  `crypt5.cpp:104,161`). Matches archivey-dev.
+- **CRC32:** `d = HMAC-SHA256(HashKey, crc.to_bytes(4,"little"))`; fold the 32-byte digest into
+  a 4-byte value by XOR of the eight little-endian uint32 words (UnRAR does it byte-wise with
+  `Digest[I] << ((I & 3) * 8)`, equivalent). Matches archivey-dev's `convert_crc_to_encrypted`.
+- **BLAKE2sp:** `tweaked32 = HMAC-SHA256(HashKey, blake2sp_digest[32])`, overwriting the 32-byte
+  digest (`crypt5.cpp:206`). This is the path archivey-dev never implemented.
+- **Gate:** the tweak is applied **iff** the file's `FHEXTRA_CRYPT_HASHMAC` (0x02) flag is set
+  (`arcread.cpp:1080`, `extract.cpp:934`) — which is exactly what v2's `_crc_is_tweaked` already
+  reads. **Correction to the earlier writeup:** encrypted-header (`-hp`) archives do **not**
+  auto-skip the tweak in the reader; unrar honours only the per-file 0x02 flag (untweaked
+  checksums under `-hp` are a *writer* choice to omit the flag). So no header-encryption
+  special-case is needed — key off 0x02 alone.
 
-- DEV implements `convert_crc_to_encrypted` = RAR's `ConvertHashToMAC`. The tweak is a
-  **one-way forward transform**, not reversible: `tweaked = f(hash_key, real_hash)`. So you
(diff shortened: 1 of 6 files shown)


review/next/02-crypto/QUESTIONS.md
```diff
@@ -0,0 +1,41 @@
...
+
+- **(recommended) Skip it**, matching rarfile and matching what `_member_hashes` already does
+  for crc32 — drop blake2sp from `member.hashes` when tweaked, optionally emitting a
+  `DIGEST_UNVERIFIABLE` diagnostic. Minimal, restores correct reads immediately.
+- **Implement the inverse tweak** so archivey actually verifies tweaked BLAKE2sp (and could
+  re-enable the tweaked crc32 too). More faithful to WinRAR, but adds crypto surface and
+  needs a `-htb`-encrypted fixture to test.
+
+Either way the current behaviour (raise `CorruptionError` on good data) is wrong. Which
+direction?
```
davitf: I remember that DEV had special handling for tweaked rar checksums (but maybe only crc32). if we have one or more passwords at open time, could we quickly check which one is correct using that check field, and then derive the correct checksums from the tweaked ones? we could then include the correct checksums in the members. if we don't know the password, then we should leave the corresponding hashes empty. (maybe just store the tweaked ones in extras)

review/next/02-crypto/QUESTIONS.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Q2 — 7z folders with no integrity anchor (F2)
+
+An encrypted 7z folder with no folder digest and CRC-less members currently confirms *any*
+password and returns garbage. The safe fix is to **fail closed** — refuse to confirm a
+password when there is no checksum to validate against. Does that conflict with any planned
+"recover what you can from a damaged/odd archive" behaviour? My read of VISION #3 is that a
+*wrong-password-not-detected* is strictly worse than a refusal, so fail-closed is right — but
+confirm you want an encrypted-but-checksum-less member to hard-error rather than hand back
+unverified bytes.
```
davitf: hard question. I'd rather not hard-error if the data might actually be correct. but that's likely to be an extremely rare case, right? how does 7z itself handle it? doesn't zip have this problem as well, isn't crc optional in them? they have a 1-byte check, but 1/256 cases will decrypt with an invalid password.

I didn't find the explicit part in the vision doc, but I think error detection is best-effort. what do you think?

review/next/02-crypto/QUESTIONS.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Q3 — 7z KDF cap and the `0x3F` sentinel (F3)
+
+1. **Cap `NumCyclesPower`?** RAR5 caps its KDF shift at 24; 7z is uncapped below the `0x3F`
+   sentinel, so `0x3E` = ~2⁶² rounds. I recommend capping at 24 (raise
+   `UnsupportedFeatureError` above it). Agree on 24, or a different ceiling?
```
davitf: do you know the actual cap the 7z code uses? is it user-selectable? what's the max?

review/next/02-crypto/QUESTIONS.md
```diff
@@ -0,0 +1,41 @@
...
+2. **Is `0x3F` really a "no-hash" sentinel?** archivey follows py7zr (copy `salt+password`,
+   no hashing). If the 7-Zip C++ reference instead treats `0x3F` as `2^63` iterations, both
+   archivey and py7zr are wrong for that one value — but no creator emits `0x3F`, so it is a
+   crafted-only concern. Worth a one-line confirmation against the reference, or explicitly
+   accept "bit-exact with py7zr (our oracle)" as the contract and move on?
```
davitf: I can get the 7z source code and ask an agent to check both this and the question above. can you write everything we'd like to know from it?

review/next/02-crypto/SUMMARY.md
```diff
@@ -0,0 +1,50 @@
...
+| F2 | **Medium** | `sevenzip_reader.py:119` `_verify_decoded_folder` | Encrypted 7z folder with **no folder digest and CRC-less members** (both format-legal) confirms *any* password and returns garbage with no error (no CRC gate, no `VerifyingStream`) | unit (`_verify_decoded_folder` accepts wrong-key garbage) | Confirmed (logic); full e2e needs crafted AES+COPY CRC-less 7z |
+| F3 | **Medium** | `crypto.py:176` `derive_sevenzip_aes_key` | 7z `NumCyclesPower` is uncapped below the `0x3F` sentinel; a hostile value `0x3E` forces ~2⁶² SHA-256 rounds (CPU DoS) during password confirmation / header decode. RAR5 caps its shift at 24; 7z has no equivalent | `parse_sevenzip_aes_properties` + derive | Confirmed |
+| F4 | Low (hardening) | `rar_unrar.py:53` `_password_arg` | Password passed to `unrar` as `-p<password>` in argv → visible to other local users via `ps`/`/proc` | code read | Confirmed; matches `rarfile` behaviour |
```
davitf: is there any way of avoiding it? passing via stdin?

review/next/02-crypto/SUMMARY.md
```diff
@@ -0,0 +1,50 @@
...
+| F5 | Low (hardening) | `rar_parser.py:1406` `_check_rar5_password` | RAR5 password-check value (key-derived) compared with `!=` rather than `hmac.compare_digest` | code read | Confirmed; WinZip AES correctly uses `compare_digest` |
```
davitf, line 38: is the difference just resilience to timing attacks? we're not dealing with secret server side data, so there's no real attack surrface I think, but it may be worth fixing for full correctness

review/next/02-crypto/QUESTIONS.md
davitf: Checked DEV (`archivey-dev@730275b`, `formats/rar_reader.py`). It has `convert_crc_to_encrypted` = RAR's `ConvertHashToMAC`, and the key detail is that the tweak is a **one-way forward transform**, not reversible: `tweaked = HMAC-based(HashKey, real_hash)`. So we can't "derive the correct checksums from the tweaked ones" — the real CRC/BLAKE2sp isn't recoverable. What DEV does (and what we should do) is compute the real hash from decrypted data, apply the same forward transform, and compare to the stored tweaked value.

Keys (PBKDF2-HMAC-SHA256 over UTF-8 pw + salt): AES key at `1<<kdf_count`, **HashKey at `(1<<kdf_count)+16`**, PswCheck at `+32`. CRC32 tweak = XOR of the eight uint32 words of `HMAC-SHA256(HashKey, crc_le32)`. BLAKE2sp tweak = `HMAC-SHA256(HashKey, blake2sp32)` (per UnRAR — flagged for source confirmation). **DEV implemented crc32 only and always dropped the tweaked crc32** (`crc32=None`), so it never false-positived; v2 regressed by dropping crc32 but keeping the tweaked blake2sp.

Plan matches yours: (1) interim — make `_member_hashes` symmetric (drop tweaked blake2sp too, stash both tweaked values in `extra`) to kill the false `CorruptionError`; (2) full — when a password is confirmed at open time, verify by forward-transform for both crc32 and blake2sp; leave `hashes` empty when the password is unknown. Written up in `QUESTIONS.md` Q1.

---

review/next/02-crypto/QUESTIONS.md
davitf: Agree — best-effort, don't hard-error. Retracting the "fail closed" recommendation.

- **7-Zip itself** relies on the folder/stream CRC to reject a wrong password; with no CRC it can't detect one either and just returns what the cipher produced. So accept-and-return matches 7-Zip.
- **ZIP is not actually exposed** the same way: WinZip AE-2 sets CRC=0 but authenticates with a **mandatory** HMAC (checked on close — verified good in this review), and ZipCrypto STORED is disambiguated by the central-directory CRC. So every ZIP encrypted member still has an anchor (HMAC or CRC). 7z is the only format here where an encrypted member can legitimately carry **zero** anchor (AES+COPY, no digest, CRC-less). And the 1/256 ZipCrypto one-byte-check case is already handled by the multi-candidate CRC pass — different problem.

So: keep current behaviour, but emit a diagnostic (`DIGEST_UNVERIFIABLE`/new `DECRYPTION_UNVERIFIED`) when an encrypted folder has no anchor, so the caller can tell "decrypted but unverifiable" from "verified". **F2 downgraded to Low.** Added a source item to confirm 7-Zip's no-CRC behaviour (`7z-source-questions.md` §B).

---

review/next/02-crypto/QUESTIONS.md
davitf: What I can state now: 7-Zip **encodes** with `NumCyclesPower = 19` (2¹⁹ ≈ 524k SHA-256 rounds, sub-ms) and doesn't expose the value in its GUI/CLI, so real archives are effectively fixed at 19. The **format** stores it in the low 6 bits of the AES property byte, so a hostile file can carry 0–0x3F regardless of what the official encoder emits.

The bit I can't confirm from Python: whether the 7-Zip **decoder** caps it or loops `1 << value` unbounded (i.e. whether 7-Zip is itself DoS-able) — that's in the source checklist (`7z-source-questions.md` §A). Independent of the answer, I'd recommend v2 apply its own cap: reject `NumCyclesPower > 24` (2²⁴ ≈ 16.7M, ~100ms), which mirrors the RAR5 KDF cap of 24 that v2 already enforces. Open to a different ceiling.

---

review/next/02-crypto/QUESTIONS.md
davitf: Done — wrote it up as `review/next/02-crypto/7z-source-questions.md`, phrased for a source-reading agent with file/line targets. It covers two codebases (7-Zip and UnRAR) and folds in both of the above questions:

- **§A (7z KDF):** does the decoder bound `NumCyclesPower` or loop `1<<value` unbounded? Is there a `0x3F` "no-hash" special case in official 7-Zip (py7zr/archivey assume one)? Encoder default + max?
- **§B (7z no-CRC):** how does 7-Zip detect a wrong password when no CRC is present — CRC-only (silent garbage) or another gate?
- **§C (RAR tweak):** confirm `ConvertHashToMAC` for **both** crc32 and blake2sp, the HashKey derivation (`+16`), and that the tweak applies only when `RAR5_XENC_TWEAKED` is set.
- **§D (unrar):** confirm there's genuinely no non-interactive stdin/env/file password channel (see my reply on the SUMMARY thread).

---

review/next/02-crypto/SUMMARY.md
davitf: Not really, unfortunately. To my knowledge RARLAB `unrar` has no scriptable non-interactive password channel other than `-p<password>` — no env var, no `--password-file`, and its interactive prompt reads from the console/tty rather than a plain piped stdin, so it can't be fed reliably from a subprocess pipe. That's why rarfile, patool, etc. all use `-p<pwd>`. So it's largely unavoidable without switching the decoder — best treated as a documented threat-model limitation (local-user confidentiality is outside the current trust boundary). I added a source item (`7z-source-questions.md` §D9) to have the UnRAR-source agent confirm there's genuinely no stdin/env/file option before we close it out as "unavoidable".

---

PR #224, opened 2026-08-04
Docs: archive #223, write opening-and-listing and reading-members

davitf:
Documentation only. The library fixes this review turned up are split out into **#225**, which is based on this branch — merge this first and GitHub retargets it to `main`.

## `83dc4c0` — archive `docs-ia-split-user-guide` (#223)

Mechanical, same shape as #222 was to #221, and **time-sensitive**: `scripts/check_openspec_archived.py` fails on `main` until it lands. `+1 added, ~2 modified, -1 removed`, matching the dry run. The published-tree requirement now enumerates the 15 real pages; the dual listing contract is owned by `errors-and-diagnostics.md`; the Gotchas digest rule replaces the old post-v1-limitations requirement. `openspec validate --specs --strict` 24/24.

## `3a43c31` + `3da0598` — `opening-and-listing.md` (76 → 176)

New: a "What you can open" table; multi-volume as its own section, led by the case nearly every caller hits (pass **any one volume** and Archivey finds the rest, across `.7z.NNN`, `.partN.rar` and `.rar`+`.rNN`, with 7z sets checked for completeness); detection naming `FORMAT_EXTENSION_CONFLICT`, and the one mismatch that is deliberately silent; `open_archive` vs `open_stream`; passwords against an unencrypted format.

Your review caught four factual errors, all mine, and all in the branch where the claim does not hold — the codec that is absent, the stream with data after it, the backend whose flag is `False`. That is now an explicit step for every remaining page, and it caught an error on the very next one.

**Three broken anchors fixed**, all created by #223 and all shipped green — two `formats.md` links into sections the Gotchas shrink deleted, one pointing at `#accelerators-and-process-aborts` where the heading says "source lifetime". The `formats.md` pair pointed *backwards*: the digest rule says Gotchas links to the owning page, and `formats.md` already stated both facts in full.

**`check_docs_nav.py` now resolves every intra-`docs/` anchor**, cross-page and same-page. `mkdocs build --strict` reports a missing anchor at INFO and exits 0, exactly as it does for a page missing from the nav. Ids come from running each file through Python-Markdown's own `toc` extension, so the check cannot drift from the slug rule the site uses. Verified against planted failures of both kinds — worth having before a rewrite, since moving a heading is the most common thing the remaining prose will do.

## `7128c33` — `reading-members.md` (84 → 159)

The opening/reading boundary, settled with the rule the outline already used for the split: *a page owns a decision if the reader makes it while doing that page's job*. `streaming=` is chosen with your hand on `open_archive`; `open()` vs `stream_members()` is chosen when you want bytes. So the solid-archive cost argument moved off the opening page, which had grown a strategy section that was never its job.

New, none of it previously documented: the `stream_members` contract — the yielded stream is valid only until the iterator advances, non-file members yield `None`, and nothing is decompressed until you read, so **iterating the whole archive without reading does not check passwords**. Also which members can be opened (links followed, non-file entries raise, a member belongs to its own reader, hence `member in reader` being identity), `read()` having no size limit, and why `extract()` has no `members=` but does accept a non-seekable source.

Table cells stay short from here on, generalised from your comment on the five-line non-seekable row: the cell states the answer, the prose below carries the condition and the reason.

---

**These pages describe `main`'s behaviour**, deliberately — including the four points #225 changes (directory `format=` ignored, streams surviving reader close, `.lz` size needing a path, deferred teardown). #225 carries the sentence updates alongside the code, so each PR is internally consistent.

Verification: `check_docs_nav.py` clean, `mkdocs build --strict` green, all `docs/` Python blocks parse, suite 2116 passed / 65 skipped.


docs/opening-and-listing.md
davitf: it's unclear to be what "get past whatever guards it" mean

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
     print(reader.format, reader.cost)
 ```
 
-Default open is **random access** (`streaming=False`). On a pipe or other non-seekable
-source, either pass a seekable file or use `streaming=True` (forward-only, one pass).
+You get random access by default. If your source is a pipe or anything else you
```
davitf: I understand it, but I think a regular user might not understand what "you get random access by default means". maybe something like "By default, you can open any member independently" (and perhaps here briefly mention the gotcha about how doing it may be inefficient in some formats, and iterating as a better option depending on use case)

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+cannot seek in, pass `streaming=True` for a forward-only single pass — otherwise
```
davitf: "anything else you cannot seek in" -> other/another non-seekable stream 

Clearer I think. the current way seems unnecessarily informal, it's fun but actually harder to understand

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
 
-`members()` / `scan_members()` assert a **complete** listing and raise on terminal
-damage; `members_report()` gives you the recoverable prefix *and* the error together.
-Iteration yields the prefix, then raises. See
-[Errors and diagnostics](errors-and-diagnostics.md#listing-a-damaged-archive) for the
-recipe and what each failure means.
+| Source | What happens |
+|---|---|
+| A path to a file | Opened and detected as usual |
+| A path to a directory | Opens as a pseudo-archive, one member per file. A directory always opens as a directory, even if you pass `format=` |
```
davitf: opening a dir with an explicit non folder format should fail, don't you think? worth registering for a future fix?

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+| An open binary stream | See below |
+| A sequence of paths or streams | The volumes of one multi-volume archive, in order. **7z and RAR only** — anything else raises. A one-item sequence is just that item |
```
davitf: maybe multi volume is worth a separate mini section. we should state that, if given a filename for the first volume, the others are found automatically

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+
+Streams have one rule worth knowing: **a seekable stream is read from wherever it
+currently is.** Archivey treats that position as byte 0 of the archive, so an
+archive embedded in a larger file needs no slicing — seek to where it starts and
```
davitf: I think this is too strong: maybe it does need slicing if there's stuff after the archive

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+hand the stream over. A non-seekable stream is fine too, with `streaming=True`;
```
davitf: non seekable is fine for some archive types, not all. worth being clear

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+Archivey buffers what it peeks at during detection and replays it, so detection
```
davitf: the detection peeking is internal magic, not worth stating here; the user-facing behavior is just that you can open any stream, even non seekable

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+never eats bytes the archive needs.
 
-## Detect without opening
+## Detection
```
davitf, line 92: I think it's worth one line stating that most users won't need this, as opening auto detects

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
 
+**Content wins over filename.** Archivey looks at the bytes first and falls back to
+the extension only when they are inconclusive. When the two disagree it uses the
+bytes and tells you, via a `FORMAT_EXTENSION_CONFLICT`
+[diagnostic](errors-and-diagnostics.md) naming both candidates — a `.jpg` that is
+really a ZIP opens fine, and you can still find out that the name lied.
+
+One disagreement is not reported, because it is not one: a `foo.tar.gz` may come
+back as plain `GZ`. Deciding whether a compressed stream contains a tar means
+looking inside it, and detection skips that when the decompressor isn't installed.
```
davitf: gz is not a good example here then, as it's in stdlib. we should use another example (xz or zstd < 3.14)?

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+The question is settled when you open the file.
```
davitf: this seems wrong. I think detection will return the same format as opening . it detects tar internally if the codec is installed; if not, opening will fall anyway, so the question is not settled

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+
+That is also the difference between the two entry points on the same `.gz`:
+
+```python
+archivey.open_archive("logs.tar.gz")   # the files inside the tar
+archivey.open_stream("access.log.gz")  # the decompressed bytes, as a stream
+```
+
+`open_archive` works on a plain `.gz` as well — you get an archive with exactly one
```
davitf: this seems like a topic for the open_atchive section, not the detection one

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
 
 ```python
 archivey.open_archive("secret.7z", password="hunter2")
 archivey.open_archive("secret.zip", password=["likely", "fallback"])
 ```
 
-List the most likely password first — especially for 7z, where each wrong candidate pays
-key derivation.
+Put the most likely password first. Every wrong candidate costs work before it is
+rejected, and on 7z that work is expensive key derivation.
```
davitf: how about: 

... rejected, which can be expensive (especially on 7z archive).

docs/opening-and-listing.md
```diff
@@ -17,56 +17,98 @@ with archivey.open_archive("photos.zip") as reader:
...
+
+Passing a password to a format that has no encryption at all — a tar, say — raises
+`UnsupportedOperationError` rather than ignoring it, since it usually means the call
+is not doing what you think. A `PasswordProvider` callable is exempt: it is only
+called if something actually asks for a password.
+
+## Damaged archives
+
+`members()` / `scan_members()` assert a **complete** listing and raise on terminal
```
davitf: "assert" sounds weird. maybe something with verify?

aso "raise on terminal damage", what's terminal? maybe "raise if the archive can't be fully read" or " if an error is detected while reading the members"

docs/opening-and-listing.md
```diff
@@ -12,61 +12,159 @@ with archivey.open_archive("photos.zip") as reader:
...
+member, named after the file. Use `open_stream` when you just want the bytes and
+know there is no tar inside.
+
+## What you can open
+
+| Source | What happens |
+|---|---|
+| A path to a file | Opened and detected as usual |
+| A path to a directory | Opens as a pseudo-archive, one member per file. This holds even if you pass `format=` |
+| An open binary stream | Any format can be read from a *seekable* stream. From a non-seekable one — a pipe, a socket, an HTTP response — only TAR (including compressed tar) and the single-file compressors work, with `streaming=True`. ZIP, 7z, RAR and ISO keep their index at the end of the file or address it by offset, so they need to seek; asking anyway raises `StreamNotSeekableError` at open, suggesting you buffer to disk or a `BytesIO` |
```
davitf: this seems too long for the table. something like Only some formats can be read from non-seekable streams such as ... . See below . we add the remaining info to the paragraph below, and maybe move the per format justification to the formats page. I  wonder if the non seekable examples should be moved to the paragraph as well

docs/opening-and-listing.md
```diff
@@ -12,61 +12,159 @@ with archivey.open_archive("photos.zip") as reader:
...
     print(reader.format, reader.cost)
 ```
 
-Default open is **random access** (`streaming=False`). On a pipe or other non-seekable
-source, either pass a seekable file or use `streaming=True` (forward-only, one pass).
+**By default you can open any member you like, in any order.** That is what most
+callers want, and it is what the example above relies on.
 
-### Damaged archives
+It is not always the cheapest thing to do, though. In a solid 7z or RAR, or any
```
davitf: should this reading cost info move to the reading page?

PR #236, opened 2026-08-13
Escape archive-derived text where it becomes a message (threat-model O9)

davitf:
Closes threat-model **O9**, registered in #235. Per `AGENTS.md`, each open item in that register becomes an OpenSpec change when tackled — this is `escape-cli-log-records`.

> **Rebased onto `main`** (was stacked on #235's branch). #235 was squash-merged, so its branch commits were never ancestors of `main` — a plain retarget would have replayed them as if new. Rebased with `--onto main`, dropping the two #235 commits whose content the squash already carries; the resulting tree was byte-identical to the pre-rebase branch, so the diff below is purely this change.

> **Reworked in place after review.** This PR originally escaped at the CLI's log handler. That was the wrong layer (below), and the fix replaces most of it. Reworked here rather than landed and reverted in a follow-up, so the spec history never claims a guarantee we no longer hold.

## The gap

`escape_member_name` exists because member names are attacker-controlled: a name carrying `\x1b[2K\r` lets an archive erase the line reporting it and author what the operator reads instead. #235 closed every CLI **print site**. It did not close the library's `logging` records, which reach the same stderr:

```
WARNING: Skipping file 'EV\x1b[2KIL\rHARMLESS.TXT': Destination already exists: /…/out/ev<ESC>[2Kil<CR>HARMLESS.txt
failed:  EV\x1b[2KIL\rHARMLESS.TXT: Destination already exists: /…/out/ev\x1b[2Kil\rHARMLESS.txt
```

Same fact one line apart — the second is the fixed print site, the first is the log record. In `logger.warning("Skipping %s %r: %s", type, name, error)` the *name* is safe by accident (`%r` makes `repr` escape it); the *error* is not, because its message embeds a destination path built from that same name.

## Why the handler was the wrong layer

A `logging.Formatter` only guards records that reach a handler someone configured. It does nothing for the route an archivey exception is *most likely* to take to a terminal: propagating uncaught, with the interpreter printing the traceback — whose final line is `str(exc)`, and which passes through no handler at all. `print(exc)` in embedding code and third-party error reporters are the same shape.

The text is dangerous because of where it **comes from**, not where it is displayed.

## The fix

Escape where archive-derived text *becomes a message*:

- **`ArchiveyError` / `ArchiveyUsageError` escape `message` at construction.** One place covers all 26 subclasses. `message`, `args[0]`, `str()` and `repr()` all render the escaped form. `archive_name`, `member_name`, `link_target` and `source_format` stay **raw** — they are data, not display — and `__str__` renders the names through `!r`.
- **`Diagnostic` escapes `message` too.** `diagnostics_collector` logs it as `"%s"`, making it the one library log site interpolating neither an exception nor `%r`. `context` stays raw as the structured channel behind `to_dict()`.
- **The primitive moves to `archivey/escaping.py`**, below both `exceptions` and `cli`.

**The two layers cannot coexist:** escaping an already-escaped message doubles every backslash in it. `_EscapingFormatter` is removed, and `format_error_detail` decides by type in one place. `CliError` is outside the hierarchy and keeps its own print-site escape.

This also **closes the `exc_info` residual** the handler-side design had to accept: the traceback's final line is the exception's message, now escaped at the source.

## Escape exactly once (review round 2)

An audit found **52 message sites** interpolating an archive-derived name with `{name!r}` — `repr` escapes, then the message escape escapes its backslashes — essentially every safety error in `filters.py`, `extraction.py`, `base_reader.py` and `reader_state.py`. The same name appeared twice in one line with two different spellings, since `member=` renders from the raw attribute. Fixed rather than accepted:

- **`escaping.quoted()`** supplies delimiting quotes *without* escaping, choosing `"` when the text holds `'`. All 52 converted.
- **`raw_message` + `raw_message_of()`** do the same for a caught exception embedded in a new message — exactly 2 sites, both broad `except Exception` handlers in `rar_parser.py`.
- **The inverse rule** now matters: the CLI handler no longer escapes, so the 9 `logger.*` sites interpolating a name with `%r` are load-bearing.

## Escaping correctness (review round 2)

Rendering delegates to `repr`, whose escape set is exactly `not str.isprintable()` (verified across the whole code space — 0 mismatches), after three confirmed bugs:

- **Astral code points emitted a five-hex-digit `\uXXXXX`**, reading back as a four-digit escape plus a stray digit. 955,086 code points affected.
- **The surrogateescape range was `U+DC00`–`U+DFFF`** rather than the `U+DC80`–`U+DCFF` that `surrogateescape` produces, reversing 768 code points into bytes they never came from.
- **The losslessness claim was false** — `U+009B` and a surrogateescaped byte `0x9B` both render `\x9b`. Narrowed to inertness.

## Names live in structured fields (review round 2)

An inventory of all 418 exception sites found 36 putting a name in the message, **20 printing it twice**. `link_target` joins `member_name` as a structured attribute rendered as `target=`; 31 of the 36 messages are now prose plus attributes. Four are irreducible and listed in `tasks.md`.

## Review round 3

- **`Destination already exists` still interpolated a native path** while its siblings used `display_path()` — the message this change was reproduced on. The cause: the audit behind that fix matched `\bdest\b`, which does not fire inside `dest_path`. Same word-boundary trap as the call-site guard, in a second script.
- **Both one-off audits are now static tests**, so *"no native paths in messages"* and *"no `!r` on archive-derived text"* are checked every run rather than resting on a sweep nobody repeats. Module exemptions must name a real file, since a stale exemption reads as coverage while providing none.
- **The residual story is reconciled** — it had disagreed across proposal, threat model and spec.

## Specification

- `error-handling` — *Exception messages are inert for terminal display*, *Archive-derived text is escaped exactly once*, `raw_message` / `link_target` in the attributes table
- `diagnostics` — *Diagnostic messages are inert for terminal display*
- `cli` — rewritten to "escaped where it becomes a message"

## Accepted residual

None outstanding. The Windows separator-doubling residual is closed: paths in messages are rendered `/`-separated by `display_path()` before escaping, guarded by a static test since the failure is invisible on Linux.

## Archived

Both `escape-cli-log-records` and `extraction-results-authoritative` are archived in the final commit — pure relocation, 11 files, all renames, 0 insertions and 0 deletions, since the deltas were already applied to `openspec/specs/`. `extraction-results-authoritative` belongs to #235, which merged unarchived; the gate names both and neither passes alone, so it is cleaned up here rather than in a further follow-up.

Held until last deliberately: two review rounds moved this change's design, and archiving early is what forced the unarchive-and-rework this branch carries.

## Verification

- **The live spoof no longer reproduces.** Both the `WARNING` and the `failed:` line render inert, and the collision line's renderings of the name now agree. Uncaught traceback checked out-of-process under `cat -v`.
- `tests/test_escaping.py` (45 tests): adversarial escaping cases, the escape-exactly-once rules, and three static call-site guards.
- Full suite in all three dependency configurations: `[all]` 2453, `[all-lowest]` 2453, `[core-only]` 2005 + 414 skipped.
- `ruff`, `pyrefly`, `ty` clean. `openspec validate --all --strict`: 25 passed. `check_openspec_archived.py` passes.


davitf:
Thanks — reviewed against HEAD `00f0686`. Findings 1–3 were the CI failures, fixed in `96a48a8`/`3eeb554`; finding 4 was a real miss and is now fixed.

**#1 pyrefly `display_path(path: object)`** — fixed, annotated `str | os.PathLike[str]`. You're right that I should have caught this: I'd been running `ruff` locally but not `pyrefly`/`ty`, which `CONTRIBUTING.md` lists.

**#2 `read_text()` without encoding** — fixed at all three sites. Your diagnosis of the cause was exact (U+201D → `0x9d` under cp1252).

**#3 abort-notice test on Windows** — fixed, taking your second option. Split in two, matching how the overwrite-report tests are already paired: the Windows-legal U+2028 spoof covers the abort print site on every platform, and the ANSI erase-and-rewrite variant carries the repo's existing `_ANSI_ONLY` marker. Your note that the WARNING line's inertness is "mostly accidental" is fair — it comes from `%r` on the name plus `OSError.__str__` using `repr` for `filename`, which is stdlib behaviour we don't own. The CLI's own print path for `OSError` is explicit (`format_error_detail` escapes anything that isn't ours).

**#4 `Destination already exists` still interpolating a native path** — confirmed and fixed. This is the one worth dwelling on, because the *cause* was worse than the symptom: the audit that found the other path sites matched `\bdest\b`, which doesn't fire inside `dest_path` — an underscore is a word character. That is the **same word-boundary trap** I'd already found and fixed in the call-site guard, still present in a second, separate script I never re-ran. Two ad-hoc sweeps, same bug, one silently under-reporting.

So both sweeps are now static tests instead of scripts: *"no message interpolates a native path"* joins *"no `!r` on archive-derived text"* and *"no `%s` for a name in a `logger.*` call"*. The claim is checked every run rather than resting on an audit nobody repeats. I also added an assertion that each module exemption names a real file — a stale exemption reads as coverage while providing none, which is how the Windows separator mismatch hid.

**Residual story** — you were right that it disagreed across four documents. Taking your option **A**: the site is fixed, the residual is deleted from `dev-docs/threat-model.md` and the `error-handling` scenario table, and the positive rule (paths rendered `/`-separated before escaping) is now a SHALL rather than docstring-only prose.

**#6 nit** — added a banner over tasks §§1–6 marking them superseded by §7.

Two decisions I've deliberately left to the maintainer rather than acting on:

- **Archiving** (your #5 / decision 2). Still unarchived, and I agree it will turn `main`'s OpenSpec gate red post-merge. I've held off because two review rounds have moved the design, and archiving early is what created the revert-vs-rework problem this PR already had to undo. Happy to archive on request — and to coordinate the leftover `extraction-results-authoritative` from #235 in the same landing.
- **Retargeting onto `main`** (your decision 3). Agreed the current diff is awkward, but retargeting alone would make it worse, not better: #235 was squash-merged, so its branch commits aren't ancestors of `main`, and a diff against `main` would replay #235's changes as if new. It needs a rebase, not a base change — a force-push to a branch under active review, which I'd rather the maintainer time.

---

PR #239, opened 2026-08-15
Workflow friction: close the review loop, archive in the implementing PR, map the code

davitf:
Findings from reading the last three weeks of PR history (#199–#238), plus fixes for the four causes that account for most of the repeating friction. No library code changes — process, docs, one CI gate.

## 1. Archiving moves into the implementing PR

The OpenSpec gate ran on `main` only, on the stated reasoning that an implementing PR is *meant* to tick its last task while still unarchived. That reasoning assumed archiving is a separate step by nature. It is not — changes here are usually proposed, implemented, and finished in one PR, and that PR is where the archive belongs.

Deferring produced both halves of the problem the gate exists to prevent:

- a window on `main` where the authority was wrong (three times running — #212, #213, #214, which is why the gate was written), and
- **five of the thirty-seven merges in three weeks** whose entire content was `openspec archive`: #214, #215, #222, #227, #238.

It also surfaced as a finding in three consecutive review rounds on #236 without ever being resolved there.

**Change:** the `openspec` job drops its `if: github.event_name != 'pull_request'`. The escape hatch is unchanged and costs one character — a change still in motion leaves its trailing task unchecked — and is now stated in the failure message, `CONTRIBUTING.md`, and `AGENTS.md` rather than only in a script docstring. #236's reason for deferring (archiving early is what turns a review round into a revert-and-rework) is exactly what the unchecked box is for.

Recommended convention, already followed by most of the archived corpus: **make the archive the change's last task**, so checking the final box and applying the deltas are one act.

Both branches verified against a scratch tree — all-checked fails, one-unchecked passes.

## 2. The review loop's second half now exists

The workflow is: one agent reviews and posts findings, a second agent works through them. Only the reviewing half was written down.

**New skill `address-review-findings`** (+ `.cursor/commands/address-review.md`):

- **Read all four channels** — top-level comments, inline review threads, formal reviews, CI check runs. Reading only the top-level comment is how half a review gets missed.
- **Who actually said this.** Agents post through the maintainer's GitHub account, so a `davitf` comment *with* a `_Generated by [Claude Code]_` footer is an agent and the same login *without* one is the human. A maintainer's inline question is a question to answer, not a finding to close.
- **Verify before acting.** On #232 one of two automated findings was a false positive checked against the wrong file, and the other was real but narrower than described. On #236 a fix landed for the symptom while the cause — an audit regex where `\bdest\b` silently failed to match `dest_path` — survived in a second script.
- **Ledger first:** every finding gets exactly one disposition — fixed / disproven / escalated / deferred-with-a-written-home. A disproven finding is usually doc-debt, not nothing.
- **Full gate before pushing**, since `ruff` alone is the most common self-inflicted CI failure here.
- **Escalate one question at a time**, written to be decidable cold — identifiers expanded, actual code quoted, measurements shown, options with consequences, a labelled recommendation.
- **Reply in the thread the finding came from, and resolve what you resolved.** #236 finished with four threads answered and fixed but never resolved.

**Addendum §10** covers the posting side so the handoff works: stable finding IDs (`F1`, `F2`, …) kept across re-reviews, located findings anchored inline so threads can be resolved individually, blocks 1 and 3 in the review body, a status table over previous IDs when re-reviewing, gates re-run rather than copied from the PR body, and the attribution footer.

## 3. `dev-docs/code-map.md`

Every module already has a good docstring, so the gap is not "what is this file" — it is *where do I start*. The map carries the tree shape, the two boundaries that hold the design (`archivey/` vs `internal/`, `backends/` vs `streams/`), the call path through a read, a task→files index, and a "which doc answers which kind of question" table.

Deliberately coarse so it does not rot; it says so, and says to deepen a module docstring rather than add a row.

It also records three things that have cost real time: listing can run twice with different member objects (dedupe on member id, not identity — and a TAR-based guardrail will not exercise it), translation returns `None` for unknowns rather than catching all, and `ArchiveyUsageError` sits outside the error tree on purpose. The closing note points at the ADR log, `review/STATUS.md`, `dev-docs/discussions/` and the threat-model register, because questions have been re-derived from scratch more than once — ADR 0015 was written only after the empty-TAR question came up three times across two rounds.

## 4. AGENTS.md corrections

Incorrect or outdated:

- A bullet in §Conventions had **lost its head** mid-list — it began "git hook with…" as a dangling fragment, and pointed at `AGENTS.md` from inside `AGENTS.md`.
- The `docs/` page list named **`safe-extraction`**, renamed to `extracting.md` in #223, and omitted `index`.
- **`review/` was absent entirely** — a whole top-level directory with the review program, live `STATUS.md`, `backlog.md`, and twelve archived rounds. An agent commissioned for a review had no pointer to it.
- `dev-docs/IDEAS.md` appeared **twice**; `dev-docs/discussions/` did not appear at all.
- The addendum's review table claimed **"all archived as of 2026-07-28"** while the docs review is in flight and simplicity-consistency archived on 2026-08-15. It now points at `STATUS.md` as the live index.
- A **broken relative link** in `code-review-skill/SKILL.md` (`../../` where `../../../` was needed).

Added:

- **Cross-platform traps** — the container is Linux, CI is Windows and macOS. `read_text()` without `encoding="utf-8"` (cp1252 vs a curly quote in a source file), control characters and `:*?"<>|` in on-disk names (`WinError 123` at creation, before the assertion is reached), path separators in compared strings, and case-insensitive filesystems. All four hit #236 after it was pushed.
- **Both type checkers, every time** — `ruff` passing is not the gate.
- A **§Review workflow** section describing the two-agent handoff and the shared-account attribution problem.

`CLAUDE.md` gains three lines: a fresh container has no memory of earlier sessions, so check the written record before re-deriving.

## Verification

- `openspec` CI job parses; `if` key confirmed removed, job still present.
- `check_openspec_archived.py` verified on both branches against a scratch tree; clean on this one.
- `ruff check` + `ruff format` clean.
- All relative markdown links in changed and added files resolve (that check is what found the pre-existing broken one).
- `check_docs_nav.py` unaffected — `dev-docs/` is unpublished, and this adds nothing under `docs/`.

## Notes for the maintainer

Two judgement calls worth a look, both easy to reverse:

1. **The PR-scoped gate can fail a PR through no fault of its own** if `main` already carries a finished-unarchived change. It cannot happen while `main` is kept clean, which the existing main-branch gate ensures — but it is the one new way this can be annoying.
2. **`code-map.md` is new surface that can drift.** Kept coarse for that reason. If you would rather not carry it, the task→files index is the part worth keeping and the tree diagram is the part most likely to rot.

---
_Generated by [Claude Code](https://claude.ai/code/session_019ERETT4fvyzD9HEgxZYzbd)_


.claude/skills/address-review-findings/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+Run the real gate, not a subset. The most common self-inflicted CI failure in this repo is
+pushing after `ruff` alone:
+
+```bash
+uv run --no-sync ruff format src/ tests/ scripts/ benchmarks/
+uv run --no-sync ruff check --fix src/ tests/ scripts/ benchmarks/
+uv run --no-sync pyrefly check      # ← both type checkers, not just one
+uv run --no-sync ty check
+uv run --no-sync pytest
+```
```
davitf: Should this become a script or pyproject.toml rule that all agents can run with a single command? These should be run both when implementing and addressing reviews

CONTRIBUTING.md
davitf: Most of these could be moved to the same script to run all checks, right?

.claude/skills/address-review-findings/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+
+Then the three dependency configs from `CONTRIBUTING.md` §"Before pushing…" when behaviour
+depends on extras or versions, plus `openspec validate --all`,
+`scripts/check_docs_nav.py`, and `scripts/check_openspec_archived.py`.
```
davitf: Could those become a single script as well, or added to the same script? (maybe we should have separate checks and test scripts)

.claude/skills/address-review-findings/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+- **Say what you did not do, and why.** Deferred and disproven findings get a line each.
+- **Report gates honestly.** Say which legs you ran and what they returned. If you skipped
+  a leg, say that instead of implying it passed.
+- **Lead with what changed**, not with a recap of the review the maintainer already has.
+- Every comment ends with the attribution footer:
+
+  ```
+  ---
+  _Generated by [Claude Code](https://claude.ai/code)_
+  ```
```
davitf: This skill might also be run on Cursor or a different env, so don't hardcode Claude here please

.claude/skills/code-review-skill/reference/archivey-review-addendum.md
```diff
@@ -565,7 +573,65 @@ three-block output shape (§0).
...
+
+### Attribution
+
+Agents post through the maintainer's account, so **every posted comment ends with the
+attribution footer**:
+
+```
+---
+_Generated by [Claude Code](https://claude.ai/code)_
+```
```
davitf: same here

.claude/skills/code-review-skill/reference/archivey-review-addendum.md
```diff
@@ -565,7 +573,65 @@ three-block output shape (§0).
...
+
+Without it, an agent's review is indistinguishable from the maintainer's own comment, and
+the next agent cannot tell which feedback is the human's. That distinction changes how the
+feedback is weighted — see the responder skill's "Who actually said this".
+
+### Do not fix while reviewing
+
+`/code-review` reports; it does not edit (SKILL.md, repo default). Leaving the fix to the
```
davitf: /code-review-skill is the skill name. /code-review is a builtin Claude / Cursor skill that is not exactly what we use

dev-docs/code-map.md
```diff
@@ -0,0 +1,137 @@
...
+  └─ member selection                                internal/selection.py
+  └─ path safety + policy                            internal/filters.py
+  └─ bomb accounting                                 internal/extraction.py
+```
+
+Three things about this path are worth knowing before you debug it:
+
+- **Listing can happen twice.** Backends with an upfront index take an index-only pass
+  before materialization, so per-member work on the listing path may run twice with two
+  different member objects. Dedupe on the **member id**, which is stable across both
```
davitf: This is actually an interesting observation. Do you know if we have decided to create separate member objects instead of reusing, and why? I was assuming we reused them and updated them in-place, but we might have chosen the frozen path. Is this already documented?

.claude/skills/address-review-findings/SKILL.md
davitf: Yes — done, as `scripts/check.sh`. Taking your suggestion from the other thread to split checks from tests, since the cadences differ (seconds vs minutes).

```bash
./scripts/check.sh --fix    # every fast gate; --fix applies ruff first
```

It mirrors CI's `lint`, `docs` and `openspec` jobs in one command: `ruff check`, `ruff format --check`, `pyrefly`, `ty`, `check_openspec_archived.py`, `openspec validate --all`, `check_docs_nav.py`, and the strict docs build.

Two behaviours chosen deliberately:

- **It runs every gate even after one fails**, then lists what failed. One run should tell you everything that is wrong, not just the first thing — stopping early is how you fix pyrefly, re-push, and discover ty.
- **Without `--fix` it writes nothing**, so it answers "will CI pass?" rather than mutating the tree. `--fix` applies `ruff format` and `ruff check --fix` first, which is what you want before committing.

Shell script rather than a `pyproject.toml` rule: there is no task runner in the repo today, and adding one (poe/hatch/just) is a new dependency for something `scripts/*.sh` already does — that is the existing convention for `setup-dev-env.sh` and `install-git-hooks.sh`.

Verified both paths: green on the current tree, and with a deliberately broken file dropped in it collected and named both failures rather than stopping at the first.

Wired into the skill here, plus `CONTRIBUTING.md` §Getting started and AGENTS.md, so the seven-command list is no longer something to get half-right.

---

.claude/skills/address-review-findings/SKILL.md
davitf: Took the parenthetical — **separate checks and test scripts**, because the cadences really are different: `check.sh` is seconds and you run it constantly, `test.sh` is minutes and `--all-configs` is several.

```bash
./scripts/test.sh                       # the everyday [all] leg
./scripts/test.sh tests/test_zip.py -k roundtrip   # extra args pass through to pytest
./scripts/test.sh --all-configs         # the full before-pushing gate
```

The `--all-configs` mode is worth more than a shortcut, because it closes a documented trap rather than restating it. Leg 2 (`--resolution lowest-direct`) doesn't just *install* older versions, it **persists** them to `uv.lock` — so every later `uv sync --frozen` silently keeps the downgraded set, and `git status` shows a few hundred lines of churn that's easy to commit by accident. `CONTRIBUTING.md` warns about this and tells you to restore by hand; it still caught someone on #232, who had to revert the lockfile and re-sync before the final lint pass meant anything.

The script restores `uv.lock` and the everyday environment in an `EXIT` trap, so it happens on success, on failure, and on Ctrl-C. The warning stays in `CONTRIBUTING.md` for people running the legs by hand, now pointing at the script as the way to avoid it.

Verified: pytest passthrough works, and `--all-configs` with extra pytest args exits 2 with a clear message rather than silently running the whole suite three times.

---

CONTRIBUTING.md
davitf: Yes — this block now leads with the two scripts, and the individual commands stay below for when you want one in isolation.

```bash
./scripts/check.sh --fix   # seconds — every fast gate; --fix applies ruff first
./scripts/test.sh          # minutes — the everyday [all] test leg
```

The §"Before pushing…" section further down collapses to `./scripts/check.sh && ./scripts/test.sh --all-configs`, with the three legs kept underneath as documentation of what that runs.

I kept the underlying commands rather than deleting them, for two reasons: the lockfile explainer under leg 2 is genuinely useful knowledge (it says *why* the checkers are pinned exactly, which the script can't teach), and a script is a worse place to learn what the gate consists of than a list. The scripts are now the default path; the commands are the reference.

One line added while I was here: **"Both type checkers must stay green — `ruff` passing is not the gate"**, since that was the actual failure mode behind all of this.

---

.claude/skills/address-review-findings/SKILL.md
davitf: Fair — I'd baked one host's string into what is actually a host-independent rule. Reframed: the requirement is that a comment be **identifiable as agent-authored**, not that it carry a particular footer.

Two mechanisms now, either of which satisfies it:

- **A distinct bot identity.** If the host posts as itself (`cursor[bot]`, `qodo-code-review[bot]`), attribution is already unambiguous and no footer is needed — which is why the Cursor review on #236 reads correctly without one.
- **An attribution footer**, when posting through a human account: a rule plus one italic line naming the tool. The Claude Code string is given as *the example for Claude Code sessions*, not as the rule.

Also added the line that the point is identifiability, and: **don't sign a comment with a tool that didn't write it**.

Changed in the same spirit in the addendum §10, `.cursor/commands/address-review.md`, `.cursor/commands/code-review.md`, AGENTS.md §Review workflow, and the "Who actually said this" section of this skill — that last one was reading the footer as *the* discriminator, when a bot login is the more reliable one where it exists.

`CLAUDE.md` keeps the concrete Claude Code string, since that file is by definition Claude-Code-specific.

---

.claude/skills/code-review-skill/reference/archivey-review-addendum.md
davitf: Done here too — this was the §10 Attribution block, now rewritten around identifiability rather than one host's footer. Details in the [reply on the skill thread](https://github.com/davitf/archivey/pull/239#discussion_r3790435302).

---

.claude/skills/code-review-skill/reference/archivey-review-addendum.md
davitf: Corrected — and confirmed the collision is real, which makes this worse than a naming nit. This session's skill list has **both** `code-review-skill` (this repo's) and a builtin `code-review` described as *"Review the current diff, or a PR number/branch/path target… Pass `--comment` to post findings as inline PR comments, or `--fix` to apply the findings to the working tree."*

So a bare `/code-review` in Claude Code doesn't just load the wrong guidance — it loads a skill with a **different output shape and an edit mode**, in a repo whose review default is explicitly report-don't-edit. The wrong one is the more dangerous one.

Fixed to say the skill is `/code-review-skill`, with the Cursor project command noted as what makes `/code-review` land correctly *there* specifically:

- the addendum line you flagged (`/code-review` reports → `/code-review-skill` reports);
- `code-review-skill/SKILL.md` — both the repo note and the frontmatter `description`, which ended with "or when the user invokes /code-review" and so was advertising the colliding name as a trigger for itself;
- AGENTS.md §Review workflow and CLAUDE.md, which I had written as "Cursor: `/code-review`".

I left `.cursor/commands/code-review.md`'s own filename and its "Cursor `/code-review` defaults" heading alone — in Cursor that file *is* what `/code-review` resolves to, so it's accurate there.

---

dev-docs/code-map.md
davitf: Traced it rather than assuming, and the short answer is: **your assumption is right and is the documented decision; the separate objects are a different thing entirely, and nobody decided them.**

### The mutability question is settled, and not the way my note implied

ADR 0007 — *"Mutable `ArchiveMember` filled in place"*, status **"accepted (reversed from an earlier frozen draft)"**. Members are a mutable stdlib dataclass, the library fills late fields in place, callers use `member.replace(**kwargs)`, and they're unhashable so you key by `name` / `member_id`. The rationale is exactly the streaming case: some fields aren't known until member data is read (gzip/ZIP-data-descriptor sizes and CRCs, link targets stored in data), and under `streaming=True` the library can't re-materialize and hand back a new object.

So: we did **not** choose the frozen path. There was a frozen draft and it was reversed.

### The separate objects are an artifact of the two passes, not a design choice

`extract_all` on a backend with `_MEMBER_LIST_UPFRONT` lists twice, and **both passes call `_iter_members()` afresh**:

- `_get_members_index_only()` — `base_reader.py:1062` — `members = list(self._iter_members())`
- `_materialize_members()` — `base_reader.py:1000` — `for idx, member in enumerate(self._iter_members())`

The first pass's list is returned inside a `MemberListReport` (`base_reader.py:1555`) but is **never stored in `self._materialized`**. So when materialization runs, the cache is still empty, the backend re-walks its index, and constructs a second set of `ArchiveMember` objects for the same members.

Nothing chose that. It falls out of the index-only result not being cached, and it's orthogonal to mutable-vs-frozen — the objects are mutable and filled in place, there are just two generations of them.

### Is it documented?

Only as a consequence, and only where it bit: the inline comment at `base_reader.py:1089-1097`, written in #232 when the duplicate `MEMBER_NAME_BIDI_CONTROL` was fixed. It documents the *rule* ("dedupe on the member id, not on this object") but not why two generations exist. No ADR, no spec scenario.

### What I changed

You read the bullet as describing a decision, which is a fair reading and my fault — it sat between two genuine decisions (the translation contract, ADR 0012) and inherited their tone. Rewritten to say plainly that it's an artifact, cite ADR 0007 for the mutability contract so the next reader doesn't have the same doubt, and keep the operational rule (dedupe on id) as the reason it's worth knowing.

The open question your comment actually surfaces — *should materialization reuse the index-only list instead of rebuilding?* — is now recorded in `dev-docs/IDEAS.md` under Performance & robustness rather than left in this thread. It'd remove both the wasted re-walk and the identity trap. Two things want care if we do it: the listing tracker deliberately **does** re-account on the second pass (`base_reader.py:1081-1085`), and `_materialize_members` has a concurrency gate around publishing. Not touched here.

---

davitf:
## Review round 1 — `8094224`

All seven comments addressed. Replied in each thread and resolved them; reopen any you disagree with.

| # | Thread | Disposition |
|---|---|---|
| 1 | Gate as a single script (SKILL.md) | ✅ `scripts/check.sh` |
| 2 | Same for the CONTRIBUTING command block | ✅ Block leads with the scripts |
| 3 | Single script, or separate check/test scripts? | ✅ Separate — `check.sh` + `test.sh` |
| 4 | Don't hardcode Claude (SKILL.md) | ✅ Rule is now identifiability |
| 5 | Same (addendum §10) | ✅ Same |
| 6 | `/code-review-skill` is the name | ✅ Corrected in 4 files |
| 7 | Member objects: decided? documented? | ✅ Answered — artifact, not decision; ADR 0007 cited; open question → `IDEAS.md` |

### The three that changed something beyond wording

**Two scripts, split by cadence** (1–3). `./scripts/check.sh --fix` runs every fast gate — ruff, **pyrefly**, **ty**, openspec archived + validate, docs nav, strict docs build — and `./scripts/test.sh [--all-configs]` runs the suite. Two deliberate behaviours: `check.sh` keeps going after a failure and lists everything that failed, because stopping early is how you fix pyrefly, re-push, and discover ty; and `test.sh --all-configs` restores `uv.lock` and your environment in an `EXIT` trap, so the `[all-lowest]` leg can't leave a downgraded resolution behind. That trap was documented and still caught someone on #232 — it's now handled rather than described. Shell scripts rather than a `pyproject.toml` runner: none exists here, and `scripts/*.sh` is the convention.

**The skill-name collision is real and worse than a nit** (6). This session's skill list carries both `code-review-skill` and a *builtin* `code-review` that "pass `--fix` to apply the findings to the working tree". So a bare `/code-review` loads a skill with a different output shape **and an edit mode**, in a repo whose review default is explicitly report-don't-edit. Also fixed the frontmatter `description`, which ended "or when the user invokes /code-review" — it was advertising the colliding name as its own trigger.

**The member-object note was right but framed wrong** (7). You asked whether separate objects were a decision and whether we'd gone frozen. Traced it: ADR 0007 is *"Mutable `ArchiveMember` filled in place"*, status "accepted (**reversed from an earlier frozen draft**)" — so your assumption is the documented decision. The separate objects are unrelated: `_get_members_index_only()` and `_materialize_members()` each call `_iter_members()` afresh, and the first pass's list is never stored in `self._materialized`, so the backend re-walks and rebuilds. Nobody decided that. My bullet sat between two genuine decisions and inherited their tone; it now says artifact-not-decision and cites ADR 0007. The open question — should materialization reuse the index-only list? — is recorded in `IDEAS.md` rather than left in a thread.

### Gates

`./scripts/check.sh` green: ruff check, ruff format, pyrefly, ty, `check_openspec_archived.py`, `openspec validate --all` (25/25), `check_docs_nav.py`, strict docs build.

**Not run: the test suite.** No Python changed this round — the diff is two new shell scripts and markdown — so `pytest` had nothing to exercise. Saying so rather than implying a green suite. CI on `1493662` was 23/23 including the Windows and macOS legs.

### One thing worth your eye

`scripts/check.sh` runs `openspec validate --all` only when the `openspec` CLI is on `PATH`, and prints `SKIPPED` otherwise rather than failing — so a bare clone without `setup-dev-env.sh` still gets the other gates. CI has no such escape hatch. If you'd rather a missing CLI be a hard local failure, that's a one-line change.

---

PR #240, opened 2026-08-15
review: commission Topic 8 (documentation content) and Topic 10 (the problem catalogue)

davitf:
Commissions two review topics and updates the three index files that track review state.

- **Topic 8** — `review/docs-content/`: documentation *content*, the counterpart to the docs IA review. That one decided where each page lives; this one decides whether the page is right.
- **Topic 10** — `review/problem-catalogue/`: one entry per non-trivial problem the project has had to solve, stated solution-neutrally. Not a review — an aggregation over what is already written.

They run in parallel on disjoint sources.

---

# Topic 8 — documentation content

## Why now

The docs prose paused in July because writing it kept turning up library defects. Those were paid — #225, Topic 9's W1–W9 in #232, and the diagnostics follow-ons #233–#236 — and Topic 9 archived on 2026-08-15. Nothing in flight moves the described surface again.

That also makes the highest-value pass a **diff rather than a re-read**: `fe6d4a7..d4668c3` is the window in which the pages were written and the behaviour changed underneath them. The pages were kept in sync opportunistically as those PRs landed (+226/−27 lines across 9 files), but only on the surfaces each PR happened to touch.

## What the brief carries

Measured at `d4668c3`, not sketched:

- **15 published pages / 2 108 lines** against the 16 `outline.md` specifies. The missing one is `how-it-works.md` — and adding it needs a `documentation` spec delta, not just a file: `openspec/specs/documentation/spec.md:78-93` enumerates the narrative pages and requires every file under `docs/` to carry a nav entry.
- **87 names in `__all__`, 56 with an `api.md` entry.** The 31 without include 21 of the exception types — the names a caller needs when writing `except`. Filed as a **question, not a defect**: 87 mkdocstrings entries may be the wrong page, and whichever way it is decided should become a guardrail.
- **39 code blocks, 35 of them `python`, none executed by CI.** That is the guardrail this pass should leave behind.
- **`outline.md`'s "~455 lines outstanding" is stale.** Re-tally before planning against it; a stale worklist was finding 5 of #223's own re-review.

Seeds were verified rather than assumed: O-2 is fixed, "what `TRUSTED` does not relax" is written — and the bounded-recursion recipe, `install.md`'s `format_availability()` section, `access-and-cost.md`'s measurement section, and the error-translation narrative are not. O-15 and O-9 are confirmed open.

## Scope and definition of done

No budget and no target date, so the triage framing inherited from `backlog.md` is gone. The ranking survives as a **sequence** — accuracy → gaps → register → quality — each pass making the next cheaper. Pass 4 goes last because on a page whose claims are unverified a structure argument is unfalsifiable, not because it is the one to drop.

§Definition of done replaces the budget escape with nine conditions, so "we did passes 1–3 and skipped 4" and "we stopped partway through pass 2" cannot look alike in the record.

## How to run it

§How to run this records the decomposition rather than leaving it re-derived every session:

- **Pass 1 splits by capability, never by page.** One behaviour lands on several pages, so a worker per page re-derives the same subsystem and can reach *different* answers about it — which is O-2 exactly. Passes 2–4 do split by page.
- **Fan out by input:output ratio.** Verification reads whole subsystems to produce a dozen rows; the claim inventory and the writing stay central.
- **Eight capability clusters** covering 20 of the 24 specs; the other four are named with the reason they have no owner.
- **Four safety rules**, including that the cross-page consistency pass is mandatory whatever the topology — #223's round-2 review found four contradictions the splits change created *in a single pass by a single agent*.

**This review writes**, unlike Topic 9: page-sized PRs, with a checkpoint after the claim table. The constraint that replaces analysis-first is that a library defect found while writing never gets fixed inside a docs PR.

---

# Topic 10 — the problem catalogue

## Not a review, an aggregation

The material is already written across ~180 documents that nothing collects: **72** archived change proposals with a `## Why`, **57** `design.md` files, **18** ADRs, **11** review summaries, **8** investigations, the threat model's nine `O` entries, `known-issues.md`, `library-analysis.md`, `open-issues.md`, and `dev-docs/history/`. One problem is currently restated in three vocabularies across a review finding, an ADR context section and a change proposal, and cannot be counted.

## Problems, not decisions — but each links to its decision

Field 4 holds how archivey answers it today and which ADR or change decided it, and is **strippable**. That structure exists because the second consumer is a **fresh-design comparison**: hand a frontier model the problems alone, ask for an architecture, compare against ours.

The brief writes that protocol down now, because it is what keeps the schema honest — including the grading rule that sorts every divergence into "already considered, cite the ADR" and "not considered", and the caveat that a model trained on archive-handling libraries shares our priors, so **convergence is weak evidence and divergence is the signal** (the #208 and #230/#231 lesson).

## The neutrality rule

A problem stated in terms of its solution is a leading question. *"Avoid copying bytes twice in the decoder stack"* presupposes a decoder stack; *"verifying a checksum and delivering the bytes are the same read"* does not. The test: a statement naming an archivey type, module or config field is a solution and needs rewriting. Three worked examples are in the brief, and `dev-docs/history/` is read first because it predates the current design and is therefore natively neutral.

## Topic 8 harvests the residue

Topic 8's capability workers now return the problems they meet in each subsystem, filed under `problem-catalogue/harvest/`. That residue is **not keyword-findable** — `workaround|quirk` matches four sites in `src/` while 136 comments name an upstream library — so it is reachable only by someone already reading the code, and those workers are the only readers who will be. Capped deliberately: a bounded list, no investigation.

---

## Navigating the labels

Topic 8's brief inherits five numbering schemes from four programmes, and two pairs collide: `O-16` is a documentation observation while `O6` is a threat-model residual (the hyphen is the whole distinction), and "phase 3" is the IA review's execute phase while "Phase 8" is a capability implementation phase. §Where the labels are defined tables all nine families with the file and line range that defines each; Topic 10's brief points at it rather than duplicating it.

## Sequencing note

`review/docs/` stays in flight alongside both. Its last deliverable is `how-it-works.md`, which D2 assigns to Topic 8, and its `outline.md` / `observations.md` are Topic 8's inputs — archiving them together avoids rewriting every citation.

## Verification

Docs-only; no code touched. `uv run --group docs python scripts/check_docs_nav.py` passes (15 pages, all in nav, all links and anchors resolve); every relative link across the six changed files resolves on disk; and every count, capability spec, line range and symbol name cited in either brief was checked against the tree rather than recalled.


davitf:
## Review round 1 — all six findings dispositioned

Review arrived out-of-band (the reviewing agent could not post), so it is recorded here rather than in threads. Every finding was verified against the tree before acting; **all six were real** — no false positives this round. Fixed in `74ca450`.

| # | Disposition | What changed |
|---|---|---|
| **F1** | **Fixed** | The O9 seed rewritten; O9 added to §Provenance as settled |
| **F2** | **Fixed** | 17 ADRs, not 18 — corrected in all three places |
| **F3** | **Decided + recorded** | `how-it-works.md` does **not** wait on the catalogue |
| **F4** | **Fixed** | Label row covers Topic 4–10; brittle line cite dropped |
| **F5** | **Fixed** | Harvest link text and href now agree |
| **F6** | **Fixed** | "implemented, this PR" → "implemented in #223, archived in #224" |

### F1 — the seed was re-opening a closed register entry

Confirmed, and worse than the row suggested: the seed was wrong on **both** halves. O9 was *registered* inside #235 and *closed* by #236, whose escaping moved to message construction — `ArchiveyError` / `ArchiveyUsageError` escape in `__init__`, `Diagnostic` escapes its `message`, and the CLI-side `logging.Formatter` was removed because two layers double every backslash. `threat-model.md` titles the entry **implemented**.

Left as written, workers would have re-litigated a closed entry — precisely the failure §Provenance exists to prevent, in the brief that defines that section.

Rewritten to the question that *is* open: escaping-at-construction is a caller-visible contract (`error-handling` and `diagnostics` both require messages *"inert for terminal display"*) and **no page states it**. The thin-`cli.md` sweep is kept. §Provenance now carries the #235-registered / #236-closed split so the next reader doesn't reconstruct it.

### F2 — a "measured" denominator that wasn't

`dev-docs/decisions/` holds **17** numbered ADRs plus `index.md`. Corrected in the Topic 10 brief, `STATUS.md` and `backlog.md`, with `index.md` noted as a router rather than a source. "~180 documents" survives either way.

### F3 — decided: no gate

Maintainer's call: **write `how-it-works.md` from D2's own sources whenever Topic 8 reaches it; fold catalogue material in later.**

Reading D2 while framing the question made it easy — D2 already names a source for each of the page's six sections (`VISION.md`, ADRs 0001/0002/0003/0006, `library-analysis.md`, the `backend-registry` spec, `dev-docs/decisions/`), and **none of them is the catalogue**. The gate was never needed. Recorded in three places, and Topic 10's brief now says explicitly that nothing in it may be written so as to block Topic 8's §Definition of done.

### Gates

- `./scripts/check.sh --fix` — **passes**: 25/25 items, plus docs nav (*15 pages, all in nav; repo, site and anchor links all resolve*) and docs build.
- Relative links across all six changed files resolve.
- Branch CI was **23/23 green** at `27462d3`; this round changes only markdown under `review/`, so no test leg is implicated. `./scripts/test.sh` not run for that reason — say if you'd rather see it.

Nothing deferred, nothing disproven.

---

PR #241, opened 2026-08-17
review(docs-content): Topic 8 pass 0 — scope.md (D-f routing)

davitf:
## Summary

Pass 0 of Topic 8 (documentation content). Produces `review/docs-content/scope.md` only — routing decisions, no guide prose.

For each of the 16 guide pages (15 under `docs/` plus the not-yet-written `how-it-works.md`), the file records:

1. The page's job (one sentence)
2. What is explicitly not on it, and where that material goes
3. Every current block routed by the D-f test from `review/docs/outline.md`

It also re-derives the brief's §B worklist under the same routing: several growth rows dissolve (already shipped in #224/#235) or convert to docstring / threat-model tasks; `extracting.md` becomes a **cut** (228 → ~110–130), not ~90 new lines.

## Checkpoint

This PR is the step-4 handoff: please steer on the four confirmation questions at the bottom of `scope.md` before pass 1 (claim inventory) starts.

## Test plan

- [x] `./scripts/check.sh --fix`
- [x] `uv run --group docs python scripts/check_docs_nav.py`

<div><a href="https://cursor.com/agents/bc-d21e2049-3caa-46ca-ab93-c6ecb91b07a9?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d21e2049-3caa-46ca-ab93-c6ecb91b07a9&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #242, opened 2026-08-17
review(docs-content): pass 0 — route all 16 guide pages under D-f

davitf:
Pass 0 of Topic 8 (`review/docs-content/brief.md` §Suggested process step 1), and only that step. Adds `review/docs-content/scope.md`.

For each of the 16 pages in the guide — the 15 under `docs/` plus `how-it-works.md`, which does not exist — it records a one-sentence job, an explicit non-coverage list saying where each excluded thing goes, and every current block routed by [`outline.md` D-f](https://github.com/davitf/archivey/blob/main/review/docs/outline.md): guide / `dev-docs/threat-model.md` / docstring. `extracting.md` is routed block by block against D-f's 228 → ~110–130 target.

**No guide prose is written and no file under `docs/` or `src/` is edited.** Prose is step 6.

## What it found

**The docstring leg does not work today.** D-f routes lookups to "the docstring, surfaced through `api.md`". Its counts are exact and unchanged (`ExtractionStatus` 1 line, `AbortOn` 13, `ExtractionResult` 5) — what they understate is that the missing depth already exists in `extraction_types.py` as `#` comments beside each member, and mkdocstrings renders none of it (`EnumMembersAsTable` reads `m.docstring`; `griffe_fieldz` does the same for dataclass fields). Verified with griffe: all six `ExtractionStatus` values, all three `AbortOn` members and all nine `ExtractionResult` fields have no docstring. So "move it to the docstring" is a `src/` write task, not a move.

**The §B worklist re-derives from ~455 lines to ~174.** Row 6 (`opening-and-listing.md` / `reading-members.md` remainders) dissolves entirely — each of its six items is cited at the line where it shipped, and both pages now need cuts rather than additions. Halves of rows 2–5 dissolve the same way, mostly into `#224`/`#225`/`#232`/`#235`; D-f converts about 30 more lines into docstring and threat-model tasks. Three rows the outline never carried are added, including `cli.md`'s two items that sit in `outline.md` §10 but never reached the §B table.

**Where the cuts concentrate:** `extracting.md` carries 36% of the reduction, 46 lines of it one ruling — §What is enforced is a threat-model inventory rendered as user documentation. Four more pages lose most of their cut to duplication rather than depth. Two pages (`philosophy.md`, `acknowledgements.md`) are net-zero and recorded as such so a later pass does not re-derive them.

Routing alone lands the guide at ~1 982 lines against D-f's ~1 600–1 800 projection. That is not a disagreement: ~40 blocks are ruled *trim*, not moved, and that shrink is pass 3's O-17 register work. Counting it here would spend it twice.

## Reconciled against the second pass (#241)

A second agent ran pass 0 independently. `scope.md` §Cross-check records both halves, since agreement between passes with shared priors is weak evidence and divergence is the informative part.

**Converged without contact:** `extracting.md` as the primary cut and §What is enforced as its bulk; `abort_on` / `NAME_SANITIZED` → docstrings; ADR/PR citations and the symlink layer count out; the bidi paragraph stays shorter; `formats.md`'s `hashlib` loop → ~8; **`support-matrix.md` §Free-threaded stays**; config-at-a-glance → `ArchiveyConfig` docstring; error-translation and measurement are real writes; §B row 6 dissolves. Two findings were found twice.

**Corrections taken from #241** — the stale `archivey-2` link is **O-4, already open**, so my S-1 was a rediscovery; must-explain **#16 is half open**, not closed (the page gives the `AUTO` size threshold but never says `ON` raises while `AUTO` falls back silently); `formats.md` §ISO must state the pycdlib patch because `gotchas.md:90` links there and the landing is silent; §Trust boundaries keeps two clauses rather than leaving whole. One further revision on re-reading my own rule: §RAR listing cost stays on `access-and-cost.md`, because a cost fact's owner is the cost page.

## All six scope questions are decided (2026-08-17)

Nothing in pass 0 is blocked. Each is recorded in `scope.md` §Questions with its reasoning rather than compressed to a verdict, because a fresh container has no memory and the argument is what stops it being re-run.

| | Decision | Effect |
|---|---|---|
| **Q1** | The docstring leg **rides with the page PR** | "No library changes in a docs PR" is read as being about behaviour, under a stated carve-out: a docs PR may touch `src/` only to add or reword a docstring, with no change to any statement, signature, default, annotation or control flow. Anything else stays `#225`'s shape |
| **Q2** | `support-matrix.md` §Free-threaded **kept at ~65** | 105 → ~65, page 152 → ~115. Cuts cross-page duplication only — no free-threading fact leaves. Closes the one block D-f explicitly declined to rule on |
| **Q3** | §D **settled before pass 6 writes the errors page** | Timing decided, shape still open. The errors page gets written once against a known reference surface; the claim inventory produces the evidence that picks the shape |
| **Q4** | `install.md` matrix re-cut is **written**, bounded to a four-row extra → formats re-index | install.md 34 → ~52; `formats.md` stays authoritative on capability |
| **Q5** | Bounded-recursion recipe stays a **pointer** | Supersedes the brief's §B row rather than leaving it unwritten; the amplification analysis stays threat-model O6's |
| **Q6** | **No fifth Home recipe** | Recorded as dropped: the founding use case has no first-screen presence, and `formats.md`'s trimmed snippet becomes the guide's only dedupe example, which raises its floor |

Q4–Q6 were the live divergences with #241, and all three went against this pass.

## Scope discipline

- D-a–D-e untouched. Every "→ page" ruling moves a *duplicate* to the page that already owns the fact; none moves an owner.
- `outline.md`'s projected-sizes table not used as targets (it carries a note saying so, and targets `extracting.md` at ~280).
- No claim verified — "shipped" means prose exists at a cited line, not that it is true. That is `claims.md`'s job.

## Checks

- `./scripts/check.sh --fix` — all checks passed
- `uv run --group docs python scripts/check_docs_nav.py` — 15 pages, all in nav; repo, site and anchor links all resolve
- Environment verified per `CLAUDE.md`: `unrar` and `7z` both present

## Next

Pass 0 is complete and unblocked. Step 3 — the claim inventory — is the next unit of work, and it now has one extra deliverable: the evidence that settles §D's shape before the errors page is written.


davitf:
## Review round 1 — all findings dispositioned

The handoff was relayed out of band rather than posted here, so this comment carries the record. No inline threads or PR comments existed to reply in; CI was green (23/23) before and after.

| # | Disposition | What changed |
|---|---|---|
| **F1** | **Fixed** — `0216bc8`..`6fa844a` | The Row 2 essay said the install matrix re-cut "dissolves outright" while the §B table and Q4 restore it. The paragraph now records the argument, the decision that overrode it, and the ~12 lines outstanding |
| **F2** | **Fixed, and it was sharper than reported** | See below |
| **F3** | **Fixed** | Row 4 listed the dropped recursion recipe as open work and totalled "~15". Open work is the config-ceiling rule alone, ~3 |
| **F4** | **Fixed** | `review/README.md` claimed a step-4 steer was awaited (step 4 is the *post-claims* checkpoint) and carried the pre-Q6 ~177. Now: pass 0 done, six questions decided, next is step 3 |
| **F5** | **Fixed, both halves** | Guide total was stated as both 1 982 and 1 985. Exception count corrected to **21 of 26** — counted rather than assumed: 27 `*Error` names in `__all__`, less `OnError` which is an enum, and `api.md` lists five |

### F2 was a bigger problem than the finding said

Removing the stale "~12 added back as the recursion recipe" clause exposed that the itemization underneath it **never summed**. It listed ~46 + ~26 + ~10 + ~14 = 96 lines of destinations against an asserted 108-line cut, and the +12 inbound made the gap worse rather than explaining it. The number was carried, not computed.

Re-tallied from the per-block rulings: **~82 lines** leave `extracting.md`, split ~32 threat-model / ~25 docstring / ~2 errors page / ~8 dropped as duplicates / ~15 absorbed into the one-clause rewrites that stay. That means **routing alone lands the page at ~145, not inside D-f's ~110–130** — the band is reached once pass 3 runs O-17 on what survives. That is now stated plainly on the page rather than rounded away, and it matches how the guide-wide arithmetic was already framed.

Guide-wide figures restated as estimates summed from per-page rulings, with the tolerance named: **≈ −275 out, ≈ 2 007 after routing** (was ≈ −300 / ≈ 1 982).

### One the review did not catch

The "three rows are new" paragraph still justified the Home dedupe recipe that **Q6 dropped** — "the cut is not complete without it" — and omitted row 10 entirely. Both fixed. The dropped row stays in the §B table at **zero** rather than disappearing, so the drop is visible to the next reader instead of looking like an oversight.

### Maintainer decisions — both taken, both A

- **`brief.md` §Hard constraints amended**, dated and citing `scope.md` Q1: the "no library changes in a docs PR" rule is about *behaviour*, so a docs PR may touch `src/` solely to add or reword a docstring, with no change to any statement, signature, default, annotation or control flow. Without it a page-PR author meets an absolute prohibition on the exact edit D-f's docstring leg requires.
- **`STATUS.md`'s Topic 8 row updated** — it still said the ~455-line worklist needed re-tallying, which is the work pass 0 did. The historical "what ships thin" paragraph keeps its original figure as a record of that change, with a dated pointer forward rather than a rewrite.

### Gates

Run here, not copied from CI: `./scripts/check.sh --fix` → all checks passed. `./scripts/test.sh` → **2453 passed, 23 skipped, 3 deselected** in 99s. `uv run --group docs python scripts/check_docs_nav.py` → 15 pages, all in nav, links resolve. `unrar` and `7z` both present, so the suite is not skipping ~109 tests quietly. `--all-configs` not run: this PR changes only Markdown under `review/`, so no behaviour depends on the dependency config.

Nothing deferred, nothing disproven — all five findings were real.

---

PR #246, opened 2026-08-17
review(docs-content): the baseline and the claim inventory (Topic 8, steps 2–3)

davitf:
Adds `review/docs-content/claims.md` — Topic 8 steps 2 and 3. One new file, nothing else touched.

Pass 0 ([#242](https://github.com/davitf/archivey/pull/242)) routed every block on all 16 pages and verified none of them: every "shipped" in `scope.md` means prose exists at a cited line, not that the prose is true. This is the artifact that makes each of those lines checkable, and the shared state the capability workers report against.

**This stops at the step-4 checkpoint.** No fan-out, and every verdict cell is empty by design.

## Step 2 — the baseline

`scripts/setup-dev-env.sh` verified `unrar` and `7z` present, so the ~109 quiet skips `CLAUDE.md` warns about did not happen — 23 skips total, itemised in the file, none of them a missing format backend.

| Gate | Result |
|---|---|
| `./scripts/check.sh` | **PASS** — all seven gates, incl. `docs/: 15 pages, all in nav; repo, site and anchor links all resolve` and 25/25 openspec items |
| `./scripts/test.sh` | **PASS** — 2453 passed, 23 skipped, 3 deselected, 88% coverage |
| `format_availability()` | **FULL for all 22 formats** |

The `format_availability()` sweep is recorded per the brief because it closes off an excuse: nothing is unavailable in this session, so *format unavailability is not an admissible "unverifiable" reason for any row*. A worker writing that verdict has to name a different reason.

Measured on **`[all]`**. The 12 rows whose truth depends on an optional library are marked `cfg`; `[all-lowest]` and `[core-only]` were not run, and that is stated rather than left implicit.

## Step 3 — the inventory

**400 rows, grouped by capability, not by page**, across the brief's eight clusters plus a ninth for the coordinator-owned `documentation` spec. Columns: the claim stated so it can be true or false · every `page:line` that states it · the `src/` or `openspec/specs/` line that would settle it · `scope.md`'s routing ruling · an empty verdict.

Grouping by page would have produced **573** rows instead of 400 and let four workers give four different answers about one behaviour — O-2 exactly, which is the defect the review exists to remove.

- **134 rows cite more than one page.** Heaviest overlaps: `formats`+`gotchas` (14), `access-and-cost`+`formats` (12), `extracting`+`gotchas` (11).
- **All 39 fenced code blocks** (35 `python`, 4 `bash`) are covered by 34 `[code]` rows.
- **8 `[TM]` rows** are recorded but out of scope to verify here, so the threat model does not inherit unverified assertions silently.
- **Ten rows have no page**: `scope.md`'s re-derived §B worklist written as rows rather than as a worklist, so a re-tally cannot lose them.
- **`how-it-works.md` contributes zero rows and that is correct** — a page with no prose makes no claim. Recorded so it does not read as an omission.
- Blocks ruled `Cut` get no row; each is listed under §Recorded drops so the drop stays visible.

Carried in from `scope.md` §Findings, not re-derived: **S-1** (the stale `archivey-2` nightly link — this is **O-4**, already open, *not* filed again) as **F-3**, and **S-2** (the policy table's missing `STANDARD` row) as **C-42**.

## Found while extracting — recorded as rows, none resolved

- **Two live cross-page contradictions**, neither in `scope.md` §Findings. **A-16**: `opening-and-listing.md:66-68` says ZIP, 7z, RAR and ISO must seek; `access-and-cost.md:145-146` says only ZIP and ISO. **A-33**: a missing codec backend raises `UnsupportedFormatError` on one page and `PackageNotInstalledError` on another.
- **Three completeness claims nobody is counting.** The exception table names **12 of 26** types and `api.md` renders **5** (D-8 / I-3) — that count is the evidence `scope.md` Q3 said would settle §D's shape. The `CostReceipt` table omits its public `notes` field (F-7), and since that block is ruled `→ DS`, a docstring promotion would carry the omission with it.
- **Two dangling references** (I-24): `access-and-cost.md:33` and `acknowledgements.md:55` cite `IDEAS.md` as a bare filename; the file is `dev-docs/IDEAS.md`, outside the site by D1/D3. `check_docs_nav.py` does not catch it because it is prose, not a link.
- **One baseline/page disagreement is paused on, not settled** (G-6): Python 3.11 here reports `ZST` as `FULL` with an empty `missing`, while neither package the pages name shows in `importlib.metadata`. Per `CONTRIBUTING.md` and the brief, a spec/design discrepancy gets asked about rather than picked silently.

## Where the suspect claims concentrate

The short read for the checkpoint, ranked by expected defect density rather than row count: **`cli.md`** (thinnest page, largest recent change — three of its 17 rows sit on `#236`), **`extracting.md` §What is enforced** (the eight `[TM]` rows), **`migrating.md`** (the one page `#235` did not sync), then the two contradictions, the ISO link with no landing, and the completeness claims.

## Constraints observed

No guide prose written and no file under `docs/` edited · no claim verified and no verdict filled · no library defect fixed · no fan-out · nothing under `src/` touched.

---
_Generated by [Claude Code](https://claude.ai/code/session_012KNZDUN7TVr5LMjSjyFkAQ)_

### DIFF
--- a/review/docs-content/claims.md
+++ b/review/docs-content/claims.md
@@ -1,0 +1,899 @@
+# Claims — steps 2 and 3 of Topic 8
+
+The baseline this pass ran on, and every checkable claim the published guide makes,
+grouped by **capability** and deduplicated across pages.
+
+Written against `main` @ `adb2e3f` (`d4668c3` + `#240`/`#242`/`#245`, none of which
+touched a file under `docs/`). Guide today: **15 pages, 2 108 lines**; the sixteenth,
+`how-it-works.md`, still does not exist.
+
+Inputs: [`brief.md`](brief.md) (the specification) and [`scope.md`](scope.md) (pass 0,
+merged in `#242`). `scope.md` is a **settled input** — its routing rulings and its six
+decided questions are carried in here, not re-derived.
+
+**This document verifies nothing.** Every `Verdict` cell is empty by design: the
+capability workers fill them after the step-4 checkpoint. What is settled here is *what
+the guide asserts*, *where each assertion appears*, and *which line of `src/` or
+`openspec/specs/` would decide it*.
+
+---
+
+# Part 1 — Baseline (step 2)
+
+## Environment
+
+`scripts/setup-dev-env.sh` run 2026-08-17; its closing verification block read:
+
+```
+=== verification
+ok   unrar: /usr/bin/unrar
+ok   7z: /usr/bin/7z
+ok   benchmark toolchain complete
+```
+
+Both named tools are present, so the ~109 quiet skips `CLAUDE.md` warns about did **not**
+happen in this session (23 skips total, itemised below). The apt step emitted two
+`403 Forbidden` proxy warnings for the `deadsnakes` and `ondrej/php` PPAs; neither
+supplies a package this repo uses, and every package the script asks for resolved.
+
+| | |
+|---|---|
+| Python | 3.11.15 (CPython, GIL-ed) |
+| Platform | `Linux-6.18.5-fc-v20-x86_64-with-glibc2.39` |
+| Dependency config | **`[all]`** (`uv sync --group dev --extra all`) — the everyday leg |
+| `archivey.__version__` | `0.2.0.dev0` |
+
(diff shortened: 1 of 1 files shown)


davitf:
merged into #248

PR #251, opened 2026-08-17
openspec: bounded source spooling — one limit for every temp file we write

davitf:
Specs-first proposal. **Nothing is implemented.** Four questions stay open in `design.md` for you.

Rewritten after review: the first draft split spooling into two kinds with asymmetric defaults, and that was wrong. Directory renamed `opt-in-source-spooling` → `bounded-source-spooling`, since "opt-in" now describes one of three settings rather than the design.

## Why

`unrar` takes a path, not a Python object (ADR 0002), so a RAR read from a stream **already** writes the whole archive to a temp file today — no limit, no configuration, and nothing in either honesty channel:

```
_ensure_archive_path calls: 1        -> /tmp/tmpqee8ey8z.rar   (whole archive)
reader.diagnostics: []               reader.cost: notes=()   (identical to a path source)
```

That's P11. Separately, ZIP/7z/RAR/ISO all refuse a pipe, because ADR 0010 refused to buffer one silently.

## The design: one limit, three values

```
spool_limit = <bytes>              # spool when needed, up to this much
spool_limit = SpoolLimit.UNLIMITED # never refuse on size
spool_limit = None                 # never spool
```

Nothing in the public surface records *why* a spool was needed.

**Why not two kinds.** The first draft separated a "tool-tax" spool (seekable source → path for `unrar`) from a "capability" spool (non-seekable source → seekable). The distinction is real inside the library and **invisible outside it**: both write the same bytes to the same directory at the same cost, both are avoided by the same remedy (pass a path), both are made safe by the same bound. The asymmetric-preference test kills it — *"I want RAR-from-stream but never pipe buffering"* is answered by the limit, and the reverse preference is incoherent. The split existed to avoid breaking RAR-from-stream reads, which Topic 9's brief explicitly rules out as a reason: *"Behaviour churn is free until the `0.2.0` tag."*

**Why not a boolean.** A boolean forces the default to be all-or-nothing; a modest number lets ordinary archives keep working from streams *and* pipes while a 40 GB one fails loudly. And `None`/`UNLIMITED` already say what `False`/`True` would, matching the `ExtractionLimits.UNLIMITED` pattern.

## What survived the collapse: timing

The one objection to merging the kinds was *when* the file gets written — and it's real. A spool for an external binary happens at the first read that needs it, so an archive can be **listed** without paying for it; a spool to make a source seekable happens at open. That's caller-visible, so it's now a requirement with its own scenario matrix, and `CostReceipt.notes` makes it observable. Specified, not turned into a knob.

## P11's open question, answered

**`CostReceipt.notes`, not a diagnostic.** A spool inside a limit the caller set is *declared*, so the `diagnostics` admission clause doesn't apply; the placement clause prefers a structured field, and `notes` is it. The reasoning inverts for today's unbounded, unreported behaviour — which is precisely why P11 is a defect.

## Free space is a heuristic; the limit is the contract

`TMPDIR` may be a different device, free space is a race, containers report host figures — and **a memory-backed temp dir turns "spool to disk" into ADR 0010's unbounded memory use**, reached through the door marked disk. Checked on this container: `/tmp` is `ext4`, so that's a known risk rather than an observed one. The caller can name the spool directory for exactly that reason.

## Out of scope, with a better framing

Caching *decompressed* payloads to speed seeking is a **caller-side wrapper-stream** concern — something archivey might ship or recommend later, not a deferred version of this change. `design.md` has the table; `tasks.md` records it that way for `IDEAS.md`.

## Open questions

1. **The default limit** — the load-bearing one, since it decides which of today's working RAR-from-stream reads start failing. A fixed size (1 GiB keeps ~everything working; 64 MiB makes the bound meaningful), `UNLIMITED` (nothing breaks and the *reporting* becomes the whole fix), or `None` (safest, removes a working capability).
2. **Packaging** — two flat fields or one frozen `SpoolLimits` object matching the `ExtractionLimits`/`ListingLimits` family.
3. **Which error when spooling is refused** — `ArchiveyUsageError` or `UnsupportedOperationError`. They sit on opposite sides of the `ArchiveyError` boundary (ADR 0012), so it isn't cosmetic.
4. **Whether `streaming=True` over a spooled source** streams from the spooled file or switches to random access.

## Impact on Topic 8

**Should not block it.** `claims.md` **E-71** states today's behaviour minimally and gets rewritten when this lands.

## Verification

`openspec validate bounded-source-spooling --strict` → valid. `./scripts/check.sh --fix` passes. Deltas against `access-mode-and-cost`, `archive-reading`, `format-rar`; P11 cross-referenced.


davitf:
## Review — `bounded-source-spooling` (OpenSpec proposal)

Reviewed at head `f7a3e02` with `/code-review-skill`, using the addendum's §9 values-first order (the artifact is a proposal, not a diff).

**What I actually ran** (not copied from the PR body): `openspec validate bounded-source-spooling --strict` → valid. `openspec archive bounded-source-spooling -y` in a throwaway worktree → **aborts** (see F1). PR CI 23/23 green; I did not re-run pytest — this PR touches no code.

---

# 1. Maintainer briefing

**What this change is.** A specs-first OpenSpec proposal (`openspec/changes/bounded-source-spooling/`, ~509 added lines, no code) that puts every temp-file copy of an *archive source* under one configured byte limit on `ArchiveyConfig` — three settings: a byte count, `UNLIMITED`, or `None` (never). It bounds and reports today's unreported RAR-from-stream whole-archive copy (P11), and as a side effect lets seek-requiring formats (ZIP/7z/RAR/ISO) open from a pipe when the limit permits. Four questions are left open in `design.md`. It also adds a note to `dev-docs/open-issues.md` P11.

**Snapshot.** Medium (509 lines, 7 files, docs/specs only). Gates: PR CI green; `openspec validate --strict` passes. **Verdict: 🔄 Request Changes** — the reasoning is strong, but the deltas do not connect to the specs they mean to change, and one core decision contradicts an existing requirement.

**Main points**

- 🔴 **All three `MODIFIED` requirement headers name requirements that do not exist.** `openspec archive` aborts on every one of them. `validate --strict` does not catch it, and CI's archive gate is green only because nothing is implemented — so this detonates at implementation time (task 6.5). **[F1]**
- 🔴 **`CostReceipt.notes` is forbidden from doing what Decision 3 asks of it.** An existing requirement says the receipt is an immutable open-time description and `notes` "SHALL NOT act as an occurrence log or counter". The change doesn't modify it. **[F2]**
- 🔴 **Spec coverage is short by at least four capabilities** — `backend-registry` (`required_source`, where the refusal actually lives and a public data field the change makes ambiguous), `archive-reading`'s "Bounded implicit temporary storage", and the non-seekable rows in `format-7z` / `format-iso` / `format-single-file-compressors`. **[F3]**
- 🟡 **`spool_limit = None` means the opposite of `None` in the family it claims to match** (`ExtractionLimits`/`ListingLimits`: `None` disables a guard = unlimited). **[F4]**
- 🟡 **The timing requirement is already false for multi-volume stream sources** — and the code diverges from the *current* spec there too, which the change should claim as a fix rather than inherit. **[F5]**
- 🟡 Unspecified decision gaps an implementor would have to guess: `stream_capability` after a spool, concurrency coordination, detection-vs-spool ordering on a pipe, a consumed pipe that exceeds the limit, spool-file security with a caller-named directory. **[F6–F10]**

**What's fine.** Decision 4 (config, not per-call) lands exactly on the dividing line `archive-reading` already draws. The free-space/`tmpfs` reasoning is genuinely good and is the right call — the limit as contract, the check as heuristic. The out-of-scope table for payload caching is the correct framing, and `design.md` recording the superseded two-kinds design instead of deleting it is the behaviour you want from proposals. 🎉

---

# 2. Implementor handoff

> Review of PR #251 — `openspec: bounded source spooling`, branch `claude/review-docs-status-g4hba4` @ `f7a3e02`. Scope: the OpenSpec proposal only (no code in this PR).

### F1 — 🔴 `[blocking]` / `CONFIRMED` — every `MODIFIED` header names a requirement that doesn't exist; `openspec archive` aborts

**Where:** all three delta files.

| Delta | Header used | Actual requirement in `openspec/specs/` |
|---|---|---|
| `specs/access-mode-and-cost/spec.md` | `Fail fast on non-seekable random access` | *no such requirement* — the rule lives inside `Declaring access mode at open_archive()` (`access-mode-and-cost/spec.md:19`) |
| `specs/archive-reading/spec.md` | `Configuration reaches the reader through ArchiveyConfig` | `Explicit configuration object` (`archive-reading/spec.md:717`) |
| `specs/format-rar/spec.md` | `RAR member payloads are read through RARLAB unrar` | `Use RARLAB unrar only for member data that needs it` (`format-rar/spec.md:127`) |

**Repro** (throwaway worktree at `f7a3e02`, deltas removed one at a time to reach each error):

```
$ openspec archive bounded-source-spooling -y
access-mode-and-cost MODIFIED failed for header "### Requirement: Fail fast on
non-seekable random access" - not found
Aborted. No files were changed.
```
```
archive-reading MODIFIED failed for header "### Requirement: Configuration reaches
the reader through ArchiveyConfig" - not found
format-rar MODIFIED failed for header "### Requirement: RAR member payloads are read
through RARLAB unrar" - not found
```

**Why it matters.** A `MODIFIED` block that matches nothing modifies nothing: the original requirements keep stating the old contract, and the change would land a fourth, duplicate requirement saying the opposite. Task 6.5 (`openspec archive` in the implementing PR) cannot run. `openspec validate --strict` reports "valid" — the check the PR body cites does not cover this — and CI's "OpenSpec changes archived" job is green only because nothing is implemented yet.

**Second failure behind the first.** The tool also requires a `MODIFIED` block to carry forward every scenario of the requirement it replaces. Renaming the access-mode header to the real one immediately produced:

```
current spec contains scenario(s) not present in the modified block: "open mode matrix".
Refresh the change spec before archiving to avoid dropping scenarios.
```

The same applies to `format-rar` (would drop `unrar dependency matrix`, including the `PackageNotInstalledError` / "MUST NOT silently use `unrar-free`/`unar`/`7z`" contract) and `archive-reading` (`Explicit configuration object` carries the literal frozen `ArchiveyConfig` schema — adding `spool_limit`/`spool_dir` means editing that code block, which the delta never touches).

**Fix direction:** re-target each `MODIFIED` block at the real requirement name and restate the requirement in full — existing prose plus your change, existing scenarios plus your new rows. Verify with `openspec archive <change> -y` on a scratch branch (and revert), not `validate --strict`.

---

### F2 — 🔴 `[blocking]` / `CONFIRMED` — Decision 3 puts an occurrence log in a field an existing requirement forbids from being one

**Where:** the delta's "Every spool is reported in the cost receipt" requirement vs `openspec/specs/access-mode-and-cost/spec.md:219` and `src/archivey/cost.py:85-109`.

The existing requirement **"CostReceipt remains an immutable open-time cost description"** says:

> `CostReceipt` SHALL describe static open-time properties only — not runtime diagnostics. Slow rewinds / seek-index failures go to reader/stream diagnostic aggregates (`diagnostics`). Static `notes` MAY caveat capability; **SHALL NOT act as an occurrence log or counter.**

Its own scenario matrix routes exactly this shape of fact the other way: "Backward seek re-decompresses → `STREAM_REWIND_REDECOMPRESSES` on diagnostics; `CostReceipt` unchanged."

The delta requires the opposite: `notes` "gains an entry naming the byte count" **when the spool happens**, which the timing requirement places at the first operation needing it — after open. That is an occurrence log with a counter, on a `@dataclass(frozen=True)` (`src/archivey/cost.py:85`) that `ar.cost` may already have handed to the caller. The change does not `MODIFIED` this requirement.

**The supporting argument is also weaker than the delta states.** The admission clause (`openspec/specs/diagnostics/spec.md:227`) admits a fact "the caller could not have determined from the declared contract of the call". Setting a *limit* does not tell a caller *whether* a given call spools — `review/docs-content/claims.md` E-71 makes the point precisely: a stored member from a stream costs nothing and the next, compressed, member costs a full archive copy, same call, same source. And the placement clause is about an operation that "returns a structured per-item report"; `ar.cost` is an open-time receipt, not the return value of the read that spools. The cited refusal scenario ("Proposed code restating a `CostReceipt` fact already returned at open time") only bites for facts already returned at open — which a first-read spool is not.

**Fix direction:** split static from occurrence, which satisfies both requirements without modifying either:
- `CostReceipt.notes` at open: a static caveat ("this source may require spooling up to N bytes to read compressed members") — that *is* a capability caveat, which the requirement explicitly permits.
- the spool event itself: a `DiagnosticCode`, matching how every other runtime cost fact is reported.

If you still want the occurrence in `notes`, the change must `MODIFIED` "CostReceipt remains an immutable open-time cost description" and say what happens to a receipt a caller already holds.

**Trigger / red–green candidate:** compressed RAR member read from a `BytesIO`; assert `notes` before and after the read — today both `()`; under the delta they must differ, which is the mutation the requirement forbids.

---

### F3 — 🔴 `[blocking]` / `CONFIRMED` — at least four more capability specs state the contract this change moves

The proposal's "Specs" section lists three capabilities. These also carry the rule:

1. **`openspec/specs/archive-reading/spec.md:701` — "Bounded implicit temporary storage."**
   > Reader ops SHALL NOT consume memory or temp storage proportional to member/archive size as an implicit side effect of open/read/validate/password-confirm. Silently spooling plaintext to a temp file is forbidden. A per-format strategy that inherently needs proportional temp storage … is allowed only when declared in that format's capability spec.

   Spooling a pipe at open is precisely "proportional temp storage as an implicit side effect of open". Needs a `MODIFIED` block.
   **This also strengthens your Why.** `format-rar` declares `unrar x`-to-tempdir (`:191`) and multi-volume stream materialization (`:354`) — it does **not** declare copying a whole single-volume source archive. So `_ensure_archive_path` is an *undeclared* strategy this requirement forbids today, not merely an unreported one. P11 is a contract violation, which is a stronger claim than the proposal currently makes.

2. **`openspec/specs/access-mode-and-cost/spec.md:19` — "Declaring access mode at open_archive()."** Carries the rule in prose *and* two matrix rows:
   > with `streaming=False` the library fails fast at open when the format needs seek (it does not buffer the source into memory or a temp file)
   > … never copies the archive into memory or a temp file.

   | `streaming=False` on non-seekable source that needs seek | Error at open … library does not buffer |
   | Seekable stream source, either mode | … never materialized to memory or disk |

   The second row is already false for RAR-from-stream (that *is* E-71) and this change falsifies both further.

3. **`openspec/specs/backend-registry/spec.md:236` — `required_source`.** This is where the refusal is actually enforced, and the proposal never mentions it:
   > `required_source` SHALL report the weakest source shape the format can be read from, so that the split between formats readable from a pipe and formats that must seek is **queryable as data** rather than discovered by catching `StreamNotSeekableError`. It SHALL be derived from the backend's `SUPPORTS_STREAMING_NON_SEEKABLE` declaration — the same fact `open_archive()` enforces.

   Once a permitting limit makes ZIP openable from a pipe, `required_source=SEEKABLE` for ZIP is ambiguous: format-intrinsic, or reachable-under-configuration? That is VISION claim 1 (honest capability signals as data) and addendum §2 no-surprises — a caller branching on `required_source` to decide "can I hand this a socket?" now gets an answer that depends on config the field doesn't see. **Decide and specify:** either `required_source` stays format-intrinsic and gains a documented "under the default spool limit, SEEKABLE formats may still accept a pipe" caveat, or it becomes config-aware. Do not leave it to the implementor.

4. **Per-format non-seekable rows:** `format-7z/spec.md:43`, `format-iso/spec.md:45` ("Seekability error at open; no implicit buffering"), `format-rar/spec.md:43`, and `format-single-file-compressors/spec.md:66-71` ("Random access over a non-seekable source SHALL fail fast at open with `StreamNotSeekableError`; the backend MUST NOT buffer an unbounded source"). The last one raises a question the proposal never asks: does a spool make `streaming=False` random access work over a piped `.xz`? The capability claim ("seek-requiring formats become openable from a pipe") says yes; nothing in the deltas says so.

**Fix direction:** extend the change's Specs list to `backend-registry`, `format-7z`, `format-iso`, `format-single-file-compressors`, and add the `archive-reading` "Bounded implicit temporary storage" delta. Cheapest structure: put the whole contract in `access-mode-and-cost` and reduce each format spec's row to "see the source spool limit", so there is one place to change next time.

---

### F4 — 🟡 `[important]` / `CONFIRMED` — `None` inverts the meaning it has in the family the design says it matches

**Where:** `proposal.md` (three settings), `design.md` Decision 2, vs `src/archivey/config.py:97-133`.

`ExtractionLimits`: "`None` on a guard field **disables that guard**." `ListingLimits`: same. So on `ArchiveyConfig` today, `None` = *unlimited*. The proposal puts `spool_limit = None` = *never spool* — the most restrictive setting — on the same object, two fields away, and Decision 2 justifies it as "`None` already means what a boolean's `False` would". Against the family it cites, `None` already means what `True` would.

This is exactly the "no surprises" rule (`openspec/project.md` §Design authority, addendum §2): a caller who has internalized `listing_limits.max_members=None` will read `spool_limit=None` as "no cap".

Secondary: `SpoolLimit.UNLIMITED` in the proposal reads as an enum member, while the family's sentinel is a `ClassVar` frozen instance (`ExtractionLimits.UNLIMITED = ExtractionLimits(...)`). Q2's packaging decision should settle that too.

**Fix direction:** `spool_limit = 0` for "never" (a byte count of zero is the natural refusal and needs no sentinel), leaving `None` free to keep the family's "no guard" meaning; or an explicit `SpoolPolicy.NEVER`. Either way, state the relationship to `ExtractionLimits`/`ListingLimits` `None` in the requirement, since the delta advertises family membership.

---

### F5 — 🟡 `[important]` / `CONFIRMED` — the timing requirement is false for multi-volume stream sources, and the current spec is already ahead of the code there

**Where:** the delta's timing requirement ("SHALL spool at the first operation that requires it, not eagerly at open") and its matrix row "Seekable stream, native metadata parse, listing only | No spool"; the format-rar row "Stream source, listing only | No spool; RAR metadata is parsed natively" — vs `src/archivey/internal/backends/rar_reader.py:432`.

`_materialize_stream_volumes()` is called from `_open_shared_source()`, which runs at `rar_reader.py:399` — **before** `_parse_archive()`. A multi-volume stream source is therefore spooled **at open today**, and listing does pay for it. The same delta separately requires multi-volume stream sources to be spooled "as a set", so the change contradicts itself: two matrix rows say listing never spools, one says the volume set does.

Worth claiming as a fix rather than inheriting: the existing `format-rar` requirement "Support multi-volume RAR sets" (`:354`) already says "For stream sources, data reads SHALL materialize ordered volumes for `unrar` **when needed**", and the matrix row says "data reads materialize volumes … **if needed**". The code materializes at open regardless. That's a live spec↔code divergence the change is well placed to close.

**Fix direction:** decide and state which — (a) multi-volume is an explicit exception that spools at open (then the timing matrix needs the row, and `format-rar:354` needs a `MODIFIED`), or (b) task 3.2 also moves materialization to first data read, matching the requirement that already exists. Also note `_materialize_stream_volumes` copies `Path` items with `shutil.copy2` (`rar_reader.py:450`) — a mixed path/stream volume set spools path volumes too, which the "Path source, any member | No spool" row denies.

---

### F6 — 🟡 `[important]` / `CONFIRMED` — a pipe that exceeds the limit is destroyed, and no scenario says so

**Where:** the delta's spool-limit matrix ("`ResourceLimitError` during the write; partial file removed").

For a non-seekable source the size is never known in advance, so the pre-flight can never apply and the limit is only enforceable mid-write. By the time the error is raised, the caller's pipe has been partially consumed and cannot be rewound: they cannot retry with a higher limit, and cannot fall back to `streaming=True` on the same source. The spec removes the partial *file* and says nothing about the destroyed *source*.

The reverse case is also unstated: for a seekable stream source, is the caller's file position restored after the spool? Today `_ensure_archive_path` reads through `self._shared.view(0)` (`rar_reader.py:539-547`), so the caller's handle is not disturbed — a property worth writing down before it is refactored away.

**Fix direction:** add scenario rows for both, and require the mid-write `ResourceLimitError` message to say the source is consumed and the operation is not retryable.

---

### F7 — 🟡 `[important]` / `PLAUSIBLE` — `stream_capability` after a spool is unspecified, and the receipt's ordering guarantee is violated

**Where:** both new `access-mode-and-cost` requirements vs `openspec/specs/access-mode-and-cost/spec.md:151` and `:167`.

1. A pipe spooled at open leaves `cost.stream_capability` ambiguous — `FORWARD_ONLY` (the source the caller passed) or `SEEKABLE` (what the reader now has)? The axis is totally ordered as "is at least as strong as", so callers compare it against `required_source` to decide whether a read pattern is available. Both answers are defensible; the implementor must not pick.
2. The existing requirement says the receipt is "computed during open before heavy I/O". A spool at open *is* heavy I/O during open, before the receipt exists. State the ordering (spool, then compute the receipt from the post-spool source — or the reverse).

**Fix direction:** one requirement sentence and a matrix row covering `stream_capability`, `required_source` (F3) and receipt-construction order for a spooled pipe.

---

### F8 — 🟡 `[important]` / `PLAUSIBLE` — no answer for concurrent readers

**Where:** the delta's "Two operations on one reader both need the source materialized | One spool, one note; materialization happens once per reader" vs `openspec/specs/reader-concurrency/spec.md:49-51, 68, 76`.

`reader-concurrency` already owns "first-touch materialization": under `CONCURRENT` it is coordinated (wait/share), overlapping first-touch materialization raises `ArchiveyUsageError`, `close()` drains, and the text explicitly contemplates "a synchronized, bounded/spooled shared decode/materialization strategy". The delta's "once per reader" says nothing about whether the source spool joins that coordination or is a second, independent mechanism — and it reuses the word *materialization* for a different operation than `reader-concurrency` uses it for, which will confuse both readers and greps.

**Fix direction:** state that the source spool is coordinated under the existing first-touch rules (or that it is explicitly outside them, and why), and pick a term — "spool" throughout for the source copy, "materialization" reserved for `reader-concurrency`'s meaning.

---

### F9 — 🟡 `[important]` / `PLAUSIBLE` — detection-vs-spool ordering on a pipe is a decision gap

**Where:** the timing matrix row "Non-seekable source, forward-only format | No spool; the format streams" vs `openspec/specs/format-detection/spec.md:266-287`.

To honour that row, the reader must know the format *before* deciding to spool. Detection over a non-seekable source consumes bytes through a `PeekableStream` that the opener owns and replays. So the implementation has a specific seam: detect through `PeekableStream` → consult `SUPPORTS_STREAMING_NON_SEEKABLE` (`backend-registry`) → spool *including the already-peeked bytes* → open the backend on the spooled path. Nothing in the proposal names it, and getting the peeked-byte replay wrong silently truncates the spooled archive by a few dozen bytes — a corruption bug, not an error.

**Fix direction:** name the seam in `design.md` (module layout is what that file is for) and add a task under §4, plus a test that the spooled file byte-equals the source for a pipe where detection peeked.

---

### F10 — 🟡 `[important]` / mechanism `CONFIRMED`, exploitability `PLAUSIBLE` — the deltas carry no spool-file security requirement, while adding a caller-named directory

**Where:** all three deltas (no requirement on file mode / `O_EXCL` / symlinks / name shape); `tasks.md` 6.2 defers it to a threat-model pass; `src/archivey/internal/backends/rar_unrar.py:139`.

The spec is the artifact under review, and it currently promises a caller-nameable spool directory with no constraints on how files are created in it. Three concrete gaps:

- **`unrar` argv.** `rar_unrar.py:132-139` builds `[unrar, "p", "-inul", …]` and appends the archive path **positionally and unprotected**: `cmd.append(str(archive_path))`. Member names get `-n./` protection precisely because "a hostile member name cannot inject an `unrar` switch or `@listfile` argument" — the archive path gets none. That is safe today only because every generated path is an absolute `mkstemp`/`mkdtemp` path. A caller-supplied *relative* `spool_dir` such as `-x` or `@list` produces an argv token `unrar` reads as a switch or a listfile. `format-rar` has a whole requirement named "Constrain unrar argv by call site" (`:145`) — this belongs in it.
- **Naming.** `_materialize_stream_volumes` derives volume names from `Path(self._archive_name).stem` (`rar_reader.py:442-448`); that is only safe while the parent directory is absolute and library-generated.
- **Creation mode.** `mkstemp` gives 0600 in a sticky `/tmp`; a caller-named shared directory does not, and nothing requires `O_EXCL`/no-follow. `docs/extracting.md:80-82` already sets a precedent for the leftover-after-hard-kill story (`.archivey-tmp-*`); the spool needs its own.

**Fix direction:** add one requirement — spool paths are resolved to absolute before reaching any subprocess argv; files are created with `O_EXCL`, mode 0600, and never through a symlink — plus the matching matrix rows. Keep task 6.2, but don't let the spec ship silent on it.

---

### F11 — 🟡 `[important]` / `CONFIRMED` — the docs claim this breaks is A-6, not only E-71

**Where:** `proposal.md` Impact §Docs and `tasks.md` 6.3, vs `review/docs-content/claims.md:294`.

The Impact section names E-71 only. A-6 is the bigger one, and Topic 8 has already flagged it:

> A-6 | Without `streaming=True` a non-seekable source **fails at open** … and archivey never silently buffers it to memory or a temp file. **Reword** … | `opening-and-listing.md:25-28`, `access-and-cost.md:141-143`, `philosophy.md:42`, `migrating.md:170-172` | **`wrong` as written**

Four published pages, and this change makes the absolute reading *more* wrong: under any non-zero default, a pipe handed to ZIP now is buffered. A-6's reword also can't be finalized until Q1 is answered, so the dependency runs in the direction the proposal claims it doesn't.

**Fix direction:** add A-6 to the Impact list and to task 6.3, and say in the Impact-on-Topic-8 section that A-6's reword — not just E-71's row — waits on the default.

---

### F12 — 🟡 `[important]` / `CONFIRMED` — ADR 0010 is left stating a rule the change moves, and the change's own decision gets no ADR

**Where:** `dev-docs/decisions/0010-no-silent-buffer-nonseekable.md`; `tasks.md` §6 (updates `open-issues.md`, threat model, `claims.md`, `IDEAS.md` — never `dev-docs/decisions/`).

The "extends, not reverses" argument is defensible and I'd accept it — but 0010's Decision text still reads "With `streaming=False` … **error at open**. Do not implicitly buffer." A future reader hitting a default-on spool and then reading 0010 gets a straight contradiction, and `docs/access-and-cost.md:142` (A-6, F11) is downstream of it. Separately, "one limit, not two kinds" is exactly the class of decision 0002/0010/0012 exist to record — it will be re-litigated in a year otherwise.

**Fix direction:** task under §6 to amend 0010 with a "refined by" note naming the spool limit, and either a new ADR for the one-limit decision or an explicit note in `design.md` saying why this one stays change-local.

---

### F13 — 🟢 `[nit]` / 💡 — Q3 is answerable from ADR 0012 without a maintainer round-trip

**Where:** `design.md` §Open questions Q3.

ADR 0012 draws the line at caller *bugs*: "undeclared concurrent open, operating on a closed reader … Archive/mode/feature limitations stay `ArchiveyError` (`UnsupportedOperationError`, etc.)". Setting `spool_limit` to never and then opening a RAR from a stream is not a bug — it is a configured refusal of a capability, which is what `UnsupportedOperationError` names. The clinching argument is adjacency: exceeding a numeric limit already raises `ResourceLimitError`, an `ArchiveyError`. Choosing `ArchiveyUsageError` for the never-setting would put two refusals from the same feature on opposite sides of the `ArchiveyError` boundary, so a caller's `except ArchiveyError` would catch one and not the other.

**Recommendation:** `UnsupportedOperationError`, with the reasoning written into `design.md` rather than left open.

---

### F14 — 🟢 `[nit]` / `CONFIRMED` — the tool's own warning on `proposal.md`

`openspec archive` prints: `⚠ Why section should not exceed 1000 characters`. Non-blocking and the tool continues, but it's free to fix while F1 is being addressed — the measured-output block and the ADR 0010 quote can move to `design.md`, which is where that depth belongs.

---

### F15 — 💡 `[suggestion]` — one number can't express the one preference that actually is asymmetric

`design.md` Decision 1 kills the two-kinds split with the asymmetric-preference test, and calls the reverse preference "incoherent". It isn't, and Q1 is where that surfaces: "spool because `unrar`'s API needs a path" overrides no statement the caller made (they asked for RAR data; the tool needs a file), while "spool because you handed me a pipe and said `streaming=True`" overrides a stated intent — which is the case ADR 0010 was written about. Decision 6 keeping *timing* as a caller-visible difference is evidence the two operations are not interchangeable outside the library.

Not reopening the collapse — one config knob is the right call. The point is narrower: **a single default number silently decides both**, and Q1 currently frames it only as "which RAR-from-stream reads start failing". See block 3, item 2.

---

**Verdict: 🔄 Request Changes.** F1 blocks archiving the change at all; F2 and F3 mean the deltas would leave contradictory requirements in `openspec/specs/`. F4–F12 are addressable in the same rewrite. The design reasoning itself is sound and mostly survives.

---

# 3. Maintainer decisions

1. **Where does a spool occurrence get reported?** (F2)
   *Why you:* the change answers P11's open question in a way an existing requirement forbids, and P11 itself never noticed that requirement — so this is a real contract decision, not an editing slip.
   **A.** Static caveat in `CostReceipt.notes` at open + a `DiagnosticCode` for the occurrence — satisfies both existing requirements, no delta needed against the immutability rule. *(recommended)*
   **B.** Occurrence in `notes` as proposed — requires `MODIFIED`-ing "CostReceipt remains an immutable open-time cost description", deciding what a caller holding an earlier `ar.cost` sees, and accepting that the receipt is no longer open-time-only.

2. **The default limit — it decides two things, not one.** (F15, and Q1)
   *Why you:* Q1 frames the default as "which RAR-from-stream reads break". Any non-zero default *also* silently converts today's `StreamNotSeekableError` on a pipe into a whole-archive disk copy for ZIP/7z/ISO — including under `streaming=True`, where the caller has explicitly said "stream this". That is ADR 0010's core case.
   **A.** One number, e.g. 1 GiB, applied to both — simplest surface; a pipe + `streaming=True` + ZIP silently spools up to 1 GiB.
   **B.** One number, plus a stated rule that `streaming=True` never spools for seekability (the spool for an external binary is unaffected) — keeps one knob, keeps ADR 0010's promise where the caller stated an intent, and answers Q4 as a side effect. *(recommended)*
   **C.** `UNLIMITED` default — nothing breaks, reporting becomes the whole fix, and P11's "unbounded" half survives.
   **D.** Never-spool default — safest, removes a capability that works today, makes RAR-from-stream a breaking change for everyone.

3. **Is `required_source` format-intrinsic or configuration-aware?** (F3, item 3)
   *Why you:* `backend-registry` sells it as *data*, the honest alternative to catching `StreamNotSeekableError`. Once a pipe can open a ZIP under a permitting limit, the field either becomes a caveated half-truth (A) or grows a config dependency it was designed not to have (B). VISION claim 1 territory, and the proposal never raises it.
   **A.** Stays intrinsic; documented caveat that a SEEKABLE format may still accept a pipe under the spool limit. *(recommended — keeps the field derivable from one backend declaration, as the requirement demands)*
   **B.** Becomes config-aware; `format_availability()` grows a config parameter.

4. **Multi-volume stream sources: exception, or fix?** (F5)
   *Why you:* code and current spec already disagree (`format-rar:354` says materialize at data reads; `rar_reader.py:432` does it at open), so whichever way you go, something changes. The proposal's timing requirement assumes the fix without saying so.
   **A.** Fix it — defer volume materialization to first data read; timing rule stays universal. *(recommended; also makes listing a multi-volume stream set free, matching the single-volume path)*
   **B.** Declare multi-volume an exception that spools at open, and add the matrix rows saying so.

5. **Does the change also claim P11 as a contract violation?** (F3, item 1)
   *Why you:* `archive-reading`'s "Bounded implicit temporary storage" allows a proportional-temp-storage strategy "only when declared in that format's capability spec", and `format-rar` declares `unrar x`-to-tempdir and multi-volume materialization — not a whole-archive copy of a single-volume source. If you agree that's undeclared, P11 upgrades from "unreported cost" to "forbidden by spec", which changes both the Why and the urgency of scheduling.

---

davitf:
Came here from reviewing #313 / #314, which land on this proposal's ground. Short answer to "does this still make sense": **the problem statement holds unchanged, but the design predates vocabulary the repo has since adopted, and two of its premises are now known to be wrong.** Worth a revision pass before anyone implements it — not a rewrite.

Checked against `origin/main` @ `7ed4879`; this PR's base is `bba02f0`, 46 commits back.

## Still true

The `Why` section is intact. Nothing merged since has touched reader-side spooling — `grep -rn -i spool src/archivey/ openspec/specs/ docs/` finds only detection code. A RAR from a stream still writes the whole archive to a temp file with no limit and nothing in either honesty channel: no backend on `main` populates `CostReceipt.notes` at all (`grep -rn "notes=" src/archivey/` → nothing). Open questions 1–4 are all still open, and #1 (the default limit) is still the load-bearing one.

## Premise 1 needs correcting: the timing rule is not what the PR says

> a spool for an external binary happens at the first read that needs it, so an archive can be **listed** without paying for it; a spool to make a source seekable happens at open

That is true for a single stream source and **false for stream volumes**. `_materialize_stream_volumes()` runs from `RarReader.__init__` (via `_open_shared_source`), so handing over a list of open volume streams copies every volume at `open()`, before any member is named. Metadata parses from the originals, so a list-only caller pays for bytes nothing reads. Measured on the two-volume `tinyvol` fixture: **1757 of 1757 bytes written to `/tmp` at open, with no `read()`.**

So the "tool-tax vs capability" split this PR collapsed does not line up with the timing distinction it kept as the one survivor — an external-binary spool can happen at either moment depending on the source shape. Either the requirement's scenario matrix grows the volume case, or the fix goes in first and the rule becomes true. I've recorded the defect and a fix direction as handbook §10 #21 in #316.

## Premise 2 needs correcting: `CostReceipt.notes` collides with a requirement

> **`CostReceipt.notes`, not a diagnostic.** … the placement clause prefers a structured field, and `notes` is it.

`openspec/specs/access-mode-and-cost/spec.md` has *"Requirement: CostReceipt remains an immutable open-time cost description"*:

> `CostReceipt` SHALL describe static open-time properties only — not runtime diagnostics… Static `notes` MAY caveat capability; **SHALL NOT act as an occurrence log or counter.**

A note written *when the spool happens* is an occurrence log, so the answer as phrased conflicts with a spec this PR already has deltas against. #314 hit the same wall implementing it. The resolution taken there: **emit the caveat at open, derived from the source shape**, not after the fact. That satisfies both this requirement and P11's own bar (*"deserves to know before the first read, not after"*), and it needs no revision to the immutability requirement. Worth folding into this proposal rather than leaving #314 to settle it code-first.

## The duplication that is the real reason to revisit

`src/archivey/detection_cost.py` **did not exist** when this was written — it landed in #273 (`detection-prefix-workspace: one-pass prefix buffer, budget, and receipt`). It already gives the repo a spool vocabulary:

- `DetectionBudget.spool_non_seekable_up_to: int` — the limit
- `DetectionCostReceipt.spooled_bytes: int` — what was actually spooled
- `within_budget()` comparing the two
- `SpooledTemporaryFile` machinery in `internal/detection_workspace.py`

This PR's `spool_limit` (bytes / `UNLIMITED` / `None`) is a second concept for the same idea, with a different name, a different "never" sentinel (`None` here, `0` there), and a different reporting channel. Landing both would put two spool budgets in one library — the duplicated-concept case Topic 9 was a whole review theme about. Open question 2 (packaging: flat fields vs a frozen `SpoolLimits`) should probably be reframed as *"reuse or generalise the detection budget/receipt pair"* before it is answered on its own terms.

## Proposal: a reporting field to go with the limit

Related, from the #314 review. Rather than prose in `notes`, declare the disk a source may cost as a number:

**`CostReceipt.max_temp_disk_bytes: int | None`** — upper bound on bytes archivey may write to serve this archive. `0` for a path source, the compressed size for a sized stream, `None` when the source size is not cheaply knowable (`compressed_source_size -> int | None` already sets that convention). `0` and `None` must stay distinct: *"will not write"* vs *"will write, cannot say how much"*.

Why it belongs in this PR rather than a RAR one:

- A field that **declares** what we would write and a field that **caps** it are two halves of one contract. Designed apart, they end up disagreeing — and the caller's real question is whether the two are compatible.
- **Per archive, not per member.** `CostReceipt` is archive-level, so a per-member figure collapses to a max anyway; a peak is what answers "does this fit in my container's `/tmp`", where a unit cost needs the access pattern first; and it is static and open-time knowable, which the immutability requirement demands.
- **It survives handbook §10 #6 layer 2 without a contract change.** Today the peak is the compressed archive size. After the synthetic single-member archive, non-solid becomes the largest packed member and solid stays roughly archive-sized. The field's meaning is identical; only the value shrinks. Layer 2 becomes a pure improvement rather than a spec revision — worth a lot if the contract gets written now.
- If it ships, **drop the note**. A non-zero value is the signal, it is machine-readable, and it sidesteps the occurrence-log problem entirely.
- One thing to specify rather than discover: under `concurrent_members=True` with per-member temps, the peak is a **sum** of live temps, not a max.

## Suggested order

#314 is currently landing a fragment of this design code-first, without referencing this PR. Either it waits for the shape to be settled here, or it ships the minimal open-time caveat and this PR absorbs it — but the contract should be written in one place, and this is the place.

---

PR #252, opened 2026-08-18
Topic 8: capability verification fan-out (pass 1 verdicts)

davitf:
Topic 8 pass 1: capability verification fan-out, merge, O-26 classification. **No guide prose.**

## Gate

- `scripts/setup-dev-env.sh`: `ok unrar`, `ok 7z`
- `list_known_formats()` → **26×FULL** (F4-corrected; not the 22 named-attr walk)
- `./scripts/check.sh --fix` — pass
- `./scripts/test.sh` — **2474 passed**, 23 skipped

## Verdicts (`claims.md`)

| Verdict | Count |
|---|---:|
| verified | 382 |
| wrong | 11 |
| out of scope ([TM]) | 8 |
| unverifiable (platform) | 1 (C-17) |

**Wrong IDs:** A-6, A-16, A-18, A-33, A-34, D-48, E-1, E-71, F-16, F-26, G-9.

Workers ran **one at a time** (A→I); coordinator spot-checked early clusters before continuing. Evidence: `review/docs-content/worker-inputs/verdicts-*.md`. Harvest: `review/problem-catalogue/harvest/*.md`.

## Cross-cluster consistency

No contradictory verdict pairs on the same behavioural fact (seek lists, buffering, rapidgzip, accelerator contain/abort, exception-name pairs). Details in `claims.md` §Pass 1 verification summary.

## O-26 / QUESTIONS

See [`review/docs-content/QUESTIONS.md`](review/docs-content/QUESTIONS.md):

- **Prose** (page PRs): A-6, A-16, A-18, A-33, D-48, E-1, E-71 docs half, F-16, F-26, G-9
- **Code** (#225 shape): **A-34** SFX detect — includes silent `BROTLI` misdetect; forced-format escape is RAR-only (Q1 corrected)
- **Open questions:** Q2 sticky under STANDARD (C-25), Q3 foreign-member `ValueError` vs `ArchiveyUsageError` (B-26), Q4/P11 RAR spill signal, Q5/scope Q3 api.md exceptions; **MD1** A-34 priority still open

## Review round-1

F1–F4 fixed; F5 deferred to gotchas page PR; MD4 deferred to `review/backlog.md` Topic 8 follow-ups. See disposition table in `QUESTIONS.md`.

## Concentration of wrongs

Heaviest on **`access-and-cost.md`** (A-6, A-16, F-16, F-26 + E-71 home TBD), then opening samples, formats matrix/SFX, unshipped `VerificationMode.STRICT`, free-threaded extras wording.

[...]

PR #253, opened 2026-08-18
OpenSpec: SFX detection + 7z start-offset (from Topic 8 A-34)

davitf:
Proposals only — no implementation yet. Split from Topic 8 claim A-34 / #252 review.

## Priority

**Raised (MD1 = B, 2026-08-18).** Implement ahead of ordinary Topic 8 docs prose fixes. Silent-success regression is mandatory (not only `FormatDetectionError`).

## Why

`format-detection` already requires SFX detection with `payload_offset`, and says parsers SHALL accept a start offset. Code never finished that:

1. **`detect_format` has no SFX scan** → low-entropy `MZ` stub can be claimed by Brotli; `open_archive` succeeds with a fabricated member (silent wrong answer). Same shape for ZIP SFX.
2. **RAR vs 7z asymmetry** → forced `format=RAR` works; forced `format=SEVEN_Z` raises `CorruptionError` (magic at byte 0 only).

## Changes (both validate `--strict`)

| Change | Role |
|---|---|
| [`sfx-format-detection`](openspec/changes/sfx-format-detection/) | SFX scan for **RAR / 7z / ZIP**; own “no silent wrong answer” requirement; **investigate** Brotli vs stub differentiation; shared `SFX_MAX`; wire `payload_offset` |
| [`sevenz-sfx-start-offset`](openspec/changes/sevenz-sfx-start-offset/) | 7z start offset + forced-format scan using the same `SFX_MAX` |

Suggested land order: **7z start-offset first** (or same train), then detection.

## Review status (`3e82bf4`) — round complete

| ID | Disposition |
|---|---|
| F1 | Fixed — ZIP in (A) |
| F2 | Fixed — start-offset arg or offset view |
| F3 | Fixed — two 7z absolute seek sites named |
| F4 | Fixed — RAR stream spill × offset risk |
| MD2 | Fixed — split requirement + probe/stub investigation |
| MD3 | Fixed — one shared `SFX_MAX` (A) |

## Out of scope here

Implementation (`/opsx:apply`), guide prose.

Trailing archive tasks left **unchecked** (changes not finished).

<div><a href="https://cursor.com/agents/bc-953aa7e9-f4b9-45f0-b325-9965a916f759?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-953aa7e9-f4b9-45f0-b325-9965a916f759&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


openspec/changes/sfx-format-detection/design.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Goals:**
+- Implement SFX scan in `detect_format` for RAR and 7z magic within a bounded window.
+- Prevent content probes from silently winning on executable stubs.
+- Hand `payload_offset` to the open path so backends read in place from the payload.
+- Tests that fail on the silent-success path (not only the raising path).
+
+**Non-Goals:**
+- Implementing 7z parser start-offset / SFX scan (sibling `sevenz-sfx-start-offset`).
+- ZIP SFX / other formats’ embedded payloads.
```
davitf: ## 🟡 F1 — ZIP SFX has the identical silent-BROTLI defect, and is excluded with no rationale

**Severity: 🟡 important · Confidence: CONFIRMED (repro)**

"ZIP SFX / other formats' embedded payloads" is listed as a Non-Goal with no reason given. Measured on `4f95c69` + `[all]`, a real `.zip` behind a 4 KiB `MZ` stub:

| Stub filler | `detect_format` | `open_archive` (auto) | forced `format=ZIP` |
|---|---|---|---|
| `MZ` + `0x90`×4094 | **`BROTLI` / PROBABLE** | **OK** — bogus member `sfx_zip_lowent.exe.uncompressed` | OK — real member `f.txt` |
| `MZ` + varied bytes | `FormatDetectionError` | `FormatDetectionError` | OK — real member `f.txt` |

That is **the A-34 shape exactly**, including the fabricated `*.uncompressed` member. So the Non-Goal cannot rest on "ZIP is unaffected". It also cannot rest on "ZIP SFX is rare" — self-extracting ZIP `.exe` files are the most common SFX form in the wild, and forced `format=ZIP` already reads them today, because `zipfile` locates the EOCD from the tail and the stub is simply ignored.

**The proposal is ambiguous about what happens to ZIP, in a way an implementor must guess at.** Two clauses in the spec delta pull in different directions:

- The format-agnostic sentence — *"When the prefix looks executable, a content probe SHALL NOT claim a stream codec (e.g. Brotli) and allow a successful open of a fabricated member"* — **would** cover ZIP, and is the only thing protecting it.
- But the scan itself is RAR/7z-only, and `No match → fall through (extension / FormatDetectionError)`.

Net effect under this spec as written: a low-entropy ZIP SFX stops being a silent wrong answer (good) and becomes `FormatDetectionError` (worse than necessary) for a file the library can already open. Whether that is intended is not decidable from the text.

**Fix — pick one and say so:**
1. Add `PK\x03\x04` to the SFX scan needles, and a matrix row for it. Cheap: the scan machinery is identical.
2. Keep ZIP out of scope, but state the reason and add a matrix row fixing the expected outcome (`FormatDetectionError`, explicitly not a silent Brotli open), so the fall-through is a decision rather than an accident.

**Repro:** prepend `b'MZ' + b'\x90'*4094` to any `zip`-produced archive; `detect_format` → `BROTLI`, `open_archive` succeeds with one fabricated member.

---

openspec/changes/sfx-format-detection/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,20 @@
+## ADDED Requirements
+
+### Requirement: Honour detection payload_offset at open
+
+When `detect_format` returns `payload_offset > 0`, `open_archive` (auto-detect
+path) SHALL open the archive at that byte offset — by positioning the source or
```
davitf: ## 🟡 F2 — "positioning the source" and "passing a start offset" are offered as equivalent; they are not

**Severity: 🟡 important · Confidence: CONFIRMED**

This requirement lets the implementor satisfy `payload_offset` *"by positioning the source **or** passing an equivalent start offset into the backend"*. Those two are not interchangeable, and the 7z backend proves it:

`sevenzip_parser.py:357` — `read_signature_and_next_header` opens with an **absolute** `fp.seek(0)`, which discards any positioning the opener performed. Line `:388` then does `fp.seek(_SIGNATURE_HEADER_SIZE + next_header_offset)`, absolute again.

So an implementor who reads this clause and chooses the "position the source" branch gets:

- **RAR** working (its parser scans forward from wherever it starts), and
- **7z** still raising `CorruptionError`,

and would reasonably conclude the sibling change `sevenz-sfx-start-offset` was at fault, when in fact the hand-off mechanism was the wrong one. That is a decision gap that sends implementation down a wrong path before any test is written — exactly what §9 asks a proposal review to catch.

**Fix:** narrow the clause. Either require an explicit start-offset argument into `backend.open_read`, or require a **bounded offset view / slice** whose byte 0 *is* the payload (which does survive an absolute `seek(0)`). Add one line stating that a bare seek on a shared handle is insufficient for backends performing absolute seeks, and name 7z as the instance. `sevenz-sfx-start-offset` Decision 1 already assumes the explicit-argument shape — this delta should not leave the weaker option open.

---

openspec/changes/sevenz-sfx-start-offset/design.md
```diff
@@ -0,0 +1,66 @@
...
+`S + relative` (today’s code assumes S=0). Document the invariant in the parser
+module.
+
+### 4. Spec lives primarily in `format-7z`
+Add an explicit requirement so format owners see it next to other 7z contracts.
+`format-detection`’s cross-cutting sentence remains; this change does not remove it.
+
+## Risks / Trade-offs
+
+- [Missed seek sites after introducing S] → Audit all `seek`/`tell` in 7z parser
```
davitf: ## 🟢 F3 — the "missed seek sites" risk is measurable; put the number in the design

**Severity: 🟢 nit · Confidence: CONFIRMED**

This risk row is the right one to name, but "audit all `seek`/`tell` in 7z parser / pipeline" reads open-ended, which makes the change look larger and riskier than it measured.

At `4f95c69` there are exactly **two** absolute seeks to rebase:

```
src/archivey/internal/backends/sevenzip_parser.py:357    fp.seek(0)
src/archivey/internal/backends/sevenzip_parser.py:388    fp.seek(_SIGNATURE_HEADER_SIZE + next_header_offset)
```

`sevenzip_pipeline.py` contains no `.seek(` calls at all.

Naming both sites turns Decision 3 into a two-line rebase plus the invariant comment. **Keep the pack-stream fixture task regardless** (`3.2`) — pack offsets may be applied through stream slicing rather than a literal `.seek(`, so the count bounds the *known* sites, not the whole surface. Stating it that way is more useful to whoever implements than an unbounded audit instruction.

---

openspec/changes/sfx-format-detection/specs/archive-reading/spec.md
```diff
@@ -0,0 +1,20 @@
...
+passing an equivalent start offset into the backend — so the reader sees the
+embedded payload in place. The system SHALL NOT copy the remainder of the source
```
davitf: ## 💡 F4 — interaction with the RAR stream temp spill (E-71 / P11) is unstated

**Severity: 💡 suggestion · Confidence: PLAUSIBLE (not traced to a failure)**

*"SHALL NOT copy the remainder of the source to a temporary file solely to strip an SFX stub"* is the right constraint and correctly ADR-0010-shaped.

The unaddressed case is the one Topic 8 just documented: a **RAR SFX arriving on a seekable stream**. `RarReader._ensure_archive_path()` already copies the whole archive to a temp `.rar` because `unrar` needs a path (E-71; `open-issues.md` P11). Once detection supplies a non-zero `payload_offset`, two mechanisms compose and the proposal does not say how:

- Does the backend receive the **offset view** (temp file contains payload only, `payload_offset` already consumed), or the **original bytes** (temp file contains stub + payload, and `unrar`'s own SFX scan re-finds the magic)?

Both can work. But they differ in what P11's eventual cost signal should report, and in whether the temp file is byte-identical to something the caller could have produced. Since P11 is an open decision about exactly that spill, one sentence here keeps the two from being decided inconsistently in different PRs.

Not blocking — it does not change the shape of either change, and no failure is traced. Worth a line in `sfx-format-detection`'s Risks table.

---

davitf (commented):
# Review — OpenSpec: SFX detection + 7z start-offset

Reviewed head `2bb18ff`, base `4f95c69`. First review of this PR. Finding IDs `F1`–`F4` are scoped to this PR (they do not continue #252's numbering).

Reviewed under addendum **§9** (values-first) rather than §8, since the artifacts are OpenSpec proposals and delta specs, not a code diff.

Findings **F1–F4** are inline. Blocks 1 and 3 follow.

---

## 1. Maintainer briefing

**What this change is.** Two OpenSpec proposals, no implementation, split out of Topic 8's A-34. `sfx-format-detection` makes `detect_format` actually perform the SFX scan its spec already requires, orders that scan ahead of content probes so an executable stub cannot be silently claimed as Brotli, and wires `FormatInfo.payload_offset` through the open path. `sevenz-sfx-start-offset` teaches the 7z parser to accept a signature at a non-zero offset, removing the RAR/7z asymmetry. Both validate `--strict`; trailing archive tasks correctly left unchecked.

**Snapshot.** 13 files, +379, all under `openspec/changes/`. Small. **Verdict: 💬 Comment** — the proposals are in good shape; F1 and F2 should be resolved before implementation starts, since both are decision gaps an implementor would otherwise have to guess at.

**Main points.**

- 🟡 **F1** — **ZIP SFX has the identical silent-`BROTLI` defect** (verified: fabricated `*.uncompressed` member, forced `format=ZIP` reads it fine), yet it is a Non-Goal with no stated reason, and the spec's two clauses disagree about what should happen to it.
- 🟡 **F2** — the `archive-reading` delta offers "positioning the source" and "passing a start offset" as equivalent. They are not: `sevenzip_parser.py:357` does an absolute `fp.seek(0)`, so the positioning branch silently fails for 7z.
- 🟢 **F3** — the "missed seek sites" risk is measurable (exactly two absolute seeks); saying so shrinks the apparent blast radius.
- 💡 **F4** — the interaction with the RAR stream temp spill (E-71 / P11) is unstated.

**What's fine** — checked, not assumed:

- **The core factual claims hold.** `sevenzip_parser.py:357` really is an unconditional `fp.seek(0)` with magic required at byte 0; `format-detection:237` really does already say native RAR/7z parsers SHALL accept a start offset; `format-7z` really does not mention offsets anywhere. The "spec already requires it, code never finished it" framing is accurate, not rhetorical.
- **O-26 posture is right.** Faced with spec-requires / code-defers, these implement the spec rather than retracting it, and Q1's Option 2 is now honest about what retraction would cost.
- **Cost honesty preserved.** Both changes explicitly forbid a whole-archive temp copy and require read-in-place, which keeps ADR 0010 and the cost-receipt story intact rather than trading correctness for convenience.
- **Scenarios are falsifiable.** Both matrices read as conformance assertions someone could write directly — including the negative cases (`not BROTLI`, "not a silent empty archive"), which is the half usually missing.
- **Rejected alternatives are recorded** in both `design.md` files, and the mandatory silent-success regression survives into `tasks.md` rather than living only in prose.

---

## 3. Maintainer decisions

1. **Is ZIP SFX in or out of this train? (F1)** Measured: a ZIP behind a low-entropy `MZ` stub detects as `BROTLI` and auto-opens with a fabricated member — the same defect A-34 describes — while forced `format=ZIP` reads the real member today, because `zipfile` finds the EOCD from the tail. *Why you:* self-extracting ZIP is the most common SFX form in the wild, so excluding it leaves the majority of real SFX files on the broken path, but including it widens a change you have just raised in priority. **Options:** (A) add `PK\x03\x04` to the scan needles now — the machinery is identical, cost is a needle plus a matrix row; (B) keep it out, but state the reason and pin the expected ZIP outcome in the matrix so the fall-through is deliberate. **Recommendation: (A)** — the incremental cost is close to zero and (B) leaves the most common case at `FormatDetectionError` for a file the library can already read.

2. **Should the "no silent wrong answer on an executable prefix" clause be its own requirement?** It currently sits inside the SFX-RAR/7z requirement, but it is broader than SFX — it governs content probes versus executable prefixes generally, and under Option (B) above it becomes the *only* thing protecting ZIP SFX. *Why you:* this is a spec-organisation call with a real consequence — buried inside an SFX requirement, it is easy to lose or narrow in a later edit, exactly the drift `format-7z` suffered when the start-offset obligation lived only in `format-detection`. **Options:** (A) split it into its own requirement in `format-detection`; (B) leave it nested. **Recommendation: (A)**, and it costs nothing now.

3. **One shared SFX window constant, or three?** `sfx-format-detection` Decision 2 says "defaults to RAR's 2 MiB unless implementation finds a reason to share one constant"; `sevenz-sfx-start-offset` says its bound "SHALL be at least as large as the RAR parser's SFX window". That is three bounds described relative to each other rather than one constant. *Why you:* this repo already paid for that pattern — O-2 was four copies of one caveat that drifted into two wrong ones, and the fix was a single source. **Options:** (A) promote `SFX_MAX` to one shared constant as part of this train; (B) leave three and rely on the "at least as large as" wording. **Recommendation: (A)** — cheaper now than after three call sites exist.

---

## What I actually ran

- `openspec validate --strict` on **both** changes in a clean worktree at `2bb18ff` — both report valid (confirming the PR body's claim rather than copying it).
- Built ZIP SFX fixtures (low-entropy and varied stubs) and ran `detect_format` / `open_archive` / forced `format=ZIP` — the F1 table.
- Read `sevenzip_parser.py` and grepped every `.seek(` in the 7z parser and pipeline — the F2 and F3 evidence.
- Confirmed `format-detection:237` carries the start-offset sentence and that `format-7z` carries none.

**Not** run: the test suite (no code changes here), and no implementation was attempted or reviewed — these are proposals.

---

PR #257, opened 2026-08-21
OpenSpec: prefixed-archive-detection — tier the search by cost, not by stub shape

davitf:
Split out of #255, which was chartered to investigate the Brotli content probe and grew this as a second, larger proposal. **Proposal only — no production code.** Rebased onto `main` after #262 merged; `openspec validate --strict --all` 29/29, `./scripts/check.sh` clean.

## The gap, measured on `main`

| file | `detect_format` | `open_archive(format=ZIP)` | stdlib `zipfile` |
|---|---|---|---|
| `zipapp` output (`#!/usr/bin/env python3` + ZIP) | `FormatDetectionError` | **opens, all members** | opens |
| Spring Boot executable JAR (`#!/bin/sh` + ZIP) | `FormatDetectionError` | **opens** | opens |
| JPEG with an appended ZIP | `FormatDetectionError` | **opens** | opens |
| makeself `.run` (`#!/bin/sh` + tar.gz) | `FormatDetectionError` | fails — genuinely needs an offset | n/a |

The ZIP rows are the awkward ones: **the backend already reads these files**, `zipapp` is in the Python standard library, and only detection refuses — so archivey is currently worse than `zipfile` on a `.pyz`. The executable cue never fires because the prefix is `#!`, not `MZ`.

## The design follows from getting one rationale straight

The existing cue is a **cost gate** — avoid reading `SFX_MAX` (2 MiB) from every source — **not a false-positive defence**. A reviewer of #255 read it the other way, and so did I until corrected. Once that's the stated reason, the tiers fall out:

| tier | bound | bounded by |
|---|---|---|
| tail probe (ZIP) | 65535 + 22 bytes for the *locator* | **the format** — EOCD comment length is a `uint16` |
| forward scan (7z/RAR/tar) | `SFX_MAX`, 1.66× on a miss | a constant we chose |
| exhaustive scan | whole source | nothing |

A bound the format hands you can run unconditionally; a bound you invented needs a reason to run.

**Widening the cue is a population cost, not a free one.** On 72,100 files under `/usr`, it enrolls 734 shebang files and 8 Mach-O against 2,868 ELF/`MZ` already paying — 26% more files entering the scan. The per-file bound is `min(size, SFX_MAX)`, and since the new population is mostly small scripts (median 2,959 B; exactly one file reaching the window) the whole tree costs 10.3 MiB more. Both halves are in `design.md`, because "every `#!` pays 2 MiB" is as wrong as "free".

**The tail probe's bound covers locating the EOCD, not validating it.** Computing an exact `payload_offset` can require walking the whole central directory, which 65,557 bytes does not bound. Measured: 4.4% of files are large enough to pay the locator in full, median file 2,239 B, **actual aggregate 0.61 GiB** against a 4.39 GiB worst case. What is *not* measured — and likely dominates a backup sweep — is one extra seek per file, so enabling this tier by default is gated on measuring that (task 1.0) rather than inherited from "format-bounded".

## Scan hits are validated, not trusted

7z's 32-byte signature header carries a CRC32 over its own StartHeader *and* declares where the archive ends — verified against stubs of 23 B, 4 KiB and 100 KB. RAR5's 8-byte marker is followed by a CRC-checked main header. `<=` as the gate, `==` as the tiebreak.

The **ZIP tail probe** is validated to the same density: the record must fit, `comment_length` must exactly consume the tail, the derived central-directory position must be in range and start with `PK\x01\x02`, and the earliest adjusted local-header offset must land on `PK\x03\x04`. ZIP64 is followed via its locator rather than rejected, and a failed candidate resumes the backward search.

This supersedes #254's explicit deferral, which took the earliest matching needle and accepted that a decoy sends the backend to a wrong offset.

## ZIP has two prefix shapes, and they store different numbers

| | prefix | first local header | stored `offset_cd` | actual CD | EOCD adjustment |
|---|---|---|---|---|---|
| `zipapp` `.pyz` (written in place) | 23 | 23 | 76 | 76 | **0** |
| concatenated (`cat stub payload.zip`) | 33 | 33 | 53 | 86 | **33** |

`zipapp` writes offsets from byte 0 of the file, shebang included; a concatenated ZIP writes them from the payload. **`payload_offset` is the absolute position of the earliest local file header** — the prefix length under both, and what the forward scan already reports for an `MZ`-prefixed ZIP. The EOCD-adjustment alternative reports a `zipapp` as `0`, making the headline case indistinguishable from an unprefixed archive.

Both shapes open either way today, so this is a definitional fix rather than a latent breakage — but it means a test asserting only "it opens" pins nothing.

## Compressor needles are scoped to the `#!` cue

A makeself/NVIDIA/Anaconda `.run` payload starts with raw gzip magic; `ustar` doesn't appear until after decompression. Maintainer decision: **stream-codec needles are searched under a `#!` cue only** — that stub shape is real for scripts and not for executables.

Needle length is what makes it affordable. Across 497 ELF binaries (176 MB), gzip's 2-byte `1f 8b` collides **423** times; the 3-byte `1f 8b 08` collides **3**. The method byte is required, a bounded decode is the actual evidence, and the inner-TAR probe resolves a hit to `TAR_GZ`. This narrows rather than reverses the `registry.sfx_magic_entries()` policy, whose stated reason is false for shebang stubs and true for executable ones.

## Also closes the Mach-O gap #254 deferred

A thin 64-bit Mach-O stub matches no cue, and `cf fa ed fe` is *structurally* a valid Brotli uncompressed meta-block header. Task **2.1a** carries the re-measured premise: a realistic-entropy stub raises `FormatDetectionError`, while a low-entropy one is claimed by **LZMA Alone** at `PROBABLE` with a fabricated member and no unconfirmed signal. Both become `SEVEN_Z` with real members.

`ca fe ba be` forces an asymmetry: it is also the Java class-file magic, and grading it *weak* would not spare a `.class` file, because a weak cue still triggers the scan. So Mach-O raises **no cue** until its header parses — nothing is lost, since a real stub parses by construction.

## Sequencing — resolved

- `sfx-format-detection` is archived (#258); all three of #254's additions verified present.
- **#262 merged, so the archive order is settled and this delta has been rebuilt on the resulting live text.** Both changes MODIFY *Executable-looking prefixes*; they edited disjoint parts, and task 0.3 lists the five inherited items. Verified by diffing the rebuilt block against live — the only live lines it no longer carries are the four this change deliberately rewrites.
- **The shared source-length helper already exists.** #261 reused `source_byte_size()`; the ZIP tail probe should build on the same dispatch, and its `None` return is the tier-2 skip condition.

## Known-wrong and deliberately not fixed here

#263's independent analysis was accepted as redesign input. It endorses this acquisition order — including far magic ahead of the content probes — and identifies two things this change keeps, both now **marked provisional in the spec** rather than quietly retained:

- **A uniform `CERTAIN` for all near magic is unsound.** gzip's two bytes are not XZ's flags CRC, 7z's `StartHeaderCRC`, or TAR's header checksum.
- **First-match-wins is not a sound selection rule** where several detectors can accept the same bytes.

It also found an implementability gap: needles carry an anchor offset **inside their own format**, so a TAR `ustar` hit at `H` means a candidate origin of `H − 257`, and `peek_more` cannot express a candidate-relative view at all. Tasks 2.5a/2.5b make a bounded `peek_range` a prerequisite, with an off-by-257 red–green.

## Public API surface

`FormatInfo` gains `prefix_kind` (always present, default `NONE`) and a `PrefixKind` enum. The exhaustive-scan opt-in is **`ArchiveyConfig.exhaustive_prefix_scan`, not a keyword argument** — `detect_format(source, *, config)` takes no per-call operational keywords, so a kwarg on `open_archive` could not be expressed there. `strict_archive_eof` is the precedent.

## Open question, deliberately not settled

Whether there are prefixed 7z/RAR files in the wild that are *not* self-extracting executables. Everything found so far says no; the known exception is script-wrapped payloads, which is why the cue gains `#!`. Old DOS/Windows-era installers and media images are the blind spot (task 5.2). That open question is also why there is no ADR yet — per `CONTRIBUTING.md`, a decision still needing one is not an ADR. Task 5.1 tracks it.

https://claude.ai/code/session_01AVTvgthVWwLQYaZHCFuxTt


davitf:
Folded #263's analysis into this change in `63e82f3`. It was accepted as redesign input, and it endorses the acquisition order here — including far magic ahead of the content probes, calling the bootable-ISO reproduction decisive. Four things needed acting on now.

## The delta contained two contradictory orders — mine

The canonical algorithm puts far magic at step 3, but the SFX requirement's fall-through sentence still read "content probes, far magic, extension". That was correct when far magic sat at step 7 and I failed to update it when I moved it. The sibling sentence no longer restates the order at all — it defers to the requirement that owns it and notes only that far magic has already run.

## Two claims are now marked provisional rather than quietly kept

`#263` is right that both are wrong, and both are bigger than this change:

1. **A uniform `CERTAIN` for all near magic is unsound.** gzip's two bytes and `.Z`'s two bytes are not XZ's stream-flags CRC, 7z's `StartHeaderCRC`, or TAR's header checksum. Flattening them to one confidence is the same failure mode that let a content probe outrank ISO's far magic.
2. **First-match-wins is not a sound selection rule** where several detectors can accept the same bytes — the three magic-less probes, and genuine polyglots.

Neither blocks prefixed-archive detection, so neither is fixed here. But the spec now says so explicitly, so an implementer can't read either as settled or build on it. The evidence-ledger model, `AmbiguousFormatError` and budget presets belong to that redesign, not to this change.

## An implementability gap nobody here had found

Needles carry an anchor offset **inside their own format**. TAR's `ustar` sits at offset **257**, so a hit at absolute `H` means the candidate begins at `H − 257` — reporting `H` would be wrong by 257 bytes and hand the backend a misaligned source.

Worse, `peek_more(length)` only ever returns the first `length` bytes of the *source*, so it cannot express a candidate-relative view at all. That makes the TAR and compressor needle tasks **unimplementable as written** until a bounded `peek_range(origin, length)` lands. 7z and RAR sidestep it only because their native parsers already accept a start offset — which is why it stayed invisible.

Now task 2.5a as an explicit prerequisite, with 2.5b red–greening the off-by-257.

## The tail probe's "format-bounded" cost was overstated

Two numbers, and I'd conflated them: the `65535 + 22` bound covers **locating** the EOCD, but validating it and computing an exact `payload_offset` can require walking the entire central directory, which that bound does not cover.

Measured over 71,983 files under `/usr`:

| | |
|---|---|
| files ≥ 65,557 B (pay the locator in full) | 3,195 — **4.4%** |
| median file size | 2,239 B |
| **actual aggregate** | **0.61 GiB** |
| worst case if all paid in full | 4.39 GiB |

So the bound is real, and #263's 4.39 GiB figure is the correct *worst case* — but it overstates the actual by ~7×, because the locator is also bounded by the source and most files are small. The cost that likely **does** dominate the founding backup workload is one extra seek per file, and neither of us measured seek latency. Enabling this tier by default is now gated on measuring that (task 1.0) rather than inherited from the phrase "format-bounded".

## Gates

`openspec validate --strict prefixed-archive-detection` clean, `./scripts/check.sh --fix` green. Still proposal-only, so `test.sh` not run.

Note the `dev-docs/IDEAS.md` overlap with #262 — both branches carry the base *Extension-first detection ordering* paragraph on purpose, and whichever merges second drops the duplicate and keeps this branch's agreement-short-circuit variant. Task 0.4a spells that out.

---

PR #259, opened 2026-08-22
OpenSpec apply: brotli-probe-framing-gate (first-block framing gate)

davitf:
## Summary

Implements OpenSpec change `brotli-probe-framing-gate`: first-block Brotli framing gate, LZMA Alone ≤13 reject, probe length via existing `source_byte_size()`, confidence split (compressed-first `PROBABLE` / uncompressed-metadata `GUESS`), and `format_unconfirmed` + `PROBE_FORMAT_UNCONFIRMED` on GUESS read failures.

## Maintainer decisions (locked)

- **D1** — Ship first-block framing only; defer chain walk → task 5.7. Length from `source_byte_size()`.
- **D3** — Keep typed read errors; add `ArchiveyError.format_unconfirmed`; new `PROBE_FORMAT_UNCONFIRMED` (`chosen_by="content_probe"`).
- **D4** — Compressed-first probe-only Brotli → `PROBABLE`; uncompressed/metadata-first → `GUESS`.
- **D6** — Light user-docs update.
- **D7** — Leave §5 follow-ups **unchecked**; **do not archive** (task 4.10 stays open).

## Review round (`/address-review-findings`)

F1–F10 and F12 fixed on `560c7fe`; F11 deferred to task 5.9 (thread left open). F5 Alone residual documented + task 5.8. Pedantic path uses `escalate_as` (review option A).

## Verification

- `./scripts/check.sh --fix` — clean
- `./scripts/test.sh` `[all]` — **2529 passed**, 23 skipped, 3 deselected
- Change remains **in-flight** (not archived)

<div><a href="https://cursor.com/agents/bc-ba18e6da-1a35-4ac7-b539-5cdb8d72e3ec?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ba18e6da-1a35-4ac7-b539-5cdb8d72e3ec&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


davitf (commented):
# Review — PR #259 `brotli-probe-framing-gate` (HEAD `403e95a`)

*Reviewed with `/code-review-skill` (code-first pass 1, then context pass 2 against the OpenSpec change, VISION/CONTRIBUTING and the threat model). Findings `F1`–`F12`; `F1`–`F7`, `F9`–`F11` are inline on the diff.*

## 1. Maintainer briefing

**What this change is.** Brotli has no magic, so detection identified it by decoding a bounded prefix — which accepts ~8% of arbitrary data and 3.5% of a real `/usr` tree, and then fabricates a single `*.uncompressed` member that fails the read with a truncation error naming a format the file never was. This PR adds a *sound* gate instead of a tuned threshold: it parses the first Brotli meta-block header (new `internal/streams/brotli_framing.py`, RFC 7932 §9.1–9.2, no Huffman decoding) and rejects a source whose first *declared* (uncompressed/metadata) block cannot physically fit in the source's byte size; LZMA Alone gets the same invariant's weak form (a source ≤ its 13-byte header). Probe-only Brotli hits are split by evidence class — compressed-first (or `.br`) keeps `PROBABLE`, uncompressed/metadata-first drops to `GUESS` — and a decode failure on a `GUESS` result now stamps `ArchiveyError.format_unconfirmed=True`, rewrites the message, and emits the new `PROBE_FORMAT_UNCONFIRMED` diagnostic.

**Snapshot.** ~960 lines added / 190 removed across 24 files (medium-large; roughly a third is OpenSpec + docs). Gates I re-ran locally on `403e95a`: `./scripts/check.sh` all green (ruff, pyrefly, ty, `openspec validate` 27/27, docs nav + build); `./scripts/test.sh` `[all]` leg **2524 passed, 23 skipped, 3 deselected** in 135s. All 23 CI checks on the PR are green, including `[all-lowest]` and `[core-only]`. I did **not** re-run `--all-configs` locally.

**Verdict: 🔄 Request Changes** — on F1 only; everything else is important-or-below. (Submitted as a comment review: GitHub won't let a review on your own PR carry that state.)

**Main points**

- 🔴 **F1** `DiagnosticPolicy.pedantic()` turns the new diagnostic into a raise *inside* the error path, destroying the typed `TruncatedError`/`CorruptionError` the change exists to preserve. Confirmed repro; `__context__` is a raw `brotli.error`.
- 🟡 **F2** The gate is silently absent on non-seekable sources ≥ 4096 bytes: the exact A-34 stub still opens as a fabricated Brotli member through a pipe. Spec'd, but the user docs state the gate unconditionally and no end-to-end test covers it.
- 🟡 **F3** `PROBE_FORMAT_UNCONFIRMED` is emitted once per *raised exception*, so three retried reads on one member produce three diagnostics.
- 🟡 **F4** `UnconfirmedFormatContext` now carries two different events; its class docstring and the documented meaning of `detected_format=None` are both false for the new `content_probe` variant.
- 🟡 **F5** Task 4.3 is checked off but half-delivered: no COFF fixture, and the OLE assertion is codec-level — end to end an OLE/CFB file is claimed by **LZMA Alone at `PROBABLE`**, so the named residual family gets no unconfirmed channel at all.
- 🟡 **F6** The partial-output regression (task 4.4) branches on `if chunk:` and therefore cannot fail — the behaviour it was written to pin is deterministic.
- 🟡 **F7** The probe callable type widened to `Callable[..., bool]`, dropping arity checking at the registry boundary; the local annotation in `registry.py` still says `Callable[[bytes], bool]`.

**What's fine.** The invariant itself is the right shape, and the zero-false-negative corpus test is the correct binding constraint (I re-ran it; it passes across qualities × lgwin × incompressible payloads). The bit parser matches the reference decoder on the two edges that usually go wrong — I checked metadata-with-`ISLAST` (`b"\x1a"`) and the empty-last stream (`b"\x06"`) against the `brotli` module, and the parser agrees with both. The message rewrite in `_mark_format_unconfirmed` respects the escape-exactly-once rule (rebuild raw, escape the outer). Spec, user docs, `open-issues` P12 and threat-model O10 all move together, and the deferrals (chain walk, archive) are recorded rather than silent.

## 3. Maintainer decisions

1. **Does `pedantic()` get to replace a typed read error? (F1)** The spec keeps `PROBE_FORMAT_UNCONFIRMED` out of `ARCHIVE_INTEGRITY_CODES` for exactly this reason, but `pedantic()` (`default=RAISE`) still catches it, and clause 1 of *A decode failure on probe-only evidence names its provenance* ("keep the same exception type — callers catching those types must keep working") then fails. Options: **A** — emit with `escalate_as=type(exc)` so a RAISE disposition still surfaces a typed error; **B** — never let this code raise (force COLLECT at the emit site) and say so in the spec; **C** — accept it as pedantic's documented posture and add a scenario row saying so. *Recommendation: A* — it uses the mechanism the collector already has for exactly this situation, and keeps the dual-channel contract intact.
2. **Is the unconfirmed channel Brotli-only on purpose, given the Alone residual? (F5)** The `GUESS`/`format_unconfirmed` split is Brotli-scoped because Alone measured 0/20 000 on *random* blobs — but the residual family this change names (OLE/CFB) is structured real-world data, and end to end it is claimed by the **Alone** probe at `PROBABLE` (verified on `main` too, so this is pre-existing, not a regression). A `.doc`/`.msi` therefore still yields `TruncatedError (format=LZMA_ALONE)` with no unconfirmed signal. Options: **A** — leave it, note the gap in P12/O10; **B** — make the unconfirmed channel provenance-based (any probe-only claim with no corroborating extension) rather than Brotli-based, in a follow-up. *Recommendation: A now plus a follow-up task, since B moves a contract this change deliberately scoped down.*
3. **Should the docs carry the "when the source length is known" caveat? (F2)** P12 and O10 have it; `docs/formats.md` and `docs/gotchas.md` state the gate flatly. Either add the caveat, or decide the user-facing simplification is acceptable and say why.

**Findings with no diff line to anchor to:**

- 🟢 **F8 · CONFIRMED** — `DetectionConfidence.GUESS` still documents itself as *"file extension only, no content confirmation"* (`src/archivey/internal/detection.py:92`). After this change a content probe produces `GUESS` too, and `DetectionConfidence` is public API (`archivey.__all__`), so the enum's own comment is now the most misleading description of it in the tree. Fix: reword to something like "no confirmation strong enough to rely on — an extension guess, or a probe hit in the weak evidence class".
- 💡 **F12** — `scripts/exploration/brotli_probe_field_survey.py` carries an independent implementation of the same RFC 7932 header parse plus `SELF_TEST_VECTORS`, and `brotli_framing.py`'s module docstring points at it as "reference behaviour". Nothing ties the two together, so they can drift silently. Porting those vectors into `test_brotli_framing_gate.py` (or cross-checking the two parsers over random blobs) would make drift visible and is a few lines. Keeping the script standalone is right — it runs on machines that don't have archivey installed.

---

davitf (commented):
# Re-review — PR #259 (HEAD `560c7fe`, was `403e95a`)

Round 2 over `560c7fe` ("fix: address review F1–F12"). Status of the previous round first, then the one new finding.

## Status of round 1

| ID | Was | Now | Evidence |
| --- | --- | --- | --- |
| **F1** 🔴 pedantic destroys the typed error | blocking | ✅ **fixed** | `escalate_as=type(exc)` + full kwargs. Re-ran the repro: `CorruptionError`, `format_unconfirmed=True`, `member_name`/`source_format` intact, message preserved. Option A as recommended, and both delta specs got the scenario row. |
| **F2** 🟡 gate absent on non-seekable ≥ 4096 | open | ✅ **fixed** | `docs/formats.md` + gotchas now scope the gate to "when the source length is known"; `test_nonseekable_pipe_skips_gate_at_detection_limit` pins both sizes end-to-end through `PeekableStream`; format-detection delta gained the short-peek row. |
| **F3** 🟡 diagnostic per raised exception | open | ⚠️ **fixed under COLLECT, regressed under RAISE** → see **F13** | default policy now counts 1 (verified); `pedantic()` still counts 3. |
| **F4** 🟡 two events in one context | open | ✅ **fixed** | Docstring now describes both events; probe channel passes `detected_format=self._format.display_name`, pinned by `test_probe_unconfirmed_context_carries_detected_format`. |
| **F5** 🟡 task 4.3 half-delivered | open | ✅ **fixed (honest pin) + deferred** | COFF-shaped fixture added; both families now pin `detect_format` → `LZMA_ALONE`/`PROBABLE` instead of routing around it; task 4.3 text, the framing-gate scenario rows, P12 and O10 all say the Brotli unconfirmed channel doesn't cover them; widening is task **5.8**. That's decision 2 option A, recorded where the next reader will find it. |
| **F6** 🟡 partial-output test can't fail | open | ✅ **fixed** | Now asserts `len(chunk) == 65536` then the raise. Robust: it asserts the requested read size against a stream that has the bytes, not a decoder-internal chunk. |
| **F7** 🟡 `Callable[..., bool]` unchecked boundary | open | ✅ **fixed** | `ContentProbe` Protocol on `CONTENT_PROBES`, the return type and the local. Verified it bites: a backend declaring `def probe(prefix: bytes) -> bool` is now a pyrefly `bad-assignment` error (ty doesn't flag it, but the repo needs both *clean*, and one catching it is what closes the hole). |
| **F8** 🟢 `GUESS` docstring | open | ✅ **fixed** | Reworded to cover both the extension guess and the weak probe class. |
| **F9** 🟢 two meanings of `length` | open | ✅ **fixed** | Comment now states it is an upper bound on reachable bytes and that over-estimating is the safe direction. |
| **F10** 🟢 local imports | open | ✅ **fixed** | Both hoisted to module level. |
| **F11** 💡 inner-TAR corroboration | suggestion | ⏭️ **deferred, task 5.9** | Agreed and recorded; fine by me. |
| **F12** 💡 no tie to the survey's vectors | suggestion | ✅ **addressed** | `test_first_metablock_matches_survey_self_test_vectors` ports all six; I diffed them against `SELF_TEST_VECTORS` and the outcome/MLEN mapping is faithful (`EXUBERANT_NIBBLE` / `BAD_PAD_LAST` → `UNDECIDED` is the right collapse). |

## New finding

- 🟡 **F13** — the F3 dedup flag is assigned *after* an emit that raises under a RAISE policy, so `pedantic()` still records the diagnostic on every retry (3 counts / 3 retained), which is the half of the live `diagnostics` dedup requirement that says "raised again; still no second record". One-line fix (set the flag first); inline for detail.

## Gates

Re-run locally on `560c7fe`: `./scripts/check.sh` green (ruff, pyrefly, ty, `openspec validate` 27/27, docs nav + build); `./scripts/test.sh` `[all]` → **2529 passed, 23 skipped, 3 deselected** in 108s. All 23 CI checks green on this HEAD, `[all-lowest]` and `[core-only]` included. I did not run `--all-configs` locally.

## Verdict

💬 **Comment** — F1 is genuinely resolved and the round was answered finding by finding, including the two "no, and here's why / here's the follow-up" cases, which is the right shape. F13 is the only thing I'd want before merge, and it is one line plus a test on the RAISE path. No maintainer decisions outstanding: decisions 1–3 from round 1 all landed as chosen (A / A + task 5.8 / caveat added).

---

PR #263, opened 2026-08-23
Analyze archive format-detection algorithm

davitf:
## Summary

- derives a format-detection algorithm from measured constraints rather than shipped control flow
- recommends staged evidence collection with totally ranked evidence classes and ordered tie-breakers instead of first-match or additive scoring
- specifies declaration-side evidence ceilings/cost/capabilities, structural validators, stopping rules, non-seekable degradation, Brotli handling, and polyglot behavior
- preserves public `FormatInfo.payload_offset: int` while allowing unknown offsets only on an internal candidate type
- settles tied maximal candidates as a dedicated `AmbiguousFormatError`; records future exhaustive try-open behavior with naming deliberately deferred
- blocks always-on ZIP tail from the default budget until backup-corpus cost measurement passes an explicit release gate
- establishes this investigation as redesign input for PR #257 before implementation
- treats filename agreement as correlated `NAME` evidence: retained and exposed, but unable to promote `BOUNDED_PROBE` or suppress `format_unconfirmed`
- requires winning-evidence exposure on detection results, readers, and unconfirmed errors, independently of any future all-candidates API

## Validation

- `./scripts/check.sh --fix` — passed
- `CXX=g++-13 ./scripts/test.sh --all-configs` — all three configurations passed
  - current/all: 2548 passed, 24 skipped
  - all-lowest: 2548 passed, 24 skipped
  - core-only: 2059 passed, 456 skipped
- CI on the merge-from-main revision is fully green

The compiler override avoids the current VM toolchain mismatch where `/usr/bin/c++` selects GCC 14 headers while only `libstdc++-13-dev` is installed.

<div><a href="https://cursor.com/agents/bc-28bed59c-e7eb-4861-99cc-2193b4b45109?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-28bed59c-e7eb-4861-99cc-2193b4b45109&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


davitf:
## Updated for shipped #267/#268, and resolved a contradiction the revision left behind

Pushed as `a7196cc` (on top of a merge of current `main`, `a3dc408`).

### The contradiction

The revision updated §9 but not the confidence-mapping table, which still read:

> `GUESS` | name-only or an empirically weak bounded probe, notably **uncorroborated** Brotli

That qualifier is exactly what §9's revision removed, so the document said probe + `.br` is `PROBABLE` in one place and `GUESS` in the other.

Resolved by stating what confidence *is* — **a projection of the winning candidate's strongest content-evidence class**, not a second parallel score. A `NAME` item cannot raise it for the same reason it cannot suppress `format_unconfirmed`. `GUESS` therefore means *"the bytes did not confirm this identity"*, not *"probably the wrong format"* — the reframing §6 already applies to `format_unconfirmed`, applied consistently. If that still reads wrong for a public enum, the fix is the labels or the exposed ledger, not re-admitting the filename to the scalar.

**Scope this implies, and it was easy to miss:** zlib and LZMA Alone report `PROBABLE` **unconditionally** today. Both are `BOUNDED_PROBE`, so both become `GUESS`. That is a larger change to shipped behaviour than the Brotli one, and those probes have no name-based branch to draw attention to it. Now stated in both §9 and the mapping table.

### Status

#267/#268 merged as `a3dc408`, so the section now reads as a **scheduled replacement of shipped behaviour** rather than an objection to a pending change.

Added one line against the obvious wrong move: **reverting #267 is not how to remove this rule.** "Old stamps, new does not" reduces to `GUESS` **and** corroborated, which is unreachable — every extension that corroborates a Brotli result has `stream is BROTLI`, exactly what made `_brotli_probe_confidence` report `PROBABLE`; the inner-TAR arm forces `PROBABLE`; zlib and Alone are unconditionally `PROBABLE`. Reverting restores the larger blind spot (Alone and zlib never stamping at all) while leaving the filename rule in place via confidence.

### Re-measured on `a3dc408`

Independent tree, 63 343 files: 23 content-probe claims, 19 fabrications (0.030%), **all 19 stamping**, and again **zero corroborated fabrications**. So the false-positive argument reconfirms across two trees.

The cost side is now sized rather than asserted. The same run found 4 genuine streams — and the whole tree contains exactly 4 files with any magic-less-codec extension, the same 4:

| genuine stream | name | corroborated | stamps today |
| --- | --- | --- | --- |
| `underscore.min.js.br` | `.br` | yes | no |
| `underscore.min.js.map.br` | `.br` | yes | no |
| `jquery.min.js.brotli` | `.brotli` | **no** | **yes** |
| `jquery.min.map.brotli` | `.brotli` | **no** | **yes** |

`.brotli` is not a registered extension — only `.br` and `.tar.br` are — so two genuine Brotli streams already carry `format_unconfirmed` on shipped `main`. Removing extension corroboration therefore newly affects **two files here**, on a tree where two of the same kind already pay that cost via an unrelated naming gap.

I also added the caveat that cuts against reading too much into it: `/usr` is the friendliest possible sample for filenames, so 4-of-4 genuine should not be taken as establishing the extension as a discriminator on the backup corpus `VISION.md` names as the founding workload.

### Follow-up scope: two sites

`_extension_corroborates` (#267) and `_brotli_probe_confidence`'s `.br`-to-`PROBABLE` rule (#261) are the same rule expressed twice. Removing only the first leaves the second contradicting this mapping. `openspec/specs/error-handling` on `main` already records that; §9 and item 12 now match.

### Separate, not folded in

`.brotli` being unregistered is a detection-data gap, not a design question — it makes real streams look uncorroborated however §6 resolves. Worth its own item; I haven't opened one.

---

PR #265, opened 2026-08-25
Probe completeness gate and Brotli chain walk

davitf:
## Summary

Adds the **completeness** rejection for fully-visible incomplete streams and a Brotli **chain walk** past the first meta-block (`probe-completeness-gate`). OpenSpec archived and synced.

### Review rounds

**F1–F12** (`617d9f0`) — all Fixed; approved on re-review.

**N1–N3** (`645a1df`) — Fixed:

| ID | Disposition |
|---|---|
| N1 | **Fixed** — re-ran census after 64 KiB drain: **29 / 150 623 (0.019%)**; docs + investigation updated; annotated older 91-of-128 figure |
| N2 | **Fixed** — drain loop restructured (`if produced >= budget` + comment covering reject-at-boundary) |
| N3 | **Fixed** — track `produced` only; no join |

### Gates

- `./scripts/check.sh --fix` — all passed
- `./scripts/test.sh` (`[all]`) — **2548 passed**, 24 skipped

### Residual (post-64 KiB drain)

[probe_residual_census_64kib.log](https://cursor.com/agents/bc-cc36700f-d59d-471e-8b5b-7259dbc36ff4/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fprobe_residual_census_64kib.log)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-cc36700f-d59d-471e-8b5b-7259dbc36ff4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-cc36700f-d59d-471e-8b5b-7259dbc36ff4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/archivey/internal/detection.py
```diff
@@ -487,8 +537,17 @@ def peek_more(length: int) -> bytes:
         length = source_byte_size(source)
```
davitf: **F1 🔴 `[blocking]` · CONFIRMED — chain walk rejects complete valid Brotli in a mid-positioned stream**

Two coordinate systems meet in `chain_proves_invalid`:

- `read_at(offset, …)` is **archive-relative** — line 179 captures `origin = source.tell()` and seeks `origin + offset`, and its docstring says so ("relative to the archive origin — the position detection started from").
- `source_length` here is `source_byte_size(source)`, the **total size from offset 0**.

For a stream handed in positioned at offset K the two differ by K. The comment just above (lines 531-535) reasons that this over-estimate is "the safe direction (never rejects a complete valid stream)". That was true for `first_block_overruns_source`, which only tests `>`. The chain walk adds *equality* tests (`brotli_framing.py:238,244`), and under an over-estimate a correctly-terminating stream reads as "declared end with trailing bytes" → rejected. **The comment's premise stopped holding in this PR and the comment was left in place.**

**Why it matters.** Lines 9-13 of this module and `core.py:178-183` both make mid-positioned seekable streams a supported contract ("an archive embedded mid-file works uniformly, without manual slicing"), and detection runs *before* the opener's zero-origin wrap. It is a regression against `main`, and it falsifies two claims this PR adds to `openspec/specs/format-detection/spec.md`: "costing **zero** genuine streams" and "budget exhaustion can never manufacture a false negative".

**Trigger** (red–green candidate; `[all]` — needs brotli):

```python
import io, os, brotli
from archivey import open_archive

raw = os.urandom(1000)
payload = brotli.compress(raw, quality=0)   # incompressible -> uncompressed first meta-block
buf = io.BytesIO(b"\x00" * 4096 + payload)
buf.seek(4096)                              # archive begins at the current position
with open_archive(buf) as a:                # PR: FormatDetectionError;  main: opens fine
    assert list(a)[0] is not None
```

Measured: `main` returns `BROTLI (content_probe)` at offsets 0, 16 and 4096 and the `open_archive` round-trip returns the original bytes. This PR returns `BROTLI` at offset 0 only, and `FormatDetectionError` at 16 and 4096, for `BytesIO` and a real file alike. Isolated to the walk:

```
chain_proves_invalid(prefix, 1004)        -> False   # true length
chain_proves_invalid(prefix, 1004 + 16)   -> True    # what source_byte_size reports at K=16
first_block_overruns_source(prefix, 1020) -> False   # old rule unaffected
```

`brotli.compress(os.urandom(n), quality=0)` produces an uncompressed first meta-block at every size I tried (100 B → 20 KB), so this is an ordinary shape, not a contrived one.

**Fix direction** (maintainer decision 1 settled — keep the rule, fix the yardstick): make the length archive-relative rather than dropping the equality checks. Subtract the stream's current position from the total size here; paths are already origin-0, and the short-peek fallback on the next line is already remaining-bytes. Then rewrite the 531-535 comment, whose safe-direction argument no longer covers the chain rule.

Please add a mid-positioned case to `tests/test_probe_completeness_gate.py` — the suite has no coverage of that contract today, which is why 23 green checks missed this.

---

src/archivey/internal/streams/brotli_framing.py
```diff
@@ -142,10 +157,90 @@ def parse_first_metablock(prefix: bytes) -> BrotliFraming:
...
+    def _get(offset: int, length: int) -> bytes | None:
+        end = offset + length
+        if offset < len(prefix):
+            # Serve whatever of the request sits in the prefix. A short chunk here is
+            # EOF-within-prefix (caller may still have a read_at for the rest).
+            in_prefix = prefix[offset : min(end, len(prefix))]
+            if end <= len(prefix):
+                return in_prefix
+            if read_at is None:
+                return in_prefix  # short = EOF for prefix-only walks
```
davitf: **F2 🟡 `[important]` · CONFIRMED — prefix-only walk conflates "prefix exhausted" with EOF**

When the request straddles the prefix end and `read_at is None`, `_get` returns the short in-prefix slice with the comment `# short = EOF for prefix-only walks`. But `source_length` — right there in the caller's signature — says the source is bigger. The short chunk then fails to parse, `parse_metablock` returns `UNDECIDED`, and line 232 turns that into `return True` ("proves invalid").

**Trigger.** On a *valid* 300 880-byte stream (`brotli.compress(os.urandom(300000), quality=0, lgwin=10)`), mid-chain headers sit at offsets 1027, 2054, 3081. With `read_at=None` and the correct `source_length`:

```
chain_proves_invalid(d[:1028], len(d), read_at=None) -> True   # valid stream "proven invalid"
chain_proves_invalid(d[:1029], len(d), read_at=None) -> True
chain_proves_invalid(d[:1028], len(d), read_at=rd)   -> False  # correct
```

Six false rejections across the first three links (prefix ending 1–2 bytes into a header).

**Why it matters.** `detection.py:549` is the only production caller and always passes `read_at`, so this is latent — but `read_at` is optional in the `ContentProbe` protocol (`registry.py:55-62`) and this function's own docstring advertises the mode ("When omitted, only bytes already in `prefix` are reachable"). The next caller inherits a rule that can manufacture the false negative the new spec forbids.

**Fix direction.** Treat prefix exhaustion as EOF only when `len(prefix) >= source_length`; otherwise return `None` (cannot disprove), which is what every other bail-out in this function does.

---

src/archivey/internal/detection.py
```diff
@@ -138,6 +143,51 @@ def _peek_prefix(source: str | Path | BinaryIO, length: int) -> bytes:
...
+            # Growing peek buffers [0, end); refuse past the non-seekable cap above.
+            buf = source.peek(end)
+            if offset >= len(buf):
+                return b""
+            return buf[offset:end]
+        if is_seekable(source):
+            origin = source.tell()
+            try:
+                source.seek(origin + offset)
+                return source.read(length)
```
davitf: **F3 🟡 `[important]` · CONFIRMED — `read_at` uses a bare `read()`, so a legal short read reads as a bad header**

`source.read(length)` may legally return fewer bytes than asked without being at EOF. The chain walk cannot distinguish that from a truncated header, so it reaches `UNDECIDED` → `chain_proves_invalid` line 232 → `return True` → rejection.

`read_exact` is imported into this very module and its docstring names the hazard: *"Unlike a single `read(n)` (which may legally return a short chunk), this loops until `n` bytes are gathered or EOF"*. `_peek_prefix` (lines 133-137) already uses it. Same file, two conventions.

**Trigger.** A seekable `BinaryIO` with an fsspec-style `size` attribute whose `read()` caps at 2 bytes, holding a valid `brotli.compress(os.urandom(300000), quality=0, lgwin=10)`:

```
max read() chunk 1048576: BROTLI (content_probe)
max read() chunk       3: BROTLI (content_probe)
max read() chunk       2: FormatDetectionError  <-- valid brotli rejected
max read() chunk       1: FormatDetectionError  <-- valid brotli rejected
```

Low likelihood for local file objects; archivey accepts arbitrary `BinaryIO`, and network/fsspec-backed streams short-read.

**Fix direction.** `read_exact(source, length)` in this branch, and in the path branch at lines 176-178.

---

src/archivey/internal/streams/codecs.py
```diff
@@ -969,23 +985,35 @@ def probes_content(self) -> bool:
...
+        # Feed the whole source when it is fully visible so "needs more input" means
+        # incomplete; otherwise keep the bounded probe sample.
+        sample = prefix[:source_length] if fully_visible else prefix[:_PROBE_PREFIX]
         try:
-            with open_codec_stream(
-                self.codec, io.BytesIO(prefix[:_PROBE_PREFIX])
-            ) as stream:
-                stream.read(_PROBE_PREFIX)
+            with open_codec_stream(self.codec, io.BytesIO(sample)) as stream:
+                stream.read(max(len(sample), _PROBE_PREFIX))
```
davitf: **F4 🟡 `[important]` · CONFIRMED — the completeness gate never decodes to termination, but spec and docstring say it does**

`stream.read(max(len(sample), _PROBE_PREFIX))` caps *output* at 256 bytes or the compressed sample length. Compression means output is routinely far larger, so the decoder never reaches the point of wanting more input and the gate is inert.

The docstring at 998-1004 says "the probe holds the whole source. A decode that ends wanting more input is then a rejection — a complete valid stream decoded in full terminates", and the spec matrix row is unqualified: `Any probe, source_length <= len(prefix), decode wants more input | Reject`.

**Trigger.** `zlib.compress(b"A" * 200000, 9)` is 218 bytes; truncate it and hand the whole thing to the probe:

```
truncated to  20B (expands to  1808B, read cap 256B) -> ZlibCodec.content_probe(...) == True
truncated to  60B (expands to 43088B, read cap 256B) -> True
truncated to 100B (expands to 84368B, read cap 256B) -> True
```

Every one of these is fully visible and incomplete — the exact case the requirement says SHALL be rejected.

**Why it matters.** The direction is conservative (missed rejections, not false ones), so this is contract honesty rather than a correctness break. But it lands in a spec that is otherwise scrupulous about this, explicitly requiring the chain walk's bounds be "declared rather than implicit" while leaving this one implicit. As written, a second implementer reading the requirement would build a stronger gate than the one measured.

**Fix direction** (maintainer decision 2 settled — raise the cap, then document): bump the output read to ~64 KiB and declare it in both the docstring and the requirement, softening the requirement text from an unqualified invariant to a bounded check. That catches the small-to-medium truncations that make up most of the measured fabrication set while keeping the worst case well below a full drain — a fully-visible 4 KiB source can expand to 4 MB through zlib alone (measured), and further through brotli/lzma, so an uncapped drain would itself be a bomb-budget decision.

Note the same cap appears twice — see F6 on line 1370.

---

src/archivey/internal/streams/brotli_framing.py
```diff
@@ -1,24 +1,32 @@
...
 
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import Enum
 
+# Declared budgets (hitting either = cannot disprove; keep the earlier verdict).
+# Link cap 8: real-tree census rejected every fabrication by link index ≤ 1;
+# revisit with hard data if a future corpus shows deeper rejecting chains.
+CHAIN_MAX_LINKS = 8
+CHAIN_MAX_BYTES_FETCHED = 4096
```
davitf: **F5 🟡 `[important]` · CONFIRMED — the declared bytes-fetched budget cannot bind and does not measure the cost**

Two problems in one constant.

**(i) It can never fire.** `CHAIN_MAX_LINKS × CHAIN_HEADER_READ = 8 × 24 = 192`, so `CHAIN_MAX_BYTES_FETCHED = 4096` is unreachable — the link cap always binds first and the byte budget is dead code.

**(ii) It measures the wrong thing.** The counter tracks bytes *returned to the walk*, not bytes read or buffered. Measured on a `PeekableStream`:

```
read_at(600000, 24) -> 24 bytes returned;
PeekableStream buffer now holds 600024 bytes (586 KiB) for a walk charged 24 bytes
```

because `detection.py:180-184` calls `source.peek(end)`, which grows the buffer over the whole `[0, end)` range. The bound that actually governs memory is `_PROBE_READ_AT_MAX_OFFSET_NONSEEKABLE` (1 MiB) in `detection.py:95` — not the "4 KiB" the PR body lists as a settled decision, and not mentioned in the spec.

`bytes_fetched += len(chunk)` (line 228) also charges prefix-served bytes that cost no I/O.

**Why it matters.** The new requirement says the walk SHALL be "bounded in both the number of links followed and the bytes read, and both bounds SHALL be declared rather than implicit". One of the two declared bounds is nominal, and the bound that actually governs memory is declared elsewhere, an order of magnitude larger.

**Fix direction** (maintainer decision 3 settled — declare the real ceiling): drop `CHAIN_MAX_BYTES_FETCHED` and the `bytes_fetched` accounting, keep the link cap as the single declared work bound, and write the 1 MiB forward-only ceiling into the requirement so the honest number sits where a reader will look for it. Detection may still buffer up to 1 MiB on a non-seekable source before declining — that is the fact to state, rather than a 4 KiB budget that never applies.

---

src/archivey/internal/streams/codecs.py
```diff
@@ -1315,25 +1343,36 @@ def extract_metadata(self, ctx: MetadataContext, member: ArchiveMember) -> None:
...
             return False
+        fully_visible = source_length is not None and source_length <= len(prefix)
+        sample = prefix[:source_length] if fully_visible else prefix[:_PROBE_PREFIX]
         try:
-            with open_codec_stream(
-                self.codec, io.BytesIO(prefix[:_PROBE_PREFIX])
-            ) as stream:
-                out = stream.read(_PROBE_PREFIX)
+            with open_codec_stream(self.codec, io.BytesIO(sample)) as stream:
+                out = stream.read(max(len(sample), _PROBE_PREFIX))
```
davitf: **F6 🟡 `[important]` · CONFIRMED — the completeness rule is implemented twice**

`LzmaAloneCodec.content_probe` (lines 1366-1376) reproduces `_decodes_sample`'s `fully_visible` / `sample` / `read(max(...))` / `except TruncatedError: if fully_visible: return False` verbatim. The only difference is the `len(out) > 0` positive-output requirement.

Two copies of one invariant that must now stay in sync — and every fix from F4 has to be applied twice, to both output caps.

This also cuts against the requirement text this PR adds: *"This is available to every probe without any interface change."* It required a second hand-written copy for the second probe.

**Fix direction.** `_decodes_sample(prefix, *, source_length=None, require_output=False)`, and have `LzmaAloneCodec.content_probe` call it after its header-plausibility gate. Per CONTRIBUTING's clean-as-you-go rule this is a pay-now, not a keep-with-reason.

---

src/archivey/internal/streams/brotli_framing.py
```diff
@@ -126,9 +140,10 @@ def _metablock(br: _Bits) -> BrotliFraming:
...
             consumed=br.pos // 8,
             declared_length=declared,
+            is_last=False,
         )
     # Compressed body — Huffman tables follow; we stop without decoding them.
-    return BrotliFraming(BrotliFirstBlock.COMPRESSED)
+    return BrotliFraming(BrotliFirstBlock.COMPRESSED, is_last=islast)
 
 
 def parse_first_metablock(prefix: bytes) -> BrotliFraming:
```
davitf: **F7 🟢 `[nit]` · CONFIRMED — `parse_first_metablock` is now an exact duplicate**

Lines 149-157 are byte-for-byte `parse_metablock(prefix, first=True)` (lines 160-168) — same eight lines under two names. This is the half of the naming cleanup that can actually drift: a fix to one won't reach the other.

**Fix direction** (maintainer decision 4 settled — rename and collapse both): collapse this to a one-line delegation or delete it and update the call sites. 24 references across five files, 4 of them in `src/`.

---

src/archivey/internal/streams/brotli_framing.py
```diff
@@ -1,24 +1,32 @@
...
 
+# Declared budgets (hitting either = cannot disprove; keep the earlier verdict).
+# Link cap 8: real-tree census rejected every fabrication by link index ≤ 1;
+# revisit with hard data if a future corpus shows deeper rejecting chains.
+CHAIN_MAX_LINKS = 8
+CHAIN_MAX_BYTES_FETCHED = 4096
+CHAIN_HEADER_READ = 24
+
 
 class BrotliFirstBlock(Enum):
```
davitf: **F8 🟢 `[nit]` — `BrotliFirstBlock` is no longer about first blocks**

The docstring was correctly generalized here ("Outcome of parsing one meta-block header"), and so was `BrotliFraming`'s — but the enum name and every call site still say `First`. The walk now reads `BrotliFirstBlock.COMPRESSED` while inspecting the fifth block in a chain.

**Fix direction** (maintainer decision 4 settled — rename and collapse both): rename to drop `First`. 49 references across five files, 23 of them in `src/` and the rest mechanical test churn.

---

dev-docs/open-issues.md
```diff
@@ -255,28 +255,24 @@ nor `DIRECTORY`. Original write-up below.
...
   `format_unconfirmed=True` and emits `PROBE_FORMAT_UNCONFIRMED`.
 - **Still three clauses, not one:** the listing can be wrong; a full read raises; **and**
   a prefix of fabricated bytes may already have been produced. Not a silent success.
-- **Re-measured after the gate** (`scripts/exploration/probe_residual_census.py`,
-  66 361 files under `/usr`): 132 probe claims, 4 of them genuine `.br`/`.brotli` files,
-  **128 fabricated (0.193%)**. Of those, **68 — 53% — carry no unconfirmed signal at
-  all**, because the channel keys on `GUESS` while LZMA Alone reports `PROBABLE`
-  unconditionally.
-- **Refs:** `openspec/changes/archive/2026-08-23-brotli-probe-framing-gate/`; investigation
+- **Refs:** `openspec/changes/probe-completeness-gate/`; investigation
```
davitf: **F9 🟢 `[nit]` · CONFIRMED — two doc refs point at the pre-archive OpenSpec path**

This line cites `openspec/changes/probe-completeness-gate/`, but the change was archived in this same PR to `openspec/changes/archive/2026-08-25-probe-completeness-gate/`. Verified: the unarchived path no longer exists in the tree.

Note this line *replaced* a correct archive ref (`.../archive/2026-08-23-brotli-probe-framing-gate/`) with the live one, and the sibling ref for the older change two lines up is still correctly archived — so it reads as an oversight rather than a convention.

Same fix needed at `dev-docs/threat-model.md:371`, which cites the same non-existent path directly beside a correctly-archived one.

---

tests/test_sfx.py
```diff
@@ -433,8 +435,17 @@ def test_a_weak_cue_still_lets_a_content_probe_answer(tmp_path: Path) -> None:
     assert framing.outcome is BrotliFirstBlock.UNCOMPRESSED
     assert framing.consumed is not None and framing.declared_length is not None
     need = framing.consumed + framing.declared_length
+    # Pad a compressed second link so the chain walk stops rather than rejecting
+    # exact EOF-without-ISLAST (same residual shape as OLE/COFF fixtures).
```
davitf: **F10 🟢 `[nit]` — duplicated fixture helper and a function-local import**

This 24-byte compressed-second-header search loop is the same code as `_compressed_second_header()` in `tests/test_probe_completeness_gate.py:32-38`. Worth lifting into a shared test util rather than keeping two copies that have to agree about what "a header that classifies as COMPRESSED" means.

Line 428 also imports `parse_metablock` inside the function while the module imports `BrotliFirstBlock` at the top — worth hoisting for consistency.

The fixture change itself is right, and the comment explaining *why* the fixture had to grow a compressed second link is exactly the kind of note that keeps this test meaningful after the walk landed. 🎉

---

tests/test_probe_completeness_gate.py
```diff
@@ -0,0 +1,292 @@
...
+    assert detect_format(io.BytesIO(data)).format == ArchiveFormat.ZLIB
+    assert ZlibCodec().content_probe(data, source_length=len(data)) is True
+
+
+def test_lzma_alone_completeness_on_fully_visible_nonterminating() -> None:
+    alone = LzmaAloneCodec()
+    header = bytes([0x5D, 0x00, 0x00, 0x01, 0x00]) + b"\xff" * 8
+    blob = header + b"\x00" * 40
+    # When the bounded sample matches but the whole source is visible and does not
+    # terminate, completeness must reject.
```
davitf: **F11 🟢 `[nit]` — a self-disabling test**

The assertion is wrapped in `if alone.content_probe(blob, source_length=None):`. The guard is `True` today — I checked — so the test does run. But if the prefix-only verdict ever flips, this test silently passes while asserting nothing, which is the opposite of what you want from the one test covering the completeness rule for LZMA Alone.

**Fix direction.** Assert the guard rather than branching on it:

```python
assert alone.content_probe(blob, source_length=None) is True
assert alone.content_probe(blob, source_length=len(blob)) is False
```

---

src/archivey/internal/detection.py
```diff
@@ -138,6 +143,51 @@ def _peek_prefix(source: str | Path | BinaryIO, length: int) -> bytes:
...
+            return None
+        if isinstance(source, (str, Path)):
+            with open(source, "rb") as f:
+                f.seek(offset)
+                return f.read(length)
+        if isinstance(source, PeekableStream):
+            # Growing peek buffers [0, end); refuse past the non-seekable cap above.
+            buf = source.peek(end)
+            if offset >= len(buf):
+                return b""
```
davitf: **F12 💡 `[suggestion]` — `read_at` re-opens the path on every link**

`with open(source, "rb")` runs per call, so a chain walk on a path source opens the file up to 7 more times, and successive links can observe a file that changed under them.

This matches `_peek_prefix`'s open/close style, so it is not a defect — but a single handle held for the duration of the walk would be cheaper and self-consistent. Non-blocking.

(Note this branch also needs the `read_exact` fix from F3 — a short `f.read(length)` here has the same consequence as on line 182.)

---

davitf (commented):
## Maintainer briefing

Reviewed `fc6bcc1` (base `main` @ `3dafac1`) with `/code-review-skill` + the archivey addendum. Findings `F1`–`F12` are inline at their locations.

**What this change is.** Two new rejection rules for the magic-less content probes. (a) *Completeness*: when the peeked prefix holds the whole source, a decode that ends wanting more input is a rejection rather than a match — applied in `_decodes_sample`, LZMA Alone and zlib. (b) *Chain walk*: `chain_proves_invalid` follows Brotli's byte-aligned self-describing meta-blocks past the first, rejecting a link that overruns the source or a declared end with trailing bytes; `detection.py` hands probes a bounded `read_at(offset, length)` so the walk can reach successor offsets past the 4 KiB prefix. Plus specs, docs, threat-model O10, and the OpenSpec archive.

**Snapshot.** ~849/163 lines across 19 files — medium. CI: **23/23 green** on this HEAD. Re-run locally on this image rather than copied from the PR body — `[all]` config, `pytest tests/test_probe_completeness_gate.py tests/test_brotli_framing_gate.py tests/test_detection.py tests/test_sfx.py` → **112 passed, 1 skipped**, matching the PR's claim.

**Verdict: 🔄 Request Changes** — on `F1` alone. (Posted as a comment rather than a formal request-changes review, since GitHub won't let a review be filed against one's own PR.)

**Main points.**

| | Finding | Gist |
|---|---|---|
| 🔴 | `F1` | Chain walk rejects **complete, valid Brotli in a mid-positioned stream**. `source_length` is measured from byte zero; chain offsets are measured from the archive start. The new equality checks turn that mismatch into a rejection. Confirmed regression vs `main`, end-to-end through `open_archive`. |
| 🟡 | `F2` | Prefix-only walk treats prefix exhaustion as EOF and reports "proves invalid" for a valid stream. Not reachable from `detection.py` today, but it is the function's documented default. |
| 🟡 | `F3` | `read_at` uses a bare `read()`; a legal short read is misread as a malformed header → rejection. `read_exact` sits in the same module with a docstring warning about exactly this. |
| 🟡 | `F4` | The completeness gate never decodes to termination — spec and docstring promise a stronger rule than the code implements. |
| 🟡 | `F5` | The declared 4 KiB byte budget can never bind (8 × 24 = 192 B) and doesn't measure the real cost: one 24-byte fetch buffered 586 KiB. |
| 🟡 | `F6` | The completeness rule is now implemented twice, so every fix has to land in both copies. |
| 🟢 | `F7`–`F12` | Duplicate parser, stale enum name, two dead doc links, duplicated test helper, a self-disabling test, per-link file reopening. |

**What's fine.** The walk logic itself is sound wherever `source_length` is correct: I differential-fuzzed 540 valid Brotli streams (7 payload shapes × qualities 0–11 × 3 `lgwin` values) against the real `brotli` decoder and found **zero** false rejections, with and without `read_at`. The "cannot disprove → keep the earlier verdict" discipline is applied consistently at every bail-out. The docs/threat-model/spec rewrites re-measure rather than recycle stale numbers. And `is_last` on a metadata meta-block — the one normative claim in the parser I distrusted on sight — does match real brotli; I checked.

---

## Maintainer decisions

All four were put to the maintainer and are **settled**. Recording them here so the implementing agent doesn't re-open them.

1. **Trailing-bytes rejection → keep the rule, fix the yardstick.** Measure the source length from the archive's start position rather than from byte zero, so the check compares like with like. Context: on this image's `/usr` (76 214 files), of 32 files the Brotli probe would otherwise accept, the walk's overrun rule rejects 7 and the trailing-bytes rule rejects exactly **1** more — `/usr/lib/jvm/java-21-openjdk-amd64/lib/modules`, 140 MB, which is precisely the large-source class the walk was built for. Small yield, landing on target, so it is worth keeping rather than dropping to dodge the bug. Details and repro in `F1`.

2. **Completeness bound → raise the cap, then document.** Bump the output read to ~64 KiB and declare it in both the docstring and the requirement, softening the requirement text from an unqualified invariant to a bounded check. Rationale: a full drain is itself a bomb-budget decision — a fully-visible 4 KiB source expands to 4 MB through zlib alone (measured), further through brotli/lzma. Details in `F4`; the second copy of the cap is `F6`.

3. **Chain-walk budget → declare the real ceiling.** Drop the dead 4 KiB counter and its accounting, keep the link cap as the single declared work bound, and write the 1 MiB forward-only ceiling into the requirement, where a reader will actually look for it. Details in `F5`.

4. **Naming → rename and collapse both.** Rename the block enum to drop `First`, and collapse the duplicate parser. ~73 mechanical references across five files, roughly two-thirds of them tests. Details in `F7` and `F8`.

One consequence worth flagging: decisions 2 and 3 both move requirement text in `openspec/specs/`, so the spec edits land in this same change rather than as a follow-up.

---

dev-docs/threat-model.md
```diff
@@ -348,25 +348,27 @@ Brotli has no magic number, so `detect_format` recognizes it by a content probe.
...
-claimed end-to-end by the **LZMA Alone** probe at `PROBABLE`, so that honesty channel
-does not cover them yet — measured, **68 of 128** fabricated probe claims on a real
-tree carry no signal (follow-up: `probe-provenance-unconfirmed`). A further
-`probe-completeness-gate` rejects a source that is fully visible and does not decode to
-completion, removing 91 of those 128.
+**Mitigation shipped (framing + completeness + chain walk):** when the source length is
+known, a first meta-block that declares more bytes than the source holds is rejected; a
+fully visible source that does not decode to completion is rejected; and a bounded
+self-describing block-chain walk rejects later overruns / trailing bytes. Residual after
+`probe-completeness-gate` on a re-measured tree (147 601 files): **30 fabricated claims
```
davitf: **N1 🟡 `[important]` · CONFIRMED — the measured residual now understates the mitigation**

This "**30 fabricated claims (0.020%)**" figure was measured when the completeness gate drained **256 bytes** of output. `F4`'s fix raised that to **64 KiB** — 256× — so the published number predates the code it describes.

**Measured.** Re-running my `/usr` census (76 214 files, same script both times, Brotli-probe layer):

| | `fc6bcc1` (reviewed) | `617d9f0` (now) |
|---|---|---|
| accepted before the walk | 32 | **29** |
| rejected by the walk's overrun rule | 7 | 7 |
| rejected only by trailing-bytes | 1 | 1 |
| **surviving both (residual)** | **24** | **21** |

The larger drain rejects 3 more files pre-walk, cutting the surviving residual by ~12.5%. (My corpus is not the PR's 147 601-file census and this is the probe layer, not end-to-end — the *direction* transfers, the magnitude won't.)

Worth noting the two rules held steady across the fix: the trailing-bytes contribution is still exactly 1, which is independent confirmation that the `F1` yardstick correction did not weaken the rule it was protecting.

**Why it matters.** These numbers are cited as evidence of how well O10 is mitigated, and the same figure appears in `dev-docs/open-issues.md:263` and — as "**91 of 128**" — in the completeness requirement at `openspec/specs/format-detection/spec.md:515`. It errs conservative, so nothing is overclaimed, but the addendum's rule is that measurements get re-run rather than carried forward, and docs move with the contract in the same change.

**Fix direction.** Re-run `scripts/exploration/probe_residual_census.py` on the new HEAD and update all three sites; or, if a full re-census isn't worth it now, annotate the figure as measured at the 256-byte drain so the next reader doesn't treat it as current.

---

src/archivey/internal/streams/codecs.py
```diff
@@ -969,23 +989,69 @@ def probes_content(self) -> bool:
...
+            with open_codec_stream(self.codec, io.BytesIO(sample)) as stream:
+                if fully_visible:
+                    # Drain up to the budget in chunks. A truncated high-ratio stream
+                    # often yields its whole expansion on the first large read without
+                    # raising; the next read is what surfaces TruncatedError. Hitting
+                    # the budget with more output still available means the bounded
+                    # check cannot disprove completeness — accept.
+                    parts: list[bytes] = []
+                    produced = 0
+                    while produced < out_budget:
```
davitf: **N2 🟢 `[nit]` — the comment undersells what the `while…else` actually does**

The code here is right — I probed the boundary and it is exact:

```
complete   out=65535 accept   out=65536 accept   out=65537 accept   out=73728 accept
truncated  →65535 reject     →65536 reject     →65537 accept     →73728 accept
```

A complete stream is never falsely rejected, and a truncation that ends *exactly* on the budget is still caught. That last case only works because the `else` branch does one more `stream.read(1)` past the budget — that read is what surfaces `TruncatedError` at the boundary.

The comment above says budget exhaustion means "the bounded check cannot disprove completeness — accept", which describes only one of the two outcomes: hitting the budget with more output available accepts, but hitting it with the stream truncated at exactly that point rejects. Since that boundary read is deliberate and load-bearing, it deserves a sentence saying so.

Also worth noting `while…else` is a well-known Python footgun — most readers parse `else` as "if the loop body never ran" rather than "if the loop finished without `break`". Given CONTRIBUTING's "resulting code is self-explanatory", either name the `else` in the comment or restructure as an explicit `if produced >= out_budget:` after the loop.

---

src/archivey/internal/streams/codecs.py
```diff
@@ -969,23 +989,69 @@ def probes_content(self) -> bool:
...
+                        chunk = stream.read(min(8192, out_budget - produced))
+                        if not chunk:
+                            break
+                        parts.append(chunk)
+                        produced += len(chunk)
+                    else:
+                        more = stream.read(1)
+                        if more:
+                            parts.append(more)
+                    out = b"".join(parts)
```
davitf: **N3 🟢 `[nit]` — the accumulated output is only needed for `require_output`**

`parts` is collected and joined on every fully-visible probe, but the result is used only by the `require_output` branch (LZMA Alone). For Brotli and zlib the join allocates up to 64 KiB and the value is discarded.

Tracking `produced` — which the loop already maintains — is enough: `require_output` needs `produced > 0`, nothing more. Dropping `parts` and the join would also make the drain loop read as what it is, a bounded discard.

Small, and detection only reaches this path for sources that fit in the peeked prefix, so the cost is bounded. Non-blocking.

---

davitf (commented):
## Re-review — `617d9f0` (was `fc6bcc1`)

Second pass over commit `617d9f0` "Address review F1–F12". Status of the previous round first, then what's new.

### Previous round: 12 of 12 addressed

| ID | Sev | Status | Verified how |
|---|---|---|---|
| `F1` | 🔴 | **Fixed** | Length is now archive-relative (`length - source.tell()` for mid-positioned seekable sources). My original repro passes at offsets 0/16/4096 for `BytesIO` and a real file, and `open_archive` round-trips at offset 4096. Regression test added. |
| `F2` | 🟡 | **Fixed** | Prefix exhaustion is EOF only when `len(prefix) >= source_length`, else declined. The 6 false rejections on a valid 300 KB stream are now 0. Regression test added, citing the exact cuts. |
| `F3` | 🟡 | **Fixed** | `read_exact` in both the path and seekable branches. A stream capped at 1-byte reads now detects `BROTLI`; it raised `FormatDetectionError` before. |
| `F4` | 🟡 | **Fixed** (decision 2) | Declared 64 KiB drain. Truncations expanding to ~1.8 KB / 22 KB / 43 KB / 64 KB now reject; past the drain they don't, and the requirement now says so in its own matrix row. Regression test added. |
| `F5` | 🟡 | **Fixed** (decision 3) | `CHAIN_MAX_BYTES_FETCHED` and its accounting removed; the 1 MiB forward-only ceiling is declared in the requirement, with the seekable-vs-forward-only distinction spelled out. |
| `F6` | 🟡 | **Fixed** | `_decodes_sample(..., require_output=True)`; the duplicated block in `LzmaAloneCodec.content_probe` is gone. One implementation. |
| `F7` | 🟢 | **Fixed** | `parse_first_metablock` deleted; call sites use `parse_metablock(..., first=True)`. |
| `F8` | 🟢 | **Fixed** | Renamed to `BrotliBlock`. No `BrotliFirstBlock` / `parse_first_metablock` references remain anywhere. |
| `F9` | 🟢 | **Fixed** | Both refs now point at the archive path; no non-archive references remain. |
| `F10` | 🟢 | **Fixed** | Helper lifted to `tests/streams_util.brotli_compressed_metablock_header`, used by both files. |
| `F11` | 🟢 | **Fixed** | The `if` guard is now two asserts, so the prefix-only verdict is pinned rather than assumed. |
| `F12` | 💡 | **Fixed** | One handle per walk, closed in a `finally` around the probe loop. |

Nothing regressed: differential fuzz against the real `brotli` decoder over 204 valid streams (payload shapes × qualities × `lgwin`) still finds **zero** false rejections.

**Gates, re-run on `617d9f0` rather than copied.** CI **23/23 green**. Locally on `[all]`: `pytest tests/test_probe_completeness_gate.py tests/test_brotli_framing_gate.py tests/test_detection.py tests/test_sfx.py` → **115 passed, 1 skipped** (was 112/1 — the three new regression tests). `ruff check` and `ruff format --check` clean. Pyrefly/ty reported 9 diagnostics in my worktree, all `missing-import` for optional extras (`brotli`, `pycdlib`, `tqdm`, …) and none in files this PR touches — an artifact of my worktree not seeing the venv's site-packages, and CI's own Lint & type-check job is green.

### New findings

Three, none blocking. Full detail inline.

- **`N1` 🟡** — the "30 fabricated (0.020%)" figure in the threat model, open-issues and the spec was measured at the old 256-byte drain, which `F4` raised 256×. On my `/usr` corpus the surviving residual moved 24 → 21 between the two HEADs, so the published number now *understates* the mitigation. Worth a re-census or an "as measured at" annotation before merge.
- **`N2` 🟢** — the drain loop's `while…else` is correct (I probed the boundary: complete streams accepted at 65535/65536/65537, truncations rejected up to and including 65536), but the adjacent comment describes only the accepting half, and the boundary `read(1)` that makes the exact case work goes unexplained.
- **`N3` 🟢** — the drained bytes are accumulated and joined on every fully-visible probe but consumed only by `require_output`; the `produced` counter already suffices.

### Verdict: ✅ Approve

The blocking finding is fixed and covered by a test, the two settled design decisions landed in code *and* spec together, and the cleanup went further than asked. `N1` is the only item I'd fold in before merge, and it is a documentation number rather than behaviour.

One observation worth keeping: `F1` and `N1` were both invisible to a green CI — one because no test exercised the mid-positioned contract, the other because a measured figure outlived the code it measured. The new mid-positioned test closes the first; the second is the kind of thing only a re-run catches.

---

PR #273, opened 2026-08-31
detection-prefix-workspace: one-pass prefix buffer, budget, and receipt

davitf:
## Summary

Implements OpenSpec change `detection-prefix-workspace`: one-pass detection plumbing via `PrefixWorkspace`, detection budget/receipt (kept out of root `__all__` until a deliberate export pass), SFX hits as candidate origins, and related tests/docs.

## Review follow-up (complete)

All review decisions applied:

1. **1B** — trim `detection-cost` to what ships; delete unused `read_tail`
2. **2 hybrid** — seek on cheap random-access `read_at`; capped buffer otherwise
3. **3A** — keep `detection_cost` out of package-root `__all__`

Independent findings F2/F4/F6–F8/F10–F13/F15 fixed earlier; F14 deferred to `detection-result-surface`.

## Test plan

- [x] `./scripts/check.sh --fix`
- [x] `./scripts/test.sh` (2610 passed)
- [ ] CI on `2b50e99`

<div><a href="https://cursor.com/agents/bc-786a752d-e913-48fc-b0be-83e1a1b47b29?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-786a752d-e913-48fc-b0be-83e1a1b47b29&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+        """
+        if candidate_origin < 0:
+            raise ValueError("candidate_origin must be non-negative")
+
+        def peek_more(length: int) -> bytes:
+            return self.peek_range(candidate_origin, length)
+
+        return peek_more
+
+    def read_at(self, offset: int, length: int) -> bytes | None:
```
davitf: ### F1 · 🔴 `[blocking]` · CONFIRMED — `read_at` buffers everything up to the offset instead of seeking

`read_at(offset, length)` is implemented as `peek_range(offset, length)` → `ensure(offset + length)`, which grows `_buf` through the **entire** range. On `main`, `_make_probe_read_at` seeked to `offset` and read `length` bytes. The Brotli chain walk (`brotli_framing.chain_proves_invalid`) drives `read_at` with offsets taken from declared meta-block lengths, so the buffered prefix now tracks how far the chain walks — bounded only by `source_length`.

**Measured, benign input** — a 48 MiB Brotli stream of incompressible data, no extension:

| | detected | bytes pulled from the file |
|---|---|---|
| `main` | `BROTLI` / `GUESS` | 172 032 (0.2 MiB) |
| this PR | `BROTLI` / `GUESS` | 29 360 180 (**28.0 MiB**) |

Same answer, 170× the I/O — and all 28 MiB is resident in `_buf`. Direct measurement of the mechanism on a 128 MiB file:

```python
with PrefixWorkspace(path_to_128MiB_file, BALANCED_BUDGET) as ws:
    ws.read_at(SIZE - 4, 4)     # returns 4 bytes
    ws.buffered_length          # 134_217_728
    ws.receipt.seeks            # 0
# peak RSS delta: +384 MiB
```

**Why it matters.** This is the VISION §1.5 budget measured in the currency the project says to measure it in (bytes and seeks), and it undercuts the synced spec's "no unbounded buffering" clause: the non-seekable side is capped at `PROBE_READ_AT_MAX_OFFSET_NONSEEKABLE` (1 MiB), the seekable side is capped at nothing. It also makes `dev-docs/threat-model.md` O11's "Memory is not the problem … time is" stale in the same PR that wrote it.

**Fix direction.** Give `read_at` a genuine random-access path: when `_random_access_handle()` is available and `SEEK` is in `capabilities()`, seek + read `length` bytes without touching `_buf`, charge a seek and `far_bytes`, then restore the handle to `len(_buf)` on the way out (the same restore `read_tail` already does). Keep `peek_range`/`ensure` for forward growth only. The non-seekable branch keeps today's buffer-with-a-cap behaviour.

Red–green candidate: assert `ws.buffered_length` stays O(length) for a far `read_at` on a seekable source, and pin the brotli case's `unique_bytes_read` against file size.

---

src/archivey/internal/detection.py
```diff
@@ -598,126 +522,103 @@ def _detect_format_body(
...
-            far_fmt = _match_magic(far_data, far)
-            if far_fmt is not None:
-                _warn_on_conflict(
-                    collector, name, ext_match, far_fmt, _ConflictEvidence.MAGIC
+            return _attach_receipt(info, workspace)
+
+        # 2. Self-extracting archives.
+        cue = executable_cue(data)
+        if cue is not ExecutableCue.NONE:
+            scan_limit = min(SFX_MAX, budget.max_scan_bytes)
```
davitf: ### F2 · 🔴 `[blocking]` · CONFIRMED — `max_scan_bytes` never bounds the SFX scan

`scan_limit = min(SFX_MAX, budget.max_scan_bytes)` is computed here, used only as an on/off test, and then discarded — `_scan_for_sfx_payload` hardcodes `find_magic_in_prefix(peek_more, needles, limit=SFX_MAX)` at line 447.

**Measured** — MZ stub + 3 MiB of junk containing no archive magic:

```
BALANCED (max_scan_bytes=2 097 152): source bytes read = 2 171 089
FAST     (max_scan_bytes=  262 144): source bytes read = 2 171 089
```

`FAST` is byte-for-byte identical to `BALANCED`. The synced spec's preset table says `FAST` = "smaller tail/scan/decode budgets", and `tasks.md` 5.1 ("`BALANCED`, `FAST`, `THOROUGH` as specified") is marked `[x]`.

**Fix direction.** Pass `scan_limit` into `_scan_for_sfx_payload` and on to `find_magic_in_prefix(..., limit=scan_limit)`. When `scan_limit < SFX_MAX`, record `record_skip("sfx_scan", TierSkipReason.BUDGET_EXHAUSTED)` so a truncated scan reports the incomplete search the spec promises for `FAST`.

Red–green candidate: same fixture, assert `FAST` reads strictly fewer bytes than `BALANCED`.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+        self._receipt.scanned_bytes += nbytes
+
+    def charge_decode(self, *, input_bytes: int = 0, output_bytes: int = 0) -> None:
+        self._receipt.decode_input += input_bytes
+        self._receipt.decode_output += output_bytes
+
+    def record_skip(self, tier: str, reason: TierSkipReason) -> None:
+        self._receipt.record_skip(tier, reason)
+
+    def read_tail(self, nbytes: int) -> bytes | None:
```
davitf: ### F3 · 🔴 `[blocking]` · CONFIRMED — the synced spec asserts behaviour no code implements

`openspec/specs/detection-cost/spec.md` lands as a **main** spec (current contract), and the change is archived with every task `[x]`. Three groups of divergence:

**(a) Two scenario rows fail.** From the spec's own "preset boundaries" table:

| Spec scenario | Spec says | Actual |
|---|---|---|
| ZIP behind a non-cueing prefix (JPEG + appended ZIP), `THOROUGH` | Found via the tail tier | `FormatDetectionError` |
| `zipapp` (`#!` prefix + ZIP), `BALANCED` | Found via the cued scan | `FormatDetectionError` |

The first fails because **`read_tail` (this method) is never called** — no call site anywhere in `src/` or `tests/`. `tail_bytes` can therefore never be nonzero, and the receipt scenario "Tail tier runs | `tail_bytes` and one `seek` charged" is likewise unreachable. The second fails because `executable_cue` recognises `MZ`/`ELF` but not a shebang.

**(b) Four budget fields are read by no code.** `completion_window_bytes`, `max_index_bytes`, `max_probe_links`, `collect_nonmaximal_candidates` appear only in `detection_cost.py`'s preset literals and in test fixtures. The Brotli walk uses its own module constant `CHAIN_MAX_LINKS`, not `budget.max_probe_links`. `THOROUGH`'s `completion_window_bytes = 1 << 62` is what the spec's "Genuine 800 KB Brotli stream … `THOROUGH` → `CERTAIN`" row depends on; nothing reads it. `max_prefix_bytes` is honoured only for the first peek (`detection.py:537`); every later tier grows past it freely.

**(c) The supporting machinery is dead and carries a latent bug.** `_fetch_forward` repositions the handle before a delta read in the `path` and `seekable` branches but **not** in the spool branch — so a `read_tail` followed by an `ensure()` on a spooled source would splice bytes from the wrong offset. Unreachable today only because nothing calls `read_tail`.

**Fix direction** — this is the pause-and-ask item; see decision 1 in the review body. Either implement the tail tier and the four knobs here, or trim the synced spec (and `tasks.md` 5.1) to what actually ships and move the rest into the `detection-evidence-ledger` / `detection-result-surface` deltas already present in this PR, deleting `read_tail` and its spool/`_random_access_handle` support until a caller exists. What should not stand is a main spec describing tiers that return `FormatDetectionError`.

---

src/archivey/internal/detection.py
```diff
@@ -535,32 +431,48 @@ def _warn_on_conflict(
...
     """
-    by_needle = {entry.magic: entry.format for entry in entries}
-    hit = find_magic_in_prefix(peek_more, tuple(by_needle), limit=SFX_MAX)
+    by_needle = {entry.magic: entry for entry in entries}
+    needles = tuple(ScanNeedle(entry.magic, entry.offset) for entry in entries)
+    hit = find_magic_in_prefix(peek_more, needles, limit=SFX_MAX)
     if hit is None:
         return None
-    offset, needle = hit
+    workspace.charge_scanned(min(workspace.buffered_length, SFX_MAX))
```
davitf: ### F4 · 🟡 `[important]` · CONFIRMED — `scanned_bytes` is charged only when the scan succeeds

```python
hit = find_magic_in_prefix(peek_more, needles, limit=SFX_MAX)
if hit is None:
    return None                                            # ← expensive path exits here
workspace.charge_scanned(min(workspace.buffered_length, SFX_MAX))
```

The charge sits after the early return, so the cheap case is billed and the expensive one is free.

**Measured:** a full 2 MiB scan that finds nothing → `scanned_bytes = 0`; a hit at offset 1 128 → `scanned_bytes = 1 241`. Exactly inverted. `within_budget`'s `scanned_bytes <= max_scan_bytes` check can therefore never trip on the case it exists for.

**Fix direction.** Charge before the `if hit is None` return — and, after F2, charge against `scan_limit` rather than `SFX_MAX`.

---

tests/atheris_fuzz/targets.py
```diff
@@ -95,10 +95,16 @@ def sevenzip_open_one(data: bytes) -> None:
...
+
     try:
-        detect_format(io.BytesIO(data))
+        info = detect_format(io.BytesIO(data), budget=BALANCED_BUDGET)
     except ArchiveyError:
         return
+    # Aggregate cost must stay inside the declared budget — pins the invariant whether
+    # limits are later resolved as per-detection or per-candidate.
+    if info.cost_receipt is not None:
+        assert info.cost_receipt.within_budget(BALANCED_BUDGET), info.cost_receipt
```
davitf: ### F5 · 🟡 `[important]` · CONFIRMED — this asserts a budget invariant that ordinary inputs violate

**Measured** `within_budget(BALANCED_BUDGET) == False` for two inputs that detect successfully:

- MZ stub + 3 MiB junk named `stub.zip` → `ZIP` by extension, `unique_bytes_read = 2 171 089` vs an allowance of 2 097 152.
- The 48 MiB Brotli file from F1 → `BROTLI`, `unique_bytes_read = 29 360 180`.

The target feeds `io.BytesIO(data)` with no name, so tripping this in CI needs a >2 MiB mutator input — unlikely today, which is why the suite is green. But `dev-docs/threat-model.md` O11 records this assertion as the standing bound ("ships … a fuzz assertion that aggregate detection cost stays inside the declared budget"), and `tasks.md` 7.2 marks it done. The invariant is false at the API level regardless of whether the fuzzer reaches it, so the recorded mitigation is weaker than the threat-model entry claims.

**Fix direction.** F1 and F2 remove both measured violations. Afterwards add a deterministic (non-fuzz) test asserting `within_budget` on these two fixtures, so the invariant is pinned by something other than mutator luck.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+            if not chunk:
+                break
+            spool.write(chunk)
+            remaining -= len(chunk)
+            self._receipt.spooled_bytes += len(chunk)
+            self._receipt.unique_bytes_read += len(chunk)
+        if remaining == 0:
+            # Source may still have more — abandon spooling; tiers needing TAIL are
+            # unavailable. Leave the already-spooled prefix usable as a forward buffer.
+            extra = source.read(1)
```
davitf: ### F6 · 🟡 `[important]` · CONFIRMED — spool abandonment loses a byte and then misreports the source length

`extra = source.read(1)` tests for more data and **discards the byte** — on a non-seekable source it is gone. The branch then sets `_raw_forward = None` (line 379), so `_fetch_forward` returns `b""` forever and every later `ensure()` sets `_source_exhausted`. `remaining_known()` consequently returns `len(self._buf)` — the spool limit — and `_detect_format_body` hands that to content probes as `source_length`, which `brotli_framing.chain_proves_invalid` uses in **equality** tests (`nxt != source_length`). So the failure mode is a wrong answer, not merely a missed one.

**Measured** — 10 240-byte pipe, `spool_non_seekable_up_to=100`:

```
prefix returned : 100 bytes  (source is 10 240)
byte 100        : missing — got[96:104] == 60616263, expected 6061626364656667
remaining_known : 100        (reported as provable)
```

This contradicts the synced spec's "Near and far prefix evidence and cued forward scans SHALL still work through replay buffering" for the pipe case. No test covers this branch — `test_spool_policy_grants_tail_to_a_pipe` only exercises a source that fits inside the bound.

**Fix direction.** Keep the probe byte (`self._buf.extend(extra)`, or push it back), and keep `_raw_forward = source` so prefix tiers can carry on reading forward past the spool bound. Leave `_source_exhausted` alone and let `remaining_known()` return `None` — unknown, not `100`. Add a test for a source larger than the bound asserting both byte-exactness of the prefix and `remaining_known() is None`.

This path is opt-in and no preset enables it (all three set `spool_non_seekable_up_to=0`), so nothing in-tree is affected today — see decision 4 in the review body.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+        self._closed = True
+        try:
+            if self._seekable_stream is not None and self._entry_pos is not None:
+                self._seekable_stream.seek(self._entry_pos)
+        finally:
+            if self._owned_path and self._path_handle is not None:
+                self._path_handle.close()
+                self._path_handle = None
+            if self._spool is not None and self._spool_abandoned:
+                self._spool.close()
```
davitf: ### F7 · 🟡 `[important]` · CONFIRMED — the spool temp file is never closed

The guard is `if self._spool is not None and self._spool_abandoned`, but `_begin_spool` sets `self._spool = None` in exactly the abandoned branch (line 376). The condition is therefore never true, and the *successful* spool case leaks a `SpooledTemporaryFile` — an on-disk file once it exceeds `max_size` — until GC finalises it.

**Measured:** `sp.closed is False` after `ws.close()` on a workspace whose spool succeeded.

**Fix direction.** Close `_spool` unconditionally in `close()` and drop the `_spool_abandoned` guard.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+            caps.add(DetectionCapability.SIZE_KNOWN)
+
+        can_seek = self._kind in ("path", "seekable") or (
+            self._spool is not None and not self._spool_abandoned
+        )
+        if can_seek and self._budget.max_seeks > 0:
+            caps.add(DetectionCapability.SEEK)
+            if self._budget.max_tail_bytes > 0:
+                caps.add(DetectionCapability.TAIL)
+        elif (
```
davitf: ### F8 · 🟡 `[important]` · CONFIRMED — capability signals disagree with what the workspace will do

Two directions, both in `capabilities()`:

- **`TAIL` advertised when it cannot be used.** This `elif` grants `TAIL` when a spool exists and `max_tail_bytes > 0` **without** `SEEK`. It is reachable only when `max_seeks == 0` — and in that state `read_tail` immediately returns `None` with `BUDGET_EXHAUSTED`. Measured: `TAIL in caps == True`, `SEEK in caps == False`, `read_tail(100) is None`.
- **`SIZE_KNOWN` withheld when the size is known.** It is granted only for `_kind in ("path", "seekable")`. A fully-spooled source has an exact known size (`_total_size` is set at line 385) and is denied the capability. Measured: `_total_size == 4096`, `SIZE_KNOWN == False`.

The synced spec is explicit that capabilities are "evaluated against source **and** budget", and that `SIZE_KNOWN` means "a cheap total size is available". Advertising a capability the budget forbids is the exact inversion of the `max_seeks = 0` row the method does get right.

**Fix direction.** Delete this `elif` (spool already sets `can_seek`), and replace the three-way `_kind` comparison for `SIZE_KNOWN` with `self._total_size is not None and self._entry_pos is not None`.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+
+        Extends the prefix buffer when the range lies in the forward-growing region.
+        Does not re-fetch bytes already buffered.
+        """
+        if origin < 0 or length < 0:
+            raise ValueError("origin and length must be non-negative")
+        if length == 0:
+            return b""
+        end = origin + length
+        self._receipt.prefix_bytes += length
```
davitf: ### F9 · 🟡 `[important]` · CONFIRMED — `prefix_bytes` is wrong in both directions

`peek_range` charges the **requested** span, so overlapping growing peeks each bill their full length, while `read_at`'s buffering side-effect lands only in `unique_bytes_read`:

| case | `prefix_bytes` | `unique_bytes_read` |
|---|---|---|
| MZ + 3 MiB junk | 3 510 302 | 2 171 089 (over-reports 1.6×) |
| 48 MiB Brotli (F1) | 102 574 | 29 360 180 (under-reports 286×) |

The docstring ("Bytes requested from the workspace") is accurate, but the field shares a name with `max_prefix_bytes = 4096`, so a reader comparing the two draws the wrong conclusion either way.

**Fix direction.** Rename to `prefix_requests_bytes` — or drop it, since `unique_bytes_read` is the honest number — and make `read_at`'s buffering visible in whichever counter the budget actually checks.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,403 @@
...
+        if end <= len(self._buf) or self._source_exhausted:
+            return bytes(self._buf)
+        needed = end - len(self._buf)
+        chunk = self._fetch_forward(needed)
+        if chunk:
+            self._buf.extend(chunk)
+            self._receipt.unique_bytes_read += len(chunk)
+        if len(chunk) < needed:
+            self._source_exhausted = True
+        return bytes(self._buf)
```
davitf: ### F10 · 🟡 `[important]` · CONFIRMED — every peek copies the whole buffer

`ensure` returns `bytes(self._buf)` here (and at lines 106, 163; `peek_range` then slices it), materialising a full copy of the buffer on every call. `_BoundedPeekReader.read` (`detection.py:229-235`) re-invokes `peek_more(end)` on each growth, so an inner-TAR probe pulling 1 MiB in codec-sized chunks copies O(n²) bytes; the SFX scan's four geometric steps copy up to 2 MiB each. This is what produced the +384 MiB peak RSS for the 128 MiB buffer measured in F1.

**Fix direction.** Have `ensure` return `None` and let `peek_range` slice `self._buf` directly (`bytes(self._buf[origin:end])` — only the requested span is copied); expose the whole-buffer `buffer` property as a `memoryview`.

---

src/archivey/internal/detection.py
```diff
@@ -598,126 +522,103 @@ def _detect_format_body(
...
+
+        # 3. Far magic (ISO's CD001 at offset 32 769).
+        if far and budget.max_far_bytes > 0:
+            far_needed = max(e.offset + len(e.magic) for e in far)
+            far_needed = min(far_needed, budget.max_far_bytes)
+            if length is not None and length < far_needed:
+                pass  # size-gated: never pay the peek
+            else:
+                far_data = workspace.peek_prefix(far_needed)
+                workspace.charge_far(
```
davitf: ### F11 · 🟡 `[important]` · CONFIRMED — a short far-magic read is charged zero

```python
workspace.charge_far(len(far_data) if len(far_data) >= far_needed else 0)
```

When the far window comes up short the read still happened, but `far_bytes` records nothing. Reachable when `remaining_known()` is `None` (unknown-size seekable stream) and the source ends inside the window — the size gate above only skips the peek when the length is *provable*.

**Fix direction.** `workspace.charge_far(len(far_data))` unconditionally.

---

src/archivey/internal/detection.py
```diff
@@ -737,15 +639,28 @@ def peek_more(length: int) -> bytes:
...
-    )
+        raise FormatDetectionError(
+            "Could not detect archive format: no magic-byte match and no usable file extension.",
+            archive_name=name,
+        )
+
+
+# Back-compat aliases used by older tests that monkeypatched the private helpers.
+# They are thin wrappers over a throwaway workspace so size-gate spies still work.
+def _peek_prefix(source: str | Path | BinaryIO, length: int) -> bytes:
```
davitf: ### F12 · 🟡 `[important]` — test-only shims left in `src/`

`_peek_prefix` is kept "so existing tests that spy on prefix peeks keep working during the transition", and line 116 re-exports `_PROBE_READ_AT_MAX_OFFSET_NONSEEKABLE` "for tests that previously imported the private name". Meanwhile `tasks.md` 2.5 — "Retire `_peek_prefix`, the `peek_more` closures and `_make_probe_read_at` in favour of workspace views" — is marked `[x]`.

Per CONTRIBUTING's clean-as-you-go rule, either update the tests to the workspace API and delete both, or record the deferral explicitly (PR note / `IDEAS.md` / `review/backlog.md`). An unspoken transition shim carried in production code is the debt shape the project aims to avoid — and "the transition" has no recorded end date.

---

src/archivey/internal/sfx.py
```diff
@@ -30,7 +30,9 @@
 
 import struct
 from enum import Enum
-from typing import BinaryIO, Callable, Sequence
+from typing import BinaryIO, Callable, NamedTuple, Sequence
+
+from archivey.internal.detection_workspace import candidate_origin_for_hit
```
davitf: ### F13 · 🟡 `[important]` — layering inversion: backends now import the detection workspace

`sfx.py` is consumed by `internal/backends/rar_parser.py` and `internal/backends/sevenzip_parser.py`, so opening a RAR or 7z now transitively imports `detection_workspace` and, through it, `tempfile`, `PeekableStream`, and `detection_cost` — all for a two-line helper:

```python
def candidate_origin_for_hit(hit_offset: int, needle_offset: int) -> int | None:
    origin = hit_offset - needle_offset
    return origin if origin >= 0 else None
```

Nothing about it concerns workspaces, and its only two callers (`scan_for_magic`, `find_magic_in_prefix`) both live in this file.

**Fix direction.** Move `candidate_origin_for_hit` into `sfx.py` beside `ScanNeedle`/`MagicHit`, and drop the import. `detection.py` already imports both modules, so nothing else needs to change.

---

src/archivey/internal/detection.py
```diff
@@ -737,15 +639,28 @@ def peek_more(length: int) -> bytes:
...
 
-    # 5. Extension-only guess.
-    if ext_fmt is not None:
-        return FormatInfo(ext_fmt, DetectionConfidence.GUESS, "extension")
+        # 5. Extension-only guess.
+        if ext_fmt is not None:
+            return _attach_receipt(
+                FormatInfo(ext_fmt, DetectionConfidence.GUESS, "extension"),
+                workspace,
+            )
```
davitf: ### F14 · 🟢 `[nit]` — no receipt on the failure path

`FormatDetectionError` is raised without `_attach_receipt`. This is often the *most* expensive detection — a full SFX scan miss, plus the far peek, plus every content probe (F2's fixture reads 2 171 089 bytes before landing here) — and it is the one case where a caller gets no cost information at all.

**Fix direction.** Attach the receipt and `unavailable_tiers` to the exception, so a caller can tell an expensive miss from a cheap one.

---

src/archivey/internal/detection.py
```diff
@@ -598,126 +522,103 @@ def _detect_format_body(
...
+        if budget.max_tail_bytes <= 0:
+            workspace.record_skip("zip_tail", TierSkipReason.NOT_ENABLED_BY_POLICY)
+        elif DetectionCapability.TAIL not in workspace.capabilities():
+            workspace.record_skip("zip_tail", TierSkipReason.CAPABILITY_UNAVAILABLE)
+
+        # Magic signals split by where they live: "near" ones fit in the default window;
+        # "far" ones (ISO's CD001 at 32 769) need an extended peek taken on demand.
+        near = [e for e in magic_entries if e.offset + len(e.magic) <= DETECTION_LIMIT]
+        far = [e for e in magic_entries if e.offset + len(e.magic) > DETECTION_LIMIT]
+        near_needed = min(
```
davitf: ### F15 · 🟢 `[nit]` — a "near" magic beyond `max_prefix_bytes` is silently unsearchable

`near` is partitioned by `DETECTION_LIMIT` (line 535), but `near_needed` is then clamped by `budget.max_prefix_bytes`. The two are independent constants that happen both to be 4 096 today. If they ever diverge, a magic classified *near* but sitting past the budget is missed with no `record_skip` — a silently incomplete search, which is exactly the distinction `TierSkipReason` exists to make visible.

**Fix direction.** When the clamp bites, record `record_skip("near_magic", TierSkipReason.BUDGET_EXHAUSTED)` — or assert the invariant that `max_prefix_bytes >= DETECTION_LIMIT`.

---

src/archivey/__init__.py
```diff
@@ -54,6 +54,15 @@
     ListingCost,
     StreamCapability,
 )
+from archivey.detection_cost import (
```
davitf: ### F16 · 💡 `[suggestion]` — the newly frozen public surface is uneven

Seven names go into `__all__` here, which makes them freeze surface. Three observations for whoever finalises it:

- **`DetectionBudget` has 12 required fields and no defaults**, so every custom budget must spell out all 12 — visible in `tests/test_detection_workspace.py:240-275`, where two fixtures do exactly that to change one field each. Defaults (or a `BALANCED_BUDGET`-derived factory) would make `replace()` the ergonomic path that `THOROUGH`'s own `# still opt-in via replace()` comment already assumes.
- **`DetectionCapability` and `DetectionCostReceipt` have no public producer.** `FormatInfo.cost_receipt` is documented in `detection.py` as "Public exposure on `FormatInfo` is `detection-result-surface`", and `capabilities()` is workspace-internal. Both types are exported but unreachable through the public API.
- **`TierSkip` / `TierSkipReason` are *not* exported**, yet they are the element type of `FormatInfo.unavailable_tiers`, which `tests/test_detection_workspace.py` already reads. A caller cannot name the type of a field they are given.

Combined with F3(b) — 4 of the 12 budget fields being inert — see decision 3 in the review body on whether this surface should freeze in this PR at all.

---

davitf (commented):
## Review: `detection-prefix-workspace` — 🔄 Request Changes

*(Posted as `COMMENT` because GitHub does not allow a formal Request-Changes review on your own PR — the verdict is 🔄, and F1/F2/F3 are the blockers.)*

Reviewed at head `35bdd71`. Findings F1–F16 are inline on the relevant lines; this body carries the briefing and the decisions that need a human call.

---

### 1. Maintainer briefing

**What this change is.** It rewires format detection so every front-of-source read goes through one new `PrefixWorkspace` (`internal/detection_workspace.py`) that owns the handle and grows a single monotonic prefix buffer, replacing the old per-tier reopen/rewind helpers (`_peek_prefix`, `_make_probe_read_at`). Alongside it lands a new public module `archivey/detection_cost.py` (`DetectionBudget`, `DetectionCostReceipt`, `DetectionCapability`, three presets), SFX hits reported as candidate origins rather than raw needle positions, and a synced main spec `openspec/specs/detection-cost/spec.md`. The stated goal — zero backward seeks, each byte fetched once — is met for the prefix tiers.

**Snapshot.** Medium-large: ~1 400 lines of `src/` + tests across 26 files (the rest is OpenSpec/docs).

**Gates I ran myself** on this head, in a worktree (not copied from CI):

| gate | result |
|---|---|
| `ruff check src tests` | All checks passed |
| `ruff format --check src tests` | 184 files already formatted |
| `pyrefly check` | 0 errors (11 warnings not shown) |
| `ty check` | All checks passed |
| `pytest tests/test_detection.py tests/test_detection_workspace.py tests/test_sfx.py tests/test_brotli_framing_gate.py` | **137 passed** |

I did **not** run `./scripts/test.sh --all-configs`; the PR reports it green and CI is green on all 23 checks. Every measurement quoted in the findings below is from a script I ran against this branch, with the `main` comparison run against `056c429`.

**Verdict: 🔄 Request Changes.**

**Main points.**

- **F1 🔴** — The seek-elimination was bought with unbounded prefix buffering. `read_at` now buffers `[0, offset)` instead of seeking. Detecting a 48 MiB Brotli file reads **28 MiB** (main: 0.2 MiB) — a 170× regression that scales with file size, on a benign input.
- **F2 🔴** — `budget.max_scan_bytes` never reaches the SFX scan; `scan_limit` is computed and dropped. `FAST` reads exactly as much as `BALANCED` (2 171 089 bytes, measured).
- **F3 🔴** — `openspec/specs/detection-cost/spec.md` is synced as current contract, but two of its own scenarios fail and 4 of 12 budget fields are read by no code. Nothing ever calls `read_tail`, so the tail tier the spec attributes to `THOROUGH` does not exist.
- **F4 🟡** — The receipt is not honest: a 2 MiB SFX **miss** charges `scanned_bytes = 0` while a fast hit charges 1 241.
- **F5 🟡** — `within_budget()` is False for ordinary inputs, and the fuzz target asserts it — with `dev-docs/threat-model.md` O11 citing that assertion as the mitigation.
- **F6/F7 🟡** — The opt-in spool path drops a byte, truncates the source, reports the truncated length to content probes as *provable*, and leaks its temp file. Untested.

**What's fine.** The instrumented-source test with the recorded before/after table (`tests/test_detection_workspace.py`) is the right way to make an access shape normative, and the error-path restore test is a good catch. The candidate-origin work (`ScanNeedle`/`MagicHit`, negative-origin discard, the decoy-resume loops in both scan functions) is correct and well-tested. Docstrings throughout explain *why*, not *what* — the zstd-framing and inner-TAR probe rationales in particular.

---

### 3. Maintainer decisions

**1. The archived spec claims tiers this change does not ship — trim the spec, or implement them here?** (F3)

`openspec/specs/detection-cost/spec.md` is now the current contract, yet two of its scenario rows return `FormatDetectionError`; `read_tail` has no caller; four budget fields are inert. `tasks.md` 5.1 and 7.2 are marked done.

- **A** — implement the tail tier + the four knobs in this PR. Grows the change materially and contradicts its own "no new detection tiers".
- **B** — trim the synced spec and `tasks.md` to what ships, and move the tail tier, completion window, probe-link cap and non-maximal collection into the `detection-evidence-ledger` / `detection-result-surface` deltas already present in this PR. Delete `read_tail` and its spool/`_random_access_handle` support until a caller exists.

*Recommendation: B.* It keeps the change honest about being plumbing, and it removes the latent spool-repositioning bug in `_fetch_forward` along with the dead code.

**2. Seeks or bytes — which currency does the access-shape contract actually mean to fix?** (F1)

The headline "zero backward seeks, each byte fetched once" is real for the prefix tiers, but `read_at` bought it by replacing a seek with buffering `[0, offset)`. On a local file that is a bad trade (170× the bytes for a 48 MiB Brotli); on a hypothetical HTTP range reader it is the right one — which is exactly the open question this PR files in `dev-docs/IDEAS.md` ("Pricing a detection source in round trips rather than bytes").

- **A** — `read_at` seeks when `SEEK` is available, buffers only when it is not. Restores main's cost, keeps the prefix tiers' win.
- **B** — keep buffering, but cap it for seekable sources too and record `BUDGET_EXHAUSTED` past the cap.

*Recommendation: A.* The seek-count win the tests pin is about the *prefix* tiers, and A leaves that intact; B leaves detection unable to answer for large files it can trivially seek in.

**3. Is `detection_cost` ready to be frozen into `__all__` now?** (F16, F3b)

Seven names go public here, including a 12-field budget of which 5 fields have no effect, and 2 types that no public API returns. Once exported these are freeze surface.

- **A** — keep the module internal until `detection-result-surface` lands the public `FormatInfo` exposure; export nothing yet.
- **B** — export now, but only the names with working semantics (`DetectionBudgetPreset` + the three presets), and add defaults to `DetectionBudget`.

*Recommendation: A*, given that decision 1B would change the field set anyway.

**4. Spool policy: fix it here, or disable it?** (F6, F7)

The path is opt-in and no preset enables it, so nothing in-tree is affected today — but a caller who opts in gets a discarded byte, a truncated prefix, a `remaining_known()` that reports the truncated length as *provable*, and a leaked temp file, with no test coverage. Fixing it here is small (a few lines plus a test), or `spool_non_seekable_up_to > 0` could raise `ArchiveyUsageError` until the abandonment path is correct. Either is defensible; shipping it silently as-is is the one option I would rule out.

---

---

src/archivey/detection_cost.py
```diff
@@ -0,0 +1,273 @@
...
+            tail_bytes=self.tail_bytes + tail_bytes,
+            scanned_bytes=self.scanned_bytes + scanned_bytes,
+            seeks=self.seeks + seeks,
+            decode_input=self.decode_input + decode_input,
+            decode_output=self.decode_output + decode_output,
+            index_bytes=self.index_bytes + index_bytes,
+            spooled_bytes=self.spooled_bytes + spooled_bytes,
+        )
+
+    def within_budget(self, budget: DetectionBudget) -> bool:
```
davitf: ### F18 · 🟡 `[important]` · CONFIRMED — `within_budget()` is still false for an ordinary detection

New in this round; this is F5's root cause surviving in a different form after the `395bd52` rework.

```python
self.unique_bytes_read <= max(max_prefix_bytes, max_far_bytes, max_scan_bytes)
                          + max_tail_bytes + spool_non_seekable_up_to
```

Seek-based `read_at` charges `unique_bytes_read` — correctly, those bytes *were* fetched — but the allowance has no term for probe reads. Any detection that runs the SFX scan to its cap and then does a probe seek exceeds the allowance by exactly the probe bytes.

**Measured** — MZ stub + 3 MiB junk named `stub.zip`, plain `detect_format(path)`:

| preset | `unique_bytes_read` | allowance | over | `within_budget` |
|---|---|---|---|---|
| `BALANCED` | 2 097 176 | 2 097 152 | **+24** | `False` |
| `FAST` | 262 168 | 262 144 | **+24** | `False` |

Traced to one call: `_read_at_via_seek(offset=2 171 065, length=24)`.

The overshoot is **bounded**: `brotli_framing.CHAIN_MAX_LINKS = 8` × `CHAIN_HEADER_READ = 24` → at most 192 probe bytes, independent of file size. So this is a small, fixed accounting gap rather than an open-ended one.

Reachability through `tests/atheris_fuzz/targets.py:107` is still gated — `io.BytesIO` with no name raises `FormatDetectionError` on this input, so the fuzz assert does not fire. But `within_budget` is a public method on a public receipt type returning `False` for a plain `detect_format(path)` call, and `dev-docs/threat-model.md` O11 still cites this assertion as the standing bound.

**Fix direction.** Add a small probe allowance to the comparison — a `max_probe_bytes` field, or simply `max_probe_links * CHAIN_HEADER_READ`, which would give the reserved `max_probe_links` its first real consumer. Then add the deterministic non-fuzz test F5 asked for, asserting `within_budget` on this fixture under both presets.

---

tests/test_detection_workspace.py
```diff
@@ -0,0 +1,465 @@
...
+
+Measured on ``main`` before this change (seekable stream through ``detect_format``):
+
+| source | reads | forward seeks | backward seeks |
+| --- | --- | --- | --- |
+| gzip | 5 | 0 | **5** — the same 30 bytes fetched five times |
+| ISO | 2 | 0 | **2** — 4 096 bytes, rewind, then 32 774 re-read from zero |
+| ZIP, TAR | 1 | 0 | 1 |
+
+The workspace makes the shape normative: zero backward seeks, at most one seek towards
```
davitf: ### F17 · 🟢 `[nit]` · CONFIRMED — "zero backward seeks" is no longer literally true after the `read_at` rework

This docstring states the shape as normative: *"zero backward seeks, at most one seek towards the end, and each source byte fetched at most once."* After `395bd52`, content-probe `read_at` seeks and restores on cheap random-access sources, so a Brotli chain walk on a **seekable stream** does ~8 backward seeks.

**Measured** (48 MiB Brotli, incompressible, handed in as a seekable stream): `forward_seeks=7 backward_seeks=8`, while the receipt reports `seeks = 0`.

**This is a documentation/telemetry nit, not an I/O problem** — I want to be clear about that, because my first read of it was harsher. The access pattern is bounded and non-redundant:

| file | seeks (fwd/back) | total read | distinct | re-fetched |
|---|---|---|---|---|
| 8 MiB | 2 / 3 | 65 561 | 65 561 | **0** |
| 24 MiB | 6 / 7 | 65 657 | 65 657 | **0** |
| 48 MiB | 7 / 8 | 65 704 | 65 704 | **0** |
| 96 MiB | 7 / 8 | 65 704 | 65 704 | **0** |
| 200–900 KiB | 1 / 2 | 65 537 | 65 537 | **0** |

Seeks saturate at 7/8 by 48 MiB and do not grow at 96 MiB — hard-capped by `CHAIN_MAX_LINKS = 8`. Total bytes read is flat at ~65 704 regardless of file size, and **no byte is fetched twice at any size**, so the "fetched at most once" half of the claim holds. The merged coverage is a contiguous `[0, 65536]` prefix plus 24-byte reads roughly 4 MiB apart — the minimum work for a sparse chain walk.

The cache cannot absorb these: it is a contiguous prefix `[0, N)`, so serving a read at 20 MiB from it means buffering 20 MiB — exactly the F1 regression this path was introduced to fix. Caching and seek-avoidance are in opposition here, and the current split picks the right side.

**Fix direction.** Two wording corrections, no code change:
- Scope this docstring's claim to the prefix tiers, e.g. *"zero backward seeks for the prefix tiers; bounded content-probe seeks restore the handle."*
- Add the probe-seek exemption to the spec's receipt table in `openspec/specs/detection-cost/spec.md`, which currently discloses only one (`seeks: int # restore-on-exit is not charged here`).

A dedicated `probe_seeks` counter is not worth it given the fixed cap of 8 — I withdraw that suggestion.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,395 @@
...
+    @property
+    def receipt(self) -> DetectionCostReceipt:
+        return self._receipt.freeze()
+
+    @property
+    def skips(self) -> tuple:
+        return tuple(self._receipt.skips)
+
+    @property
+    def buffer(self) -> memoryview:
```
davitf: ### F19 · 🟢 `[nit]` · CONFIRMED — the `buffer` memoryview freezes the buffer

New in this round, introduced by the F10 fix. An exported `memoryview` over a `bytearray` blocks resize, so holding the returned view makes any later growth raise:

```python
mv = ws.buffer
ws.peek_prefix(200)
# BufferError: Existing exports of data: object cannot be re-sized
```

Nothing in `src/` reads the property, and the one test that does (`tests/test_detection_workspace.py:363-364`) takes transient slices, so nothing breaks today.

The concern is that the docstring — *"Live view of the prefix buffer — no copy. Callers that need ownership copy."* — reads as an invitation to hold the view, which is precisely the thing that breaks. The next caller to keep it across an `ensure` gets a `BufferError` from a line that looks innocent.

**Fix direction.** A docstring sentence is enough: note that the view is valid only until the next `ensure` / `peek_range`, and that callers should slice or `.tobytes()` before the buffer grows. (Returning `memoryview(bytes(...))` would also work but gives back the copy F10 removed.)

---

src/archivey/internal/detection_workspace.py
```diff
@@ -0,0 +1,395 @@
...
+        return None
+
+    @staticmethod
+    def _seek_is_expensive(stream: BinaryIO) -> bool:
+        # Lazy import: archive_stream must not import the detection workspace.
+        from archivey.internal.streams.archive_stream import ArchiveStream
+
+        return isinstance(stream, ArchiveStream)
+
+    def _read_at_via_seek(self, handle: BinaryIO, offset: int, length: int) -> bytes:
```
davitf: ### F20 · 🟢 `[nit]` · PLAUSIBLE — `_read_at_via_seek` is not exception-safe for the spool

`handle.seek(restore)` runs only on the success path. `read_exact` returns short on EOF rather than raising, so this needs a genuine `OSError` from the handle mid-read — narrow, and opt-in-only in reach.

It matters for exactly one source kind. `_fetch_forward`'s `path` and `seekable` branches re-seek defensively (`if handle.tell() != expected: handle.seek(expected)`), and `close()` restores the caller's stream, so both self-heal. The **spool** branch still does a bare `read_exact(self._spool, nbytes)` with no reposition — so an interrupted probe read leaves the spool at the probe offset and the next `ensure()` splices bytes from the wrong place.

Worth noting that this makes the restore in `_read_at_via_seek` load-bearing for spool correctness. That is a fine design — it is what retired the latent reposition bug from the old `read_tail` — but it is currently unstated and unguarded.

**Fix direction.** Either wrap the read in `try/finally` so `restore` always runs, or give the spool branch of `_fetch_forward` the same `tell() != expected` guard the other two have. The second is slightly better: it makes the invariant uniform across all three branches rather than relying on a caller elsewhere to maintain it.

---

davitf (commented):
## Re-review at `2b50e99` — 💬 Comment (no blockers remain)

Previous round was `35bdd71`. Every F1–F16 outcome below was verified by re-running the original probes against this head, not read off the reply threads.

### Status of the previous round

| ID | Sev | Claimed | Verified |
|---|---|---|---|
| **F1** | 🔴 | Fixed `395bd52` | ✅ **Fixed.** `read_at(SIZE-4, 4)` on 128 MiB → `buffered_length 0`, `unique 4` (was 134 217 728). 48 MiB Brotli: **65 704** bytes (was 29 360 180; `main` was 172 032 — now better than `main`). |
| **F2** | 🔴 | Fixed `34ffb24` | ✅ **Fixed.** `FAST` unique/scanned 262 144 vs `BALANCED` 2 097 152. |
| **F3** | 🔴 | Fixed `4e7c1cf` (1B) | ✅ **Fixed.** `read_tail` + `_random_access_handle` gone; `THOROUGH` tail off (`0/0`); the two false scenario rows now read "Not found" *with the reason*; reserved fields named in spec and dataclass docstring; live fields enumerated. |
| **F4** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** Miss charges `scanned_bytes = 2 097 152` (was 0). |
| **F5** | 🟡 | Fixed `395bd52` | ⚠️ **Partly — see F18.** The Brotli violation is gone, but `within_budget()` is still `False` on the SFX-miss fixture, now by a bounded 24 bytes from a different cause. Replying on that thread. |
| **F6** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** 101 bytes retained, byte-exact against source, `remaining_known() is None`. |
| **F7** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** Spool closed after `close()`. |
| **F8** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** `SIZE_KNOWN` True on a full spool; `TAIL` absent when `max_seeks == 0`. |
| **F9** | 🟡 | Clarified `395bd52` | ✅ **Accepted as documented.** No rename, but the docstring states the semantics and points at `unique_bytes_read`. The discipline asked for was honest labeling — this qualifies. |
| **F10** | 🟡 | Fixed `34ffb24` | ✅ **Fixed** (in-place `ensure`, `memoryview` buffer) — introduced **F19**. |
| **F11** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** `far_bytes = 32 774`. |
| **F12** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** Both shims deleted, no callers left. |
| **F13** | 🟡 | Fixed `34ffb24` | ✅ **Fixed.** Helper moved into `sfx.py`; `sfx.py` imports no workspace. |
| **F14** | 🟢 | Deferred | ✅ **Deferral accepted.** Written home to `detection-result-surface`, thread left open. It does genuinely expand the exception surface. |
| **F15** | 🟢 | Fixed `34ffb24` | ✅ **Fixed.** `('near_magic', 'budget_exhausted')` recorded. |
| **F16** | 💡 | Fixed `2b50e99` (3A) | ✅ **Fixed.** Out of `archivey.__all__`, still importable as `archivey.detection_cost`. |

Eight new regression tests, one per fixed finding. That is the right shape.

---

### 1. Maintainer briefing

**What changed.** Four commits took the three decisions (1B trim, 2 hybrid, 3A internal) plus the independent findings. The substantive rework is `read_at`: it now seeks on cheap random-access sources (path, full spool, non-`ArchiveStream` seekable stream) and falls back to capped prefix buffering only for non-seekable or expensive-seek sources. `read_tail` and the reserved-tier machinery are deleted, the main spec is trimmed to what ships, and `detection_cost` leaves the package root.

**Snapshot.** +231/−2 lines of `src/` since `35bdd71`.

**Gates I re-ran at `2b50e99`** (not copied from CI): `ruff check` + `ruff format --check` clean · `pyrefly check` 0 errors · `ty check` clean · `pytest tests/test_detection.py tests/test_detection_workspace.py tests/test_sfx.py tests/test_brotli_framing_gate.py` → **145 passed** (was 137). I did not run `--all-configs`; the PR body reports `./scripts/test.sh` at 2610 passed.

**Verdict: 💬 Comment.** All three blockers are genuinely resolved. What remains is one 🟡 and three 🟢.

**Main points.**

- **F18 🟡** — `within_budget()` is still `False` for an ordinary `detect_format(path)`: seek-based `read_at` adds to `unique_bytes_read` with no matching allowance term. Bounded at ≤192 bytes, so a small fixed accounting gap.
- **F17 🟢** — the test module's "zero backward seeks" claim and the receipt's `seeks = 0` are both inaccurate for probe walks. Documentation only; see below.
- **F19 🟢** — `buffer`'s `memoryview` freezes the `bytearray`; the docstring invites holding it.
- **F20 🟢** — `_read_at_via_seek` isn't exception-safe, and the spool branch of `_fetch_forward` still lacks the reposition guard the other two have.

**A correction to my last round.** I flagged the probe seeks as "unbounded, uncounted." The seek count is hard-capped — `CHAIN_MAX_LINKS = 8`, `CHAIN_HEADER_READ = 24`, so ≤8 seek pairs and ≤192 probe bytes regardless of file size. Measured across 8/24/48/96 MiB, seeks saturate at 7 forward / 8 backward by 48 MiB and do not grow, total bytes read stays flat at ~65 704, and **zero bytes are re-fetched at any size**. The access pattern is bounded and non-redundant; only the two statements *describing* it are stale. F17 is 🟢 accordingly, and I withdraw the suggested `probe_seeks` counter.

**What's fine.** The `read_at` split is well-judged — `_seek_is_expensive` correctly singles out `ArchiveStream` (codecs that service a backward restore by re-decoding), and parking richer round-trip pricing in `IDEAS.md` is right. The spec rewrite is the standout: `# reserved: …` per field, a "Live budget fields today" paragraph, and preset rows that state *why* something is not found rather than quietly dropping the claim. The spool fix keeps the look-ahead byte **and** drops the false `REMAINING_KNOWN` — both halves.

---

### 3. Maintainer decisions

**1. Does `within_budget()` get a probe term, or does the assertion get narrowed?** (F18) Either add a small probe allowance — `max_probe_links * CHAIN_HEADER_READ` would give the reserved `max_probe_links` its first real consumer — or change the fuzz target and threat-model O11 to assert something narrower that is true today (e.g. `scanned_bytes <= max_scan_bytes` alone). What I would avoid is leaving a public predicate that returns `False` for ordinary use while O11 cites it as the standing bound. *Recommendation:* the probe term.

Nothing else needs a human call. F14's deferral and F9's docs-only resolution both look right, and F17/F19/F20 are wording-and-guard fixes with no design question in them.

---

tests/atheris_fuzz/targets.py
davitf: **Re-review at `2b50e99`: partly fixed — reopening as F18.**

The Brotli half is genuinely fixed: 48 MiB Brotli on a path is now `within_budget(BALANCED) == True` at 65 704 unique bytes. Confirmed.

But `within_budget()` is still `False` on the *other* fixture from this thread — MZ stub + 3 MiB junk named `stub.zip`, plain `detect_format(path)`:

| preset | `unique_bytes_read` | allowance | over | `within_budget` |
|---|---|---|---|---|
| `BALANCED` | 2 097 176 | 2 097 152 | **+24** | `False` |
| `FAST` | 262 168 | 262 144 | **+24** | `False` |

Different cause than before. F2 fixed the scan overshoot, so `scanned_bytes` is now exactly at the cap — but `395bd52`'s seek path charges probe reads to `unique_bytes_read`, and the allowance has no term for them. Traced to a single call: `_read_at_via_seek(offset=2 171 065, length=24)`.

The good news is that it is bounded: `CHAIN_MAX_LINKS = 8` × `CHAIN_HEADER_READ = 24` → at most 192 probe bytes, independent of file size. So it needs a small fixed allowance, not a redesign.

Fuzz reachability is unchanged — `io.BytesIO` with no name still raises `FormatDetectionError` on this input, so `targets.py:107` does not fire. But the predicate is public and returns `False` for ordinary use, and `dev-docs/threat-model.md` O11 still cites it as the standing bound.

Carried forward as **F18** on `detection_cost.py` with the fix direction; decision 1 in the re-review body covers the probe-term-vs-narrow-the-assertion choice.

---

PR #275, opened 2026-08-31
OpenSpec: register prefixed-archive-detection split and defaults

davitf:
Proposal-only — no library code. Registers the implementation split and the defaults settled before any of the four landing PRs, so `prefixed-archive-detection` is not reviewed as one 64-task change.

Review round 1 (`7a3d214`) and round 2 (`1a56c5a`) are closed. Maintainer comments on `1a56c5a` moved the spend cap onto `ArchiveyConfig.detection_budget` (`5dc613f`): not a `budget=` keyword (ambiguous with decompression; most callers never set one), and not an `exhaustive_prefix_scan` bool. `#273`'s `detect_format(..., budget=)` is removed when that field lands (task 3.1).

## Settled

- **Four PRs, in order:** cue+zipapp → ZIP tail (budget-gated) → `prefix_kind` + scan validation → makeself needles plus TAR/gzip `start_offset`.
- **Format-owned hooks, ZIP-named tail.** ZIP/7z/RAR/gzip validators and the ZIP tail locator live on the format module. Validators return `HitOutcome` (`NOT_THIS_FORMAT` / `VALID` / `DAMAGED`); this change treats `DAMAGED` as skip. `detected_by="zip_tail_probe"` stays; there is no locator registry. No parallel `DetectionDeclaration`.
- **Detection spend cap is `ArchiveyConfig.detection_budget`.** `None` → BALANCED. Not a keyword on `open_archive` / `detect_format`. Tail probe off under `BALANCED`. JPEG+appended ZIP is a `THOROUGH` case once `max_tail_bytes` is raised.
- **No `OTHER_FORMAT`.** `PrefixKind` is `NONE` / `EXECUTABLE` / `SCRIPT` / `UNKNOWN`. How the offset was found is `detected_by`.
- **No `exhaustive_prefix_scan` bool.** Exhaustive scan is a larger `max_scan_bytes` on that same field. `THOROUGH.max_scan_bytes` stays at `SFX_MAX`.
- **Makeself is not detection-only.** TAR/single-file `start_offset` ships in the same PR as the compressor needles (task 5.7). Uncompressed TAR-behind-stub waits for the ledger checksum validator.
- **Makeself-aware `SKIP`/`COMPRESS` locator** is parked in `dev-docs/IDEAS.md`.

## Follow-up for #274 (`archive-origin-reporting`)

Drop `OTHER_FORMAT` from its `ArchiveInfo` table. Keep `payload_offset is None` as the origin-not-established axis. Forced `format=ZIP` should not run the tail probe under `BALANCED`. (Already applied on that branch as of `fe45330`.)

`openspec validate --strict` is clean for `prefixed-archive-detection`, `detection-evidence-ledger`, and `detection-result-surface`.

<div><a href="https://cursor.com/agents/bc-a04e58da-145b-4a70-ac6e-69f9f91c765a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a04e58da-145b-4a70-ac6e-69f9f91c765a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -497,10 +503,11 @@ carries about itself:
...
 - **RAR 4**: the 7-byte marker block SHALL be followed by a parseable main header.
 - **Stream codecs** (shebang cue only): the magic SHALL include the compression-method
-  byte where the format has one, and the candidate SHALL decode a bounded prefix
-  successfully. A hit that decodes SHALL then be run through the existing inner-TAR probe,
-  so a script-wrapped gzipped tar resolves to `TAR_GZ` rather than `GZIP`. These needles
-  are the shortest in the set, so the decode is doing the work the magic cannot.
+  byte where the format has one. Identity is the cheap structural validator (gzip header
+  checks; bzip2 block-size digit plus first-block / EOS marker; xz flags CRC; and the
+  equivalents the ledger lists). A hit that identifies SHALL then be run through the
+  existing inner-TAR probe, so a script-wrapped gzipped tar resolves to `TAR_GZ` rather
```
davitf: ## F1 🔴 `[blocking]` · CONFIRMED — ZIP is missing from the scan-hit validator list, and Block 1 makes the unvalidated needle reachable on every `#!` file

**Location:** this requirement, `specs/format-detection/spec.md:491-528`.

The requirement opens with an unqualified *"A forward-scan hit SHALL be validated before it is reported, using evidence the candidate carries about itself"*, then enumerates **7z** (`StartHeaderCRC` + end-offset), **RAR 5** (main-header CRC32), **RAR 4** (parseable main header), and **stream codecs** (this bullet). ZIP is absent, and the scenario matrix at `:518-528` has no ZIP row. So the ZIP path either violates that SHALL or is silently exempt, and the spec does not say which.

That matters more here than anywhere else in the change, because ZIP's `PK\x03\x04` needle is the mechanism the whole of Block 1 rests on — `tasks.md:6-8` ("Cue + zipapp — `#!` / Mach-O cue, ZIP local-header needle already present"). It is also the weakest needle in the set by a wide margin: four bytes with no checksum, searched across a 2 MiB window, against 7z's six-plus-CRC. And the justification for widening the cue population by 26% (`proposal.md:38-41`, 742 more files per 2 868 on a `/usr` tree) is *"a hit is essentially never wrong, so it can be reported at high confidence"* (`proposal.md:57-63`) — a claim the validator list does not actually cover for ZIP.

**Trigger — reproduced on this HEAD (`2c8631e`, `src/` unchanged by this PR):**

```python
import io, sys; sys.path.insert(0, "src")
from archivey import detect_format, open_archive
stub = (b"\x7fELF" + b"\x00"*60 + b"# helper blob follows\n"
        + b"PK\x03\x04" + b"not a zip at all, just four bytes\n" + b"\x00"*4000)

detect_format(io.BytesIO(stub))
# FormatInfo(format=ArchiveFormat.ZIP, confidence=PROBABLE,
#            detected_by='sfx_scan', payload_offset=86, ...)

open_archive(io.BytesIO(stub))
# CorruptionError: Could not open ZIP archive: truncated or corrupt ZIP
#                  (central directory unreadable) (format=ZIP)
```

An ordinary non-archive is reported as a *damaged ZIP* rather than "not an archive" — a dishonest error, and the wrong side of VISION §1.4 (damaged input is first-class; a non-archive must not be labelled one).

Today this is confined to `MZ`/ELF-cued files, and the same bytes behind a shebang correctly raise `FormatDetectionError` — **I verified both.** Block 1 is precisely what removes that protection:

```python
sh = b"#!/bin/sh\n# blob\nPK\x03\x04nope\n" + b"\x00"*4000
detect_format(io.BytesIO(sh))   # today: FormatDetectionError  → after Block 1: ZIP claim
```

Realistic carriers among the newly-enrolled population: installer scripts with an embedded base64 or binary payload, scripts that `grep`/`printf` for ZIP magic, fixture generators, and any `#!` file with an appended blob that is not a ZIP.

**Fix direction.** Add a ZIP bullet to this requirement at the same bar the change now imposes on stream codecs ("cheap structural validator … without running a content probe"). A local file header carries plenty: `version_needed` ≤ a sane maximum, the reserved GP-flag bits clear, `compression_method` in the known set, and `filename_length` / `extra_field_length` plausible and in-bounds against the source. Stronger, if you want the `CERTAIN` that `format-zip` already claims for the tail probe: require the candidate to resolve to a real EOCD + central directory — which also merges cleanly with the Block 2 tail work. Add the matching scenario row (*"`#!` stub whose text contains `PK\x03\x04` but no valid local header → not reported; scan continues"*) and a red–green task alongside 4.2, which already covers exactly this shape for gzip.

Worth noting: `zip_reader.py:1456-1462` already reasons about needle *selectivity* — it keeps only the local header, dropping the EOCD and spanned markers because "as needles inside a 2 MiB stub window they would claim any executable that happens to contain those four bytes". This finding is that the same reasoning was never extended into a *validator*.

---

openspec/changes/prefixed-archive-detection/specs/archive-reading/spec.md
```diff
@@ -81,23 +7,17 @@ that much from every source a caller opens is not free. A caller who knows bette
...
-
-Because it arrives through `config=`, it SHALL follow the existing rules for configuration
-that a given call cannot act on: a source that is already a plain archive at offset 0 is
-matched at tier 1 and the flag simply never applies, which is not an error and SHALL NOT be
-reported as an unused argument.
+The opt-in SHALL be a `DetectionBudget` whose `max_scan_bytes` exceeds `SFX_MAX`, passed to
+`detect_format(..., budget=)` and threaded through `open_archive` the same way. It SHALL
+NOT be an `ArchiveyConfig` field: `#273` already gave detection a cost-control channel, and
+a second flag would be two knobs for one decision. A source already matched at offset 0
+never consults the scan bound, which is not an error.
```
davitf: ## F2 🔴 `[blocking]` · CONFIRMED — the exhaustive-scan opt-in lost its freeze surface when the `archive-reading` MODIFIED requirement was deleted

**Location:** these lines, against live `openspec/specs/archive-reading/spec.md:20-36`.

This PR deletes the whole `MODIFIED Requirement: Explicit configuration object` block (77 lines) — correct, since `exhaustive_prefix_scan` is gone. But the surviving ADDED requirement now says the opt-in is *"a `DetectionBudget` whose `max_scan_bytes` exceeds `SFX_MAX`, passed to `detect_format(..., budget=)` and **threaded through `open_archive` the same way**"*, and `tasks.md:123` (3.1) makes it explicit: *"`open_archive` SHALL pass the budget through. Threading that is this task."*

Nothing declares that. The live `archive-reading` spec freezes `open_archive`'s surface as a signature code block (`openspec/specs/archive-reading/spec.md:24-36`) with eight keyword-only parameters and no `budget`. No delta in this change modifies `Requirement: Opening an archive for reading`. I grepped both the delta and the live `archive-reading` and `detection-cost` specs: `budget` appears on `open_archive` nowhere outside prose.

So on `openspec archive` this change ships a self-contradictory `archive-reading` spec — one requirement saying the opt-in is threaded through `open_archive`, another showing a signature that cannot express it.

This is the same hazard the *old* task 3.1 spelled out for the config field:

> the dataclass there is the freeze surface, so the field must land in it rather than in prose elsewhere, or `openspec archive` will not ship it

The discipline was dropped along with the field it was written for.

**Fix direction.** Add a `MODIFIED Requirement: Opening an archive for reading` to this delta restating the signature with `budget: DetectionBudget | DetectionBudgetPreset | None = None`. Two sub-questions to settle there rather than leave inferred:

- How does `budget=` interact with the existing *"an explicit argument the resolved backend cannot act on is handled by its intent"* rule (`openspec/specs/archive-reading/spec.md:41-50`)? A budget is a *resource offered*, not an assertion about the archive, so by that table it permits-and-diagnoses rather than refuses — but the spec should say so.
- Is `format=` (which bypasses detection) plus `budget=` a silent no-op or a diagnostic?

Add the corresponding scenario row, and make task 3.1 name the requirement it must land in.

---

openspec/changes/prefixed-archive-detection/specs/archive-reading/spec.md
```diff
@@ -81,23 +7,17 @@ that much from every source a caller opens is not free. A caller who knows bette
...
-The option SHALL default to off. When enabled it SHALL search the whole source for the same
-validated container signatures the cued scan uses, and a hit SHALL be subject to the same
-structural validation (`format-detection`) — the opt-in changes *how far* detection looks,
-never *how much evidence* it demands. A hit found only this way SHALL report
-`prefix_kind = UNKNOWN` and `detected_by = "exhaustive_scan"`.
+The default `BALANCED` budget SHALL leave the scan at `SFX_MAX`. When a larger bound is
+set, detection SHALL search that far for the same validated container signatures the cued
+scan uses — the opt-in changes *how far* detection looks, never *how much evidence* it
+demands. A hit found only this way SHALL report `prefix_kind = UNKNOWN` and
+`detected_by = "exhaustive_scan"`.
```
davitf: ## F3 🟡 `[important]` · CONFIRMED — the opt-in now depends on a type that is deliberately not public, and this change documents it to users anyway

**Location:** these lines and `tasks.md:161` (4.11), against `openspec/specs/detection-cost/spec.md:10-13`.

The live `detection-cost` spec says plainly:

> The types live in `archivey.detection_cost` and are accepted by `detect_format(..., budget=)`, but they are **not** re-exported from `archivey.__all__` yet. Public freeze of the root surface (top-level vs subpackage vs internal) is deferred to `detection-result-surface`.

`detect_format`'s own docstring repeats it (`src/archivey/internal/detection.py:485-487`), and I confirmed `archivey/__init__.py` exports `detect_format` but no budget type.

Meanwhile this change adds a `SHALL`-level, user-facing requirement titled *"An exhaustive prefix scan is available and off by default"*, and task 4.11 will document it in `docs/formats.md`. As written, that documentation must tell callers to construct a type from a module that is not part of the public surface, whose freeze belongs to a change sequenced *after* this one.

Worse, no preset expresses it. Checked on `main` (`src/archivey/detection_cost.py:202-233`):

```
BALANCED_BUDGET.max_scan_bytes  == _SFX_SCAN_BYTES  (2 MiB)
THOROUGH_BUDGET.max_scan_bytes  == _SFX_SCAN_BYTES  (2 MiB)   # identical
FAST_BUDGET.max_scan_bytes      == 256 KiB
```

So "exhaustive" is unreachable without a hand-built or `replace()`d budget — there is no preset a user can name.

The proposal presents the move as pure win — *"a second flag would be two knobs for one decision"* (`specs/format-detection/spec.md:255-259`) — and does not acknowledge what it costs. `ArchiveyConfig` is exported, documented and frozen; `DetectionBudget` is none of those yet. That trade-off belongs in `design.md` §Implementation decisions alongside the others, not discovered during Block 3.

**Fix direction.** Pick one and write it down:

- **(a)** re-export `DetectionBudget` / `DetectionBudgetPreset` / the presets at the root here, taking that freeze away from `detection-result-surface` — needs a task and a `detection-cost` delta amending the stability note;
- **(b)** raise `THOROUGH.max_scan_bytes` (or add a preset) so the documented opt-in has a public spelling needing no unexported import;
- **(c)** hold the exhaustive-scan requirement and task 4.11's exhaustive-scan prose until `detection-result-surface` freezes the surface, shipping only the tail / `prefix_kind` work here.

---

openspec/changes/prefixed-archive-detection/tasks.md
```diff
@@ -1,23 +1,39 @@
...
+   `.pyz` / `.whl` / `.apk` extensions, cost bound for shebang non-archives. No
+   `prefix_kind`, no tail probe, no compressor needles.
+2. **ZIP tail probe** — budget-gated (`THOROUGH.max_tail_bytes`, not `BALANCED`). Workspace
+   tail helper. Both ZIP offset conventions. JPEG+ZIP as an enabled-path test.
+3. **`prefix_kind` + scan validation + exhaustive-via-budget** — public enum
+   (`NONE`/`EXECUTABLE`/`SCRIPT`/`UNKNOWN`), 7z/RAR hit validation, conflict-diagnostic
+   members, `open_archive` passes `DetectionBudget`.
+4. **Makeself** — shebang-only compressor needles (gzip + bzip2 + xz/zstd/lz4/lzip),
+   inner-TAR at the hit, and TAR/single-file `start_offset` (task 5.7) in the same PR.
+
```
davitf: ## F4 🟡 `[important]` · CONFIRMED — the four-block split is not mapped onto the tasks

**Location:** this `## Implementation blocks` section, `tasks.md:1-16`.

The PR exists so the change *"is not reviewed as one 64-task change"*, and this section describes the four PRs in prose. But the task list below it is still one flat sequence: **51 open tasks, of which 3 name a block** (2.7, 5.7, and 3.5 only by oblique reference). An implementor picking up Block 1 has to reverse-engineer its scope from prose, and several tasks do not map obviously:

- **3.4** (register `.jar` / `.pyz` / `.whl` / `.apk`) is Block 1 by the prose, but sits in §3 under `prefix_kind`.
- **3.5** says *"only reachable under a budget that grants `TAIL`"* — that is Block 2 work, sitting in the Block 3 section, with its explanatory clause pointing at Block 1.
- **2.9** (the `registry.sfx_magic_entries()` policy comment) is arguably Block 1 or Block 4: the comment it fixes is about stream codecs (Block 4), but it becomes wrong the moment Block 1 lands.
- **4.3** (SFX matrix across PE / ELF / Mach-O / shebang, for 7z and RAR) spans at least two blocks.
- **4.7** (`prefix_kind` values for the fixtures in 4.1–4.3) is Block 3 but depends on Block 4 fixtures.

**Fix direction.** Tag every open task `(Block N)`, or add an explicit task-ID roster under each of the four bullets here. Cheap — and it is the deliverable this PR is named for.

---

openspec/changes/prefixed-archive-detection/proposal.md
```diff
@@ -100,11 +102,10 @@ tier detection accordingly instead of applying one rule to all of them.
...
 - Public API: `FormatInfo` gains `prefix_kind` (always present, default `NONE`) and a
-  `PrefixKind` enum; `ArchiveyConfig` gains `exhaustive_prefix_scan: bool = False`. The
-  opt-in is a **config field, not a keyword argument** — `detect_format` takes no per-call
-  operational keywords, so a kwarg on `open_archive` could not be expressed there. Files
-  that used to raise `FormatDetectionError` now open — a behaviour change, and the point of
-  the change.
+  `PrefixKind` enum (`NONE` / `EXECUTABLE` / `SCRIPT` / `UNKNOWN`). Tail and exhaustive
+  scan are `DetectionBudget` numbers (`max_tail_bytes`, `max_scan_bytes`), not
+  `ArchiveyConfig.exhaustive_prefix_scan`. Files that used to raise `FormatDetectionError`
+  now open — a behaviour change, and the point of the change.
```
davitf: ## F5 🟡 `[important]` · CONFIRMED — `proposal.md` still asserts the defaults this PR reverses

**Locations:** `proposal.md:78`, `:89`, `:103` (and `:83`) — all outside this hunk, which is why this comment is anchored on the Public API bullet that *was* corrected.

`What Changes` was rewritten, but three downstream sections were not, and each now states the opposite of the settled decision:

- **`:78`** — *"the SFX requirement is replaced by a tiered … rule: an **always-on tail probe** for self-locating containers"*. It is off under `BALANCED`, and off under `THOROUGH` too until Block 2 raises it.
- **`:89`** — *"a probe the *format* bounds (ZIP's 64 KiB tail) **can run always**"*. Same contradiction, inside the Decisions section that is supposed to be the durable record.
- **`:103`** — Impact/Modules still lists *"`src/archivey/config.py` (the opt-in field)"*, directly above **this** bullet, which says the opt-in is **not** an `ArchiveyConfig` field.

`:83` (*"`archive-reading` — `open_archive` gains the exhaustive-scan opt-in and passes it through"*) is now only half-true and should say what it gains — see **F2**.

This is more than tidiness: `proposal.md` is what a reader opens first, and `design.md:208-210` sets the rule that stale deltas get revised *"in the same commit as the matching block, not in a drive-by rewrite"* — which reasonably covers the spec deltas, but not the proposal's own Decisions section, which is the record of what was settled.

**Fix direction.** Rewrite `:78`, `:89`, `:103` and `:83` to match the four decisions. If any is deliberately left describing the end state rather than the default, say so inline.

---

openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -497,10 +503,11 @@ carries about itself:
...
+  than `GZIP`. The inner-TAR decode is a *resolution* step, not the format identity.
```
davitf: ## F6 🟡 `[important]` · CONFIRMED — the stream-codec validator changed to "structural", but its scenario and its test task still specify the bounded decode

**Location:** this line (`specs/format-detection/spec.md:505-511`) vs `:523` and `tasks.md:148` (task 4.2).

This bullet was rewritten so that *"Identity is the cheap structural validator … The inner-TAR decode is a **resolution** step, not the format identity."* Two places still encode the mechanism it replaced:

- **`:523`** scenario row: *"`#!` stub + gzip magic that fails **a bounded decode** | Not reported; scan continues"*.
- **`tasks.md:148`** (4.2): *"a `#!` stub whose own text contains `1f 8b 08` is rejected by **the bounded decode**"*.

A conformance test written from either would assert the wrong rejecting mechanism — and would pass for the wrong reason. For the shell-script case the structural check should reject *first*, because the byte after `1f 8b 08` is FLG and almost every printable-ASCII value sets one of its reserved bits (5–7). That is a **better** test than the decode one; it just isn't the one the task asks for.

**Fix direction.** Restate `:523` as *"gzip magic whose header bytes fail the structural check"*, and add a second row for the inner-TAR decode as the `GZIP`-vs-`TAR_GZ` resolution step. Reword 4.2's negative to name the structural rejection. Keep 4.2's `#!` + non-tar gzip → `GZIP` case — that one *is* the decode-as-resolution assertion and is correct as written.

---

openspec/changes/prefixed-archive-detection/tasks.md
```diff
@@ -92,18 +111,18 @@
...
-- [ ] 2.8 Resolve a compressor hit through the **existing inner-TAR probe** at the hit offset so a script-wrapped gzipped tar reports `TAR_GZ`, not `GZIP`, and validate it by decoding a bounded prefix — for these needles the decode is the evidence, not the magic
+- [x] 2.5a ~~**Prerequisite for 2.6–2.8: a bounded candidate-relative read.**~~ — shipped by `#273` as `PrefixWorkspace.peek_range` / `candidate_view` and `ScanNeedle` / `MagicHit.candidate_origin`. Nothing to invent here.
+- [ ] 2.5b Report `payload_offset` as the **candidate origin**, not the needle hit. A TAR hit reported at `H` rather than `H - 257` is wrong by 257 bytes and hands the backend a misaligned source; add a red–green for exactly that off-by-257. **Waits on 2.6**, which waits on the ledger.
+- [ ] 2.6 **Deferred to `detection-evidence-ledger`.** Do not add TAR's `ustar` as a container needle here — five bytes with no checksum is a false-positive risk inside a PE/ELF stub, and that change already specifies the checksum validator that would make the needle safe.
```
davitf: ## F7 🟢 `[nit]` · CONFIRMED — decided-not-to-do tasks are still open checkboxes

**Locations:** this line (2.6) and `tasks.md:167` (5.3).

Both read as decisions *not* to do the work —

- **2.6** *"**Deferred to `detection-evidence-ledger`.** Do not add TAR's `ustar` as a container needle here …"*
- **5.3** *"Dropped with `OTHER_FORMAT`. …"*

— but both carry `- [ ]`. Elsewhere in this same file the convention is applied correctly: 1.0, 2.5a, 5.5 and 5.6 are `- [x]` with `~~strikethrough~~` over the superseded text.

They inflate the open count to 51 and leave task 4.12 (archive this change) unable to finish on a clean list.

**Fix direction.** `- [x]` plus strikethrough on the superseded clause, matching how 2.5a was handled two lines above.

---

openspec/changes/prefixed-archive-detection/design.md
```diff
@@ -285,3 +337,18 @@ revision**, in the same PR that shipped the hoist:
...
+takes. Exhaustive scan and the ZIP tail become budget numbers (`max_scan_bytes`,
+`max_tail_bytes` / `max_seeks`). `open_archive` on `main` does not yet pass a budget
+through — threading that is part of the tail/exhaustive block, not a new config field.
+
+**`archive-origin-reporting` (#274) is sequenced after this change** because it reuses
+`PrefixKind` on `ArchiveInfo`. It also unifies the RAR/7z origin resolver onto `MagicHit`
+and asks whether forced `format=ZIP` should run the tail probe to fill
+`payload_offset`. Under the default decided above the answer is no: `UNKNOWN` +
+`payload_offset is None` stays the honest forced-ZIP answer unless the caller opted into
+a budget that grants `TAIL`.
```
davitf: ## F8 🟢 `[nit]` · PLAUSIBLE — the `OTHER_FORMAT` instruction for #274 has no home in the tree

**Location:** this §Sequencing paragraph, plus the §Implementation decisions note (*"`#274` … currently lists `OTHER_FORMAT` … That PR rebases onto this enum"*) and the PR body's *Follow-up for #274*.

This PR added durable sequencing notes to `openspec/changes/detection-evidence-ledger/tasks.md` and `.../detection-result-surface/tasks.md`. But `archive-origin-reporting` is **not in the tree** — it is unmerged PR #274 — so it got only a mention here and in the PR description. If #274 merges before this change, nothing in the repo tells it to drop `OTHER_FORMAT` or to keep `payload_offset is None` as the origin-not-established axis.

Tagged `PLAUSIBLE` rather than CONFIRMED because the PR body is arguably a sufficient record under `CLAUDE.md`'s *"the durable record is the repo plus the PR history"*, and because merge order may well make it moot.

**Fix direction.** Either a one-line comment on #274 pointing at this decision, or a sentence here saying explicitly that the instruction is carried on #274's thread rather than in the tree — so a future reader does not go looking for a note that was never written.

---

davitf (commented):
Review of **PR #275** at HEAD `2c8631e`, via `.claude/skills/code-review-skill` + its archivey addendum (§9 values-first, since this is an OpenSpec proposal rather than code). Findings `F1`–`F8` are posted as inline comments anchored at their locations.

**Verdict: 🔄 Request Changes** — recorded here rather than as a GitHub `REQUEST_CHANGES` event, which GitHub does not permit on one's own pull request.

---

# 1. Maintainer briefing

**What this change is.** Proposal-only — no library code. It rewrites `openspec/changes/prefixed-archive-detection/` (proposal, design, tasks, three spec deltas) to record decisions settled before implementation starts: split the 64-task change into **four PRs**; the **ZIP tail probe is off under `BALANCED`**; **no `OTHER_FORMAT`** in `PrefixKind`; **no `ArchiveyConfig.exhaustive_prefix_scan`** — tail and exhaustive scan become `DetectionBudget` numbers now that #273 shipped `detect_format(..., budget=)`; makeself needles ship in the same PR as the TAR/single-file `start_offset` backend work. Sequencing notes were added to `detection-evidence-ledger` and `detection-result-surface`, and the Makeself-aware `SKIP`/`COMPRESS` locator was parked in `dev-docs/IDEAS.md`.

**Snapshot.** 226 / 174 lines across 10 files — small, prose only. `openspec validate --strict` **re-run by me** on this HEAD: clean for `prefixed-archive-detection`, `detection-evidence-ledger`, `detection-result-surface`, and the two untouched changes. No test/lint gates apply (no code changed), so no pytest / ruff / pyrefly claim is made here.

**Main points.**

| ID | Sev | Conf | Gist |
| --- | --- | --- | --- |
| **F1** | 🔴 | CONFIRMED | ZIP's `PK\x03\x04` scan needle has no structural validator, and Block 1 widens the cue to every `#!` script. Reproduced on this HEAD: a non-archive stub containing those four bytes is reported `ZIP`, and `open_archive` calls it a `CorruptionError`. |
| **F2** | 🔴 | CONFIRMED | Deleting the `archive-reading` MODIFIED requirement removed the opt-in's freeze surface. Nothing declares `open_archive(budget=)`, so this change would archive into a spec that contradicts itself. |
| **F3** | 🟡 | CONFIRMED | The opt-in moved onto `DetectionBudget`, which `detection-cost` explicitly keeps out of `archivey.__all__` pending a *later* change — and no preset expresses "exhaustive". Task 4.11 documents it to users regardless. |
| **F4** | 🟡 | CONFIRMED | The four-block split, this PR's central deliverable, is not mapped onto the tasks: 51 open tasks, 3 name a block. |
| **F5** | 🟡 | CONFIRMED | `proposal.md` contradicts its own new defaults in three places — still says always-on tail probe, still claims a `config.py` opt-in field. |
| **F6** | 🟡 | CONFIRMED | The stream-codec validator changed from "bounded decode" to "structural", but its scenario row and task 4.2 still specify the decode. |
| **F7** | 🟢 | CONFIRMED | Decided-not-to-do tasks (2.6, 5.3) left as open `- [ ]` checkboxes. |
| **F8** | 🟢 | PLAUSIBLE | The "drop `OTHER_FORMAT`" instruction for #274 has no home in the tree. |

F1 and F2 block; both are prose fixes within this PR.

**What's fine.** The Mach-O half is argued as a live defect with a measured repro rather than a tidy-up — the strongest part of the case. Deferring `ustar` to the ledger's checksum validator rather than claiming on five bytes is the right call, and it is applied consistently across the change and the sibling task lists. Task 1.3's earliest-local-header rule is correct: I built a stdlib `zipapp` and confirmed that slicing at the first local header still opens (stdlib `zipfile`'s `concat` goes to `-23` and compensates exactly), so the rule holds under both offset conventions as claimed.

---

# 3. Maintainer decisions

1. **What validator does a ZIP scan hit have to pass? (F1)** Needed because the change's validator requirement asserts *all* scan hits are structurally confirmed while omitting the one needle Block 1 depends on — and I reproduced a non-archive being reported as a damaged ZIP on this HEAD.
   - **A** — cheap local-header sanity (version, reserved flag bits, method, name/extra lengths in-bounds). Fast, no seek, matches the bar the change just imposed on stream codecs; will still admit a rare crafted false positive.
   - **B** — require the candidate to resolve to a real EOCD + central directory. Essentially no false positives, and reaches the `CERTAIN` that `format-zip` already claims for the tail probe; costs a seek to the tail, which overlaps the Block 2 work deliberately deferred here.
   - **C** — accept the gap for Block 1 and record it, deferring validation to Block 2.
   - *Recommendation:* **A** in Block 1 — a few bytes of parsing, and it closes the script-population regression Block 1 itself creates — with **B** folded in when Block 2 lands the tail helper. **C** would ship the widened cue and the weakest needle together, which is exactly the combination the repro exercises.

2. **Where does the detection budget enter `open_archive`, and does this change re-export the budget types? (F2 + F3)** Two halves of one call, both consequences of moving the opt-in off `ArchiveyConfig`.
   - The **signature half is not optional**: something must modify `Requirement: Opening an archive for reading`, or the archived spec contradicts itself. Open sub-questions: how `budget=` interacts with the "assertion vs resource offered" rule, and whether `format=` (which bypasses detection) plus `budget=` is a no-op or a diagnostic.
   - The **export half** is a real choice: (a) re-export the budget types here, taking the root-surface freeze away from `detection-result-surface`; (b) raise `THOROUGH.max_scan_bytes` or add a preset so the documented opt-in has a public spelling; (c) hold the exhaustive-scan requirement and task 4.11's prose until `detection-result-surface` freezes the surface.
   - *Recommendation:* **(b)** — it keeps the freeze decision where you already put it, gives task 4.11 something public to document, and is one number. Note it also makes "exhaustive" a preset-reachable tier rather than a hand-built budget, which may be worth stating as a decision in its own right.

3. **Should the block split be encoded per task, or left as prose? (F4)** Needed because it is this PR's stated purpose, and I could not confidently assign several tasks (3.4, 3.5, 2.9, 4.3, 4.7) to a block from the text alone. Low-stakes and cheap either way — flagged because "an implementor can pick up Block 1 without re-deriving its scope" is the thing this PR is buying.

---

openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -8,21 +8,42 @@ look. Detection SHALL try, in order:
...
+scan continues). The later `detection-evidence-ledger` scheduler MAY record the same
+failure as a lower-evidence candidate without changing the validator function.
+
+This change SHALL NOT introduce `DetectionDeclaration`, `EvidenceClass`, competing-candidate
+ranking, or `AmbiguousFormatError`. Those belong to `detection-evidence-ledger`. Cue
+restriction for compressor needles is a backend-declared collection consulted under a
+`#!` cue, not a bitfield on `MagicSignature`. Tail location is a format-owned locator the
+generic "budget grants `TAIL`" step calls; ZIP is the first (and currently only) format
+that registers one. Inner-TAR after a compressor hit stays the existing probe-at-offset
+resolution (`TAR_GZ` versus bare `GZIP`), not a gzip-specific branch.
```
davitf: ## F9 🟡 `[important]` · CONFIRMED — the tail tier is generic in one place and ZIP-named in four others, and the change does not say which it is

**New in `3c32d1c`.** Locations: this paragraph; `design.md:285-289`; `tasks.md:174` (3.3a); `openspec/changes/detection-result-surface/tasks.md:64` (4.4); and shipped code at `src/archivey/internal/detection.py:528-532`.

This paragraph specifies the tail step as **format-neutral with a registry**: *"Tail location is a format-owned locator the generic 'budget grants `TAIL`' step calls; ZIP is the first (and currently only) format that registers one."* `design.md:285-289` says the same, adding *"that is data, not a ZIP-only tier inlined in the detector."*

Four other places spell it ZIP-specific:

1. **`tasks.md:174` (3.3a)** — `detected_by = "zip_tail_probe"`. That is a **public string on `FormatInfo`**.
2. **`detection-result-surface/tasks.md:64` (4.4)** reserves that exact ZIP-named spelling for this change to supply — so the naming is being frozen across two changes at once.
3. **Shipped by #273**: `detection.py:528-532` records the tier skip under the hardcoded key `"zip_tail"`, in the generic `_detect_format_body`, with two ZIP-specific policy branches:
   ```python
   if budget.max_tail_bytes <= 0:
       workspace.record_skip("zip_tail", TierSkipReason.NOT_ENABLED_BY_POLICY)
   elif DetectionCapability.TAIL not in workspace.capabilities():
       workspace.record_skip("zip_tail", TierSkipReason.CAPABILITY_UNAVAILABLE)
   ```
4. Meanwhile `DetectionBudget.max_tail_bytes` and `DetectionCapability.TAIL` are already format-neutral.

The tension is sharper than a naming nit, because this change's **own `design.md:191-203` (§Why the tail probe does not generalise)** argues at length that no other format can ever register a locator — 7z's pointer lives at offset 12, RAR's header sizes are backward-unwalkable varints, tar and raw streams have no tail structure. So the registry indirection is introduced with a population of exactly one and a documented argument that it will stay one.

That is not necessarily wrong — the stated goal is that `detection-evidence-ledger` can wrap the callable instead of extracting it (`design.md:274-278`), and a named format-owned locator serves that whether or not a second format ever appears. But an implementor cannot tell from the text whether Block 2 should:

- keep `zip_tail` / `zip_tail_probe` and treat "registered locators" as prose, or
- rename the shipped skip key and the reserved `detected_by` value to something format-neutral (`tail`, `tail_probe`).

**Why it matters now rather than later.** `detected_by` is public and about to be frozen by `detection-result-surface`. If a second format ever does register a tail locator, `detected_by="zip_tail_probe"` is a wrong answer for it — the kind of format-specific surprise §2 of the review addendum exists to prevent. Renaming after the freeze costs a public-string migration; deciding now costs a sentence.

**Fix direction.** Say explicitly which it is. If the tier is generic, make 3.3a emit `"tail_probe"` and file a note on `detection-result-surface` 4.4 to reserve that spelling instead, and have Block 2 rename `record_skip("zip_tail", …)`. If it is ZIP-specific in name and merely format-owned in implementation, drop "registered locators" from both this paragraph and `design.md:285-289` and say "the ZIP backend owns the locator function; the detector calls it" — which is all the ledger-wrapping goal actually needs.

---

openspec/changes/prefixed-archive-detection/design.md
```diff
@@ -205,6 +205,102 @@ It was tempting to make tier 2 "probe the last 64 KiB for anything". It only wor
...
+  tail probe does not generalise); that is data, not a ZIP-only tier inlined in the
+  detector.
+- **Cue restriction** for compressor needles is a separate backend-declared
+  collection consulted when the cue is `#!`, not a homegrown `cue_mask` bitfield on
+  `MagicSignature`. Container needles stay on `SFX_MAGIC` as today.
+- **Failure policy stays in the detector.** A failed validator in this change is not
+  reported (scan continues). The ledger later records the same failure as a
+  lower-evidence / damaged candidate (its tasks 4.6 / 4.7). Baking skip-and-continue
+  into the validator function would force a rewrite; a pass/fail (or small identity
+  outcome) lets the caller change.
```
davitf: ## F10 🟡 `[important]` · CONFIRMED — what a validator *returns* is left undecided, and the cheap answer forces exactly the rework the four-PR split exists to avoid

**New in `3c32d1c`.** Locations: this bullet; `detection-evidence-ledger/tasks.md:104-107` (4.6) and `:110-112` (4.7).

This bullet says failure policy stays in the detector, and that the validator hands back *"a pass/fail (or small identity outcome)"*. That parenthetical is the whole decision, and it is left open.

Trace it forward into the ledger tasks this PR just wrote:

- **ledger 4.6**: *"Each validator declares its failure disposition (reject / identified-but-damaged / `INCOMPLETE`) … Prefixed-archive's functions return identity (pass/fail); this task is the wrap that adds disposition, not a rewrite of the parse."*
- **ledger 4.7**: *"a 7z with a failed `StartHeaderCRC` is **still `SEVEN_Z`**, and the read raises `CorruptionError` — identity is graded, not erased."*

4.7's end state requires distinguishing **"not a 7z at all"** from **"a 7z whose header is corrupt."** A `bool` cannot carry that distinction — the information is destroyed at the point the validator returns. So if Block 1/3 ships `-> bool`, ledger 4.6's "wrap that adds disposition" is not a wrap: every validator signature and every call site changes, which is precisely the extraction-and-rewrite the four-PR shaping was designed to prevent (`design.md:274-278`, and 0.6's *"Baking skip-and-continue into the validator function would force a rewrite"*).

The same gap shows up within this change alone. The stream-codec validator has to resolve `GZIP` versus `TAR_GZ` (`specs/format-detection/spec.md:531-535`, task 2.8) — that is an identity outcome, not a boolean — while the ZIP and 7z validators as specified read like predicates. Three validators, two different return shapes, no stated contract.

**Trigger (design-level, no repro possible yet).** Block 3 lands `validate_7z_hit(view) -> bool`; ledger 4.7 then needs `SEVEN_Z` + damaged from the same call site and has to change the signature, its four call sites, and the tests pinning them.

**Fix direction.** Decide the return type in this PR and write it into 0.6, since 0.6 is the task every block obeys. A small enum costs nothing now and makes 4.6 a genuine wrap:

```python
class HitOutcome(Enum):
    NOT_THIS_FORMAT = "not_this_format"   # scan continues (this change's only behaviour)
    VALID           = "valid"
    DAMAGED         = "damaged"           # identity holds, structure does not
```

This change would treat `DAMAGED` exactly like `NOT_THIS_FORMAT` (skip and continue — the first-match behaviour it specifies), so nothing observable moves here; the ledger then only changes the *policy* that reads the outcome, never the parse. If you would rather not commit to a shape yet, say so explicitly in 0.6 — *"validators return a boolean in these four PRs; widening it to a disposition is ledger task 4.6's cost, accepted"* — so it is a recorded decision rather than a gap discovered in Block 3.

---

openspec/changes/prefixed-archive-detection/specs/archive-reading/spec.md
```diff
@@ -1,103 +1,113 @@
...
+
+> **Refuse** when the argument is an **assertion about this archive**. **Permit, and
+> record a diagnostic**, when it is a **resource offered for use if needed.**
+
+| Argument | Intent | Behaviour when the backend cannot act on it |
+| --- | --- | --- |
+| `format=` | assertion — "I claim this is a ZIP" | refuse when it cannot hold (see the directory rule below) |
+| `password=` | resource — a keyring | permit in **every** form; `PASSWORD_ARGUMENT_UNUSED` |
+| `encoding=` | resource — a hint for name decoding | permit; `ENCODING_ARGUMENT_UNUSED` |
+| `budget=` | resource — a detection spend cap | unused when `format=` bypasses detection: permit, silent no-op, no diagnostic. Detection never ran, so the bound was never consulted. Not an assertion about the archive. |
```
davitf: ## F11 🟡 `[important]` · CONFIRMED — the `budget=` row contradicts the rule it is a row of

**New in `7a3d214`** (this is part of the F2 fix, which is otherwise exactly right).

Four lines above this row, the requirement states the rule in a blockquote:

> **Refuse** when the argument is an **assertion about this archive**. **Permit, and record a diagnostic**, when it is a **resource offered for use if needed.**

This row then classifies `budget=` as **"resource — a detection spend cap"** in the Intent column, and specifies **"permit, silent no-op, no diagnostic"** in the behaviour column. By the rule as written, a resource gets a diagnostic. The two sibling resource rows do exactly that (`PASSWORD_ARGUMENT_UNUSED`, `ENCODING_ARGUMENT_UNUSED`); `budget=` is carved out without the rule being amended, and there is no `BUDGET_ARGUMENT_UNUSED` name.

The scenario table repeats the carve-out at `:66` — *"`format=ZIP` plus a non-default `budget=` | Opens; budget is a silent no-op (detection skipped); no unused-argument diagnostic"* — so it is deliberate, not a slip.

The reason given in the row (*"Detection never ran, so the bound was never consulted"*) is a real distinction: `password=` / `encoding=` go unused because the *backend* cannot act on them, whereas `budget=` goes unused because the caller's own `format=` skipped detection. But that distinction appears nowhere in the rule, so the requirement currently reads as self-contradicting, and a caller writing generic code over unused-argument diagnostics gets asymmetric behaviour across the three arguments it calls resources — the kind of per-format/per-argument surprise the addendum's §2 "no surprises" rule is about.

Note also that this same requirement's prose promises *"Unused-argument diagnostics are emitted before the reader is returned, so they are readable without listing anything"* — a promise the `budget=` row opts out of silently.

**Fix direction.** Either is defensible; the requirement just has to say which:

- **(a) Emit it.** Add `BUDGET_ARGUMENT_UNUSED`, matching the other two rows and the rule as stated. Costs one diagnostic on a combination that is arguably a caller mistake worth surfacing.
- **(b) Keep silence, and amend the rule.** Extend the blockquote with the third case — *a resource whose consuming stage the caller explicitly skipped is permitted silently, because the caller caused the skip* — and say so in the row. This is the stronger option if you expect `format=` + `budget=` to be common in wrapper code that always passes a budget.

I'd lean **(b)**: a library that threads a budget through a generic helper would otherwise emit a diagnostic on every forced-format call, which is noise the caller cannot act on.

---

openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -8,21 +8,42 @@ look. Detection SHALL try, in order:
...
+Hit validators and self-locators SHALL live on the same side of that line: a named
+function on the format module (or collected next to `SFX_MAGIC`), taking a
+candidate-relative view. The generic detector SHALL call those functions; it SHALL NOT
+contain format-specific branches for ZIP, 7z, RAR, or gzip validation, or for locating a
+ZIP EOCD. A failed validator during this change's first-match scan is not reported (the
+scan continues). The later `detection-evidence-ledger` scheduler MAY record the same
+failure as a lower-evidence candidate without changing the validator function.
+
+This change SHALL NOT introduce `DetectionDeclaration`, `EvidenceClass`, competing-candidate
+ranking, or `AmbiguousFormatError`. Those belong to `detection-evidence-ledger`. Cue
```
davitf: ## F12 🟢 `[nit]` · CONFIRMED — "SHALL NOT introduce competing-candidate ranking" collides with task 2.3, which requires exactly that

**New in `3c32d1c`.**

This sentence bars *"`DetectionDeclaration`, `EvidenceClass`, **competing-candidate ranking**, or `AmbiguousFormatError`"*, and `tasks.md:169` (0.6) repeats the ban.

But `tasks.md:171` (task 2.3, Block 3) requires:

> Validate a 7z hit: `StartHeaderCRC` over the 20-byte StartHeader, and `offset + 32 + NextHeaderOffset + NextHeaderSize <= source length`, **preferring an exact end match when several candidates validate**

Preferring one of several validating candidates *is* candidate ranking. It is intra-format and tiny — a tie-break inside the 7z validator, not a cross-format scheduler — and I am confident that is what is meant. But the SHALL is unqualified, and an implementor reading 0.6 before 2.3 has to guess whether the tie-break is permitted.

**Fix direction.** One qualifier: *"competing-candidate ranking **across formats** (the scheduler's job)"*, or add to 2.3 *"an intra-format tie-break inside one validator, not the cross-format ranking 0.6 defers."* Same wording in 0.6.

---

openspec/changes/prefixed-archive-detection/tasks.md
```diff
@@ -1,23 +1,55 @@
+## Implementation blocks
+
+Four PRs, in order. Details and the decisions behind the split: `design.md`
+§Implementation decisions. Every open task is tagged `(Block N)` so an implementor
+can pick up one block without re-deriving its scope.
```
davitf: ## F13 🟢 `[nit]` · CONFIRMED — two small gaps in the otherwise-complete block map

The block map added here is thorough, and I verified it mechanically rather than by eye: parsing the file, **45 open tasks carry a `(Block N)` tag, and every per-block roster agrees with every per-task tag — zero mismatches.** The remaining open tasks carry an explicit non-block scope (`(every block)` on 0.6 and 4.10, `(finishing PR)` on 4.12, `(finishing)` on 5.1, `(maintainer)` on 5.2), all covered by the catch-all paragraph at `:24-28`. Two leftovers:

**1. Task 0.4 is the one open task with no tag at all.** The claim on this line is *"Every open task is tagged `(Block N)`"*; 0.4 (*"The framing gate's implementation has landed (#261, `bee7735`) …"*) has neither a block tag nor a scope tag. It reads as an informational prerequisite note rather than work, so the fix is probably `(context)` or checking it off like 0.5 — not assigning it a block.

**2. Task 2.9 looks mis-blocked: it is tagged `(Block 1)` but describes Block 4's mechanism.** 2.9 updates the `registry.sfx_magic_entries()` policy comment, which today reads *"the stream codecs never do, since a stub plus a bare compressed stream is not a thing anyone produces"*, and asks the new comment to **name the cue restriction**. But the shebang-only compressor-needle collection does not exist until Block 4 (2.7). After Block 1 the existing comment is still *true* — Block 1 adds no stream-codec needles — so rewriting it in Block 1 documents a mechanism that is not there yet, and a reader between Block 1 and Block 4 sees a comment describing needles the code does not have.

**Fix direction.** Retag 2.9 `(Block 4)` and move it from the Block 1 roster at `:11` to the Block 4 roster at `:22`. If you want *something* in Block 1, split it: Block 1 fixes the cue comment (task 2.2 already covers that), Block 4 fixes the stream-codec sentence.

---

davitf (commented):
Re-review (round 2) of **PR #275** at HEAD `3c32d1c`, covering `7a3d214` (round-1 fixes) and `3c32d1c` (format-owned validators). Round 1 reviewed `2c8631e`.

**Verdict: 💬 Comment** — round 1's two blockers are properly closed. Nothing in the new material blocks; three 🟡 items are decisions worth settling before Block 1 starts, since two of them freeze public strings or signatures.

---

# Status of round 1 (F1–F8)

| ID | Round 1 | Status | Where |
| --- | --- | --- | --- |
| **F1** | 🔴 ZIP scan needle unvalidated | ✅ **Fixed** | ZIP bullet added to the validator requirement (`format-detection:524-529`); new task **2.2a** (Block 1) for the check; new task **4.1a** red–green that pins *both* the `#!` and the ELF-cued form of my repro; two scenario rows; `design.md:253-257`. Option **A** with EOCD confirmation deferred to Block 2 and said so explicitly. |
| **F2** | 🔴 opt-in lost its freeze surface | ✅ **Fixed** | `MODIFIED Requirement: Opening an archive for reading` added with `budget: DetectionBudget \| DetectionBudgetPreset \| None = None` in the signature, a `budget=` row in the intent table, and two scenario rows. Both sub-questions I raised (intent-rule interaction, `format=` + `budget=`) are answered — one of the answers is **F11** below. |
| **F3** | 🟡 opt-in depends on a non-public type | ✅ **Fixed as a recorded decision** (not my recommendation, and the reasoning is better than mine) | Rejected option (b) with a real argument — *"Raising [`THOROUGH.max_scan_bytes`] would make every `THOROUGH` caller scan the whole source, which is a different product decision from enabling the ZIP tail"* — and recorded the consequence in the requirement, `design.md`, and the PR body: exhaustive scan is a hand-built budget, the types stay in `archivey.detection_cost`, this change does not take the freeze from `detection-result-surface`. That is the gap closed, which is what I asked for. |
| **F4** | 🟡 block split not mapped to tasks | ✅ **Fixed** | Per-block task rosters plus a `(Block N)` tag on every task. Verified mechanically, not by eye: 45 tagged tasks, rosters and tags agree, **zero mismatches**. Two leftovers in **F13**. |
| **F5** | 🟡 proposal contradicts its own defaults | ✅ **Fixed** | All three sites (`:78`, `:89`, `:103`) plus the `archive-reading` capability bullet. |
| **F6** | 🟡 validator vs bounded decode | ✅ **Fixed** | Scenario rows rewritten, a third row added separating identity from resolution, **and** task 4.2's wording corrected to "structural header check" — both halves, including the one that was only in a task. |
| **F7** | 🟢 decided-not-to-do tasks open | ✅ **Fixed** | 2.6 and 5.3 now `[x]` with strikethrough. |
| **F8** | 🟢 #274 instruction homeless | ✅ **Fixed** | `design.md:397-400` states the instruction rides #274's thread because there is no in-repo change to annotate, and records that it is already applied there as of `fe45330`. |

Nothing from round 1 is still open or superseded.

---

# 1. Maintainer briefing

**What changed since round 1.** `7a3d214` addressed all eight findings. `3c32d1c` then added genuinely new material: an architecture constraint saying hit validators and the ZIP tail locator are **format-owned callables the generic detector calls**, never `if ZIP` branches in `detection.py` — shaped so `detection-evidence-ledger` *wraps* them rather than re-extracting parse logic. That is carried into `format-detection` (two new paragraphs), `design.md` (§Format-owned hooks), new task **0.6** binding every block, and matching notes on ledger tasks 0.0 / 3.3 / 4.1 / 4.2 / 4.5 / 4.6 / 4.7 / 10.6.

**Snapshot.** Now 517 / 221 across 11 files; still proposal-only, still zero code. `openspec validate --strict` **re-run by me on `3c32d1c`**: clean for all three changes.

**CI: two macOS legs are red, and it is not this PR's.** `macos-latest / py3.11 / all` and `py3.14 / all` both fail in ~17s, before any test runs, at `.github/workflows/ci.yml:214`:

```
Cask 'rar' has been disabled because it does not pass the macOS Gatekeeper check!
It will be disabled on 2026-09-01.
```

That is today's date, and the diff since round 1 touches only `openspec/` and `dev-docs/` — I verified no code or workflow file is in it. Main's last run (`64d2f6c`) predates the disablement, so I cannot show it red there yet, but it will be on main's next macOS run. Worth noting the step immediately deletes the `rar` writer it just installed (`ci.yml:215-217` — "Cask ships both rar + unrar; drop the writer"), so the workflow only ever wanted `unrar`. Fixing it is a separate PR; see block 3.

**Main points (all new material).**

| ID | Sev | Conf | Gist |
| --- | --- | --- | --- |
| **F9** | 🟡 | CONFIRMED | The tail tier is specified as generic-with-a-registry in two places and ZIP-named in four others — including the public `detected_by="zip_tail_probe"` that `detection-result-surface` is about to freeze. |
| **F10** | 🟡 | CONFIRMED | What a validator *returns* is left as "pass/fail (or small identity outcome)". A bool cannot express ledger 4.7's "still `SEVEN_Z`, but damaged", so it forces the signature rework the four-PR split exists to avoid. |
| **F11** | 🟡 | CONFIRMED | The new `budget=` intent row classifies it a "resource" and then exempts it from the diagnostic the rule two lines above requires of resources. |
| **F12** | 🟢 | CONFIRMED | 0.6's "no competing-candidate ranking" collides with task 2.3's "prefer an exact end match when several candidates validate". |
| **F13** | 🟢 | CONFIRMED | Task 0.4 is the one untagged open task; task 2.9 is tagged Block 1 but describes Block 4's mechanism. |

**What's fine.** The F1 fix is better than what I asked for — task 4.1a pins the ELF-cued form of the repro as well as the shebang one, so the existing live defect gets a regression test rather than just the new path. The F3 response is the strongest part of the round: it declines the recommendation with a reason that is more correct than the recommendation was. The ledger cross-references in `3c32d1c` are unusually careful — ledger 4.7 explicitly names the behaviour change ("prefixed-archive's first-match loop skips a failed CRC; this task is the behaviour change") rather than letting two changes quietly disagree about what a failed 7z CRC means.

---

# 3. Maintainer decisions

1. **Is the tail tier ZIP-named or format-neutral? (F9)** Needs you because it freezes a public string in two changes at once. `detected_by="zip_tail_probe"` (task 3.3a) is reserved in that exact spelling by `detection-result-surface` 4.4, while the new architecture text calls the tier generic with "registered locators" — and this change's own §Why the tail probe does not generalise argues no second format can ever register one.
   - **A** — tier and value stay ZIP-named; drop "registered locators" from the two new paragraphs and say the ZIP backend owns the locator function. Simplest, and all the ledger-wrapping goal actually requires.
   - **B** — tier is generic; rename `detected_by` to `"tail_probe"`, tell `detection-result-surface` 4.4 to reserve that instead, and have Block 2 rename the `record_skip("zip_tail", …)` key #273 shipped.
   - *Recommendation:* **A**, on the strength of your own §generalise section — a registry with a permanently-one population is speculative generality, and the format-owned function gives the ledger everything it needs without it.

2. **What does a validator return? (F10)** Needs you because it is the contract every block obeys via 0.6, and the cheap answer is the expensive one. Ledger 4.6 calls itself "the wrap that adds disposition"; ledger 4.7 needs "a 7z with a failed `StartHeaderCRC` is still `SEVEN_Z`". A `bool` destroys that distinction at the return, so 4.6 stops being a wrap and becomes a signature change across every validator and call site.
   - **A** — small `HitOutcome` enum now (`NOT_THIS_FORMAT` / `VALID` / `DAMAGED`); this change treats `DAMAGED` exactly like `NOT_THIS_FORMAT`, so nothing observable moves, and the ledger later changes only the policy that reads it.
   - **B** — boolean now, and record in 0.6 that widening it is ledger 4.6's accepted cost.
   - *Recommendation:* **A** — it is three lines, and it is the difference between the ledger wrapping the parse and rewriting it, which is the entire premise of the four-PR shaping.

3. **Does an unused `budget=` emit a diagnostic? (F11)** Needs you because the requirement currently says both. The intent table calls `budget=` a resource; the rule says resources are permitted *and diagnosed*; the row says silent. Either add `BUDGET_ARGUMENT_UNUSED` to match `password=` / `encoding=`, or extend the rule with the third case (a resource whose consuming stage the caller's own `format=` skipped). *Recommendation:* the latter — otherwise wrapper code that always threads a budget emits a diagnostic on every forced-format open that the caller cannot act on.

4. **Who fixes the macOS `rar` cask? (not a finding on this PR)** Not this PR's to carry — it is proposal-only, and the cause is an external cask disabled today. But it will red every macOS `all` / `all-lowest` leg on main from now on, so it wants its own PR against `.github/workflows/ci.yml:211-217`. Since that step installs the cask only to delete the `rar` writer two lines later, sourcing `unrar` directly looks like the smaller fix. Happy to open that separately if you want it.

---

openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -207,6 +239,7 @@ archivey.detect_format(
     source: str | Path | BinaryIO,
     *,
     config: ArchiveyConfig | None = None,
+    budget: DetectionBudget | DetectionBudgetPreset | None = None,
```
davitf: budget -> detection_budget? it's unclear otherwise (it could be a decompression budget)

Actually, should this be a field in the ArchiveyConfig object instead of a separate arg? most users won't care about this

openspec/changes/prefixed-archive-detection/specs/format-detection/spec.md
```diff
@@ -247,13 +281,10 @@ reports what precedes the payload*.
...
-takes no per-call operational keywords, so a flag that must work on both `detect_format`
-and `open_archive` has exactly one place to live: `ArchiveyConfig.exhaustive_prefix_scan`,
-whose declaration is in the `archive-reading` *Explicit configuration object* requirement —
-**that dataclass is the freeze surface, and this requirement does not restate it.** It
-SHALL NOT be added as a keyword argument to `open_archive`, which would leave
-`detect_format` unable to express it.
+**The exhaustive-scan opt-in is a detection budget, not an `ArchiveyConfig` field.**
+`detect_format(..., budget=)` already exists (`detection-cost`, shipped by `#273`). A
+second flag on `ArchiveyConfig` would be two knobs for one cost decision. `open_archive`
+SHALL pass the caller's budget through rather than growing `exhaustive_prefix_scan`.
```
davitf: previously existing code is not a good justification, we're striving for zero debt. detect_format also receives a config arg, so it could use the budget from there. let's think about what makes more sense from the user's POV

PR #276, opened 2026-09-01
ci: build macOS unrar from source after Homebrew disabled the rar cask

davitf:
Homebrew disabled the `rar` cask on 2026-09-01 because the official macOS binaries fail Gatekeeper notarization. That was how this repo installed RARLAB `unrar` on the macOS CI legs.

## Approach

Compile the published **UnRAR source** (freeware decompressor, not the trialware `rar` writer) from a **pinned git commit** of a GitHub mirror, then cache the built binary:

- `scripts/install-rarlab-unrar.sh` does `git init` + `fetch --depth 1` of [`pmachapman/unrar@d861246`](https://github.com/pmachapman/unrar/commit/d8612461815f7feffcf7f1bf9c5942125b25c1b4) (`Updated to 7.2.7`; binary banner **UNRAR 7.23**), then runs the upstream unix makefile. The commit SHA is the pin — GitHub's generated archive tarballs are not used.
- GitHub Actions caches `${RUNNER_TEMP}/rarlab-unrar` keyed on the script + OS + arch. The cache stores the *built binary*; the commit pin is checked when the script actually builds, not on a cache hit.
- `setup-dev-env.sh` uses the same script on Darwin, installing into `$(brew --prefix)/bin` when Homebrew is present (macOS zsh does not put `~/.local/bin` on PATH).

CI macOS does not download from rarlab. The Windows leg still does (`unrarw64.exe`, unpinned) — tracked in `dev-docs/IDEAS.md`.

## Did dropping the writer skip tests?

No new CI skips. macOS CI already deleted `rar` after the cask install; Linux/Windows never had the writer. The corpus RAR column and `tests/fixtures/rar/` are already committed (ADR 0016).

Four tests still shell out to `rar a` at runtime and skip on every CI leg. Inventory: `tests/fixtures/rar/README.md`, ADR 0016 follow-up, `dev-docs/IDEAS.md`. Closing that is the same move as 0016 — generate once, commit, drop the live `rar a`.

Also dropped a mistaken `@requires_binary("rar")` on the corpus rarfile oracle; those rows use committed fixtures and run with unrar alone.

## Alternatives considered

| Option | Why not |
|---|---|
| Homebrew `rar` cask | Disabled 2026-09-01 (fails Gatekeeper) |
| Download from rarlab every job | Hits rarlab; same unnotarized binary Homebrew dropped |
| `brew install carlocab/personal/unrar` | Personal tap; bottles stop at Sequoia while `macos-latest` is Tahoe |
| conda-forge `libunrar` via micromamba | Adds a second package manager to a uv workflow |
| `unar` | archivey rejects it (`find_rarlab_unrar`) |
| GitHub `/archive/{sha}.tar.gz` | Generated on demand; gzip bytes can change without the tree changing |

<div><a href="https://cursor.com/agents/bc-3e6f5aae-2f67-43b3-80e6-3144384ca93e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3e6f5aae-2f67-43b3-80e6-3144384ca93e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,16 +208,26 @@ jobs:
       - name: Install unrar (Linux)
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y unrar
+      # Homebrew disabled the `rar` cask on 2026-09-01 (official binaries fail
+      # Gatekeeper notarization). Compile RARLAB UnRAR from the published source
+      # instead: that is the same unrar Debian/conda-forge ship, it does not
+      # install the trialware `rar` writer, and a GHA cache of the binary means
+      # jobs after the first do not fetch from rarlab. `unar` is not a substitute
+      # — archivey's finder requires the RARLAB banner.
+      - name: Cache compiled RARLAB unrar (macOS)
+        if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
+        uses: actions/cache@v4
+        with:
+          path: ${{ runner.temp }}/rarlab-unrar
+          key: rarlab-unrar-${{ hashFiles('scripts/install-rarlab-unrar.sh') }}-${{ runner.os }}-${{ runner.arch }}
+          save-always: true
       - name: Install unrar (macOS)
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
         run: |
-          brew install --cask rar
-          # Cask ships both rar + unrar; drop the writer. Nothing in the suite needs
-          # it — corpus RAR archives are committed (see the Linux step above) — and
-          # removing it keeps the "no trialware on CI" property explicit rather than
-          # incidental.
-          rm -f "$(brew --prefix)/bin/rar"
-          command -v unrar
+          DEST="${RUNNER_TEMP}/rarlab-unrar"
+          ./scripts/install-rarlab-unrar.sh --dest "$DEST" --cache-dir "$DEST"
+          echo "$DEST" >> "$GITHUB_PATH"
+          test -x "$DEST/unrar"
           command -v rar && exit 1 || true
       - name: Install unrar (Windows)
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'Windows'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,11 +209,11 @@ jobs:
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y unrar
       # Homebrew disabled the `rar` cask on 2026-09-01 (official binaries fail
-      # Gatekeeper notarization). Compile RARLAB UnRAR from a pinned GitHub
-      # mirror of the published source (not rarlab): same unrar Debian/conda-forge
-      # ship, no trialware `rar` writer, and a GHA cache of the binary so later
-      # jobs skip the download. `unar` is not a substitute — archivey's finder
-      # requires the RARLAB banner.
+      # Gatekeeper notarization). Compile RARLAB UnRAR from a pinned git commit
+      # of the pmachapman/unrar GitHub mirror (not rarlab, not GitHub's generated
+      # archive tarball). The GHA cache stores the *built binary*; the commit pin
+      # is checked when the script actually builds, not on a cache hit. `unar` is
+      # not a substitute — archivey's finder requires the RARLAB banner.
       - name: Cache compiled RARLAB unrar (macOS)
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
         uses: actions/cache@v4
@@ -225,7 +225,7 @@ jobs:
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
         run: |
           DEST="${RUNNER_TEMP}/rarlab-unrar"
-          ./scripts/install-rarlab-unrar.sh --dest "$DEST" --cache-dir "$DEST"
+          ./scripts/install-rarlab-unrar.sh --dest "$DEST"
           echo "$DEST" >> "$GITHUB_PATH"
           test -x "$DEST/unrar"
           command -v rar && exit 1 || true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,11 +209,11 @@ jobs:
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install -y unrar
       # Homebrew disabled the `rar` cask on 2026-09-01 (official binaries fail
-      # Gatekeeper notarization). Compile RARLAB UnRAR from the published source
-      # instead: that is the same unrar Debian/conda-forge ship, it does not
-      # install the trialware `rar` writer, and a GHA cache of the binary means
-      # jobs after the first do not fetch from rarlab. `unar` is not a substitute
-      # — archivey's finder requires the RARLAB banner.
+      # Gatekeeper notarization). Compile RARLAB UnRAR from a pinned GitHub
+      # mirror of the published source (not rarlab): same unrar Debian/conda-forge
+      # ship, no trialware `rar` writer, and a GHA cache of the binary so later
+      # jobs skip the download. `unar` is not a substitute — archivey's finder
+      # requires the RARLAB banner.
       - name: Cache compiled RARLAB unrar (macOS)
         if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
         uses: actions/cache@v4
--- a/scripts/install-rarlab-unrar.sh
+++ b/scripts/install-rarlab-unrar.sh
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 # Build RARLAB UnRAR (the freeware decompressor, not the trialware writer) from
-# the published source tarball and install the `unrar` binary.
+# a pinned GitHub mirror of the published source, and install the `unrar` binary.
 #
 # Why this exists: Homebrew disabled the `rar` cask on 2026-09-01 because the
 # official macOS binaries fail Gatekeeper notarization. Compiling from source
-# produces a local binary CI will actually run, and the tarball is cached so
-# jobs after the first do not fetch from rarlab. archivey's finder requires
-# the RARLAB banner — `unar` / `unrar-free` / `7z` are not substitutes.
+# produces a local binary CI will actually run. The tarball is fetched from
+# GitHub (not rarlab) and cached so later jobs do not re-download. archivey's
+# finder requires the RARLAB banner — `unar` / `unrar-free` / `7z` are not
+# substitutes.
+#
+# Source: https://github.com/pmachapman/unrar (mirror of rarlab UnRAR source).
+# Pin both the commit and the archive sha256; bump both together.
 #
 # Usage:
 #   scripts/install-rarlab-unrar.sh --dest DIR [--cache-dir DIR]
@@ -14,15 +18,17 @@
 # Writes DIR/unrar. No-ops if that path is already executable.
 set -euo pipefail
 
+# Mirror commit "Updated to 7.2.7" — same tree as rarlab unrarsrc-7.2.7.
 UNRAR_VERSION="7.2.7"
-UNRAR_SHA256="01d903a7dcf413cb2925696d7796e48e38d471f79bfe7ef3ad2aebf6c12dbefd"
-UNRAR_URL="https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz"
+UNRAR_COMMIT="d8612461815f7feffcf7f1bf9c5942125b25c1b4"
+UNRAR_SHA256="3d385eb009bbd657981ef317fc6625be9b627e699095019349ffec48da7fcf04"
+UNRAR_URL="https://github.com/pmachapman/unrar/archive/${UNRAR_COMMIT}.tar.gz"
 
 DEST=""
 CACHE_DIR="${XDG_CACHE_HOME:-${HOME}/.cache}/archivey/unrarsrc"
 
 usage() {
-  sed -n '2,16p' "$0" | sed 's/^# \?//'
+  sed -n '2,19p' "$0" | sed 's/^# \?//'
 }
 
 while [ $# -gt 0 ]; do
@@ -68,7 +74,7 @@ if ! command -v curl >/dev/null 2>&1; then
 fi
 
 mkdir -p "$CACHE_DIR"
-tarball="${CACHE_DIR}/unrarsrc-${UNRAR_VERSION}.tar.gz"
+tarball="${CACHE_DIR}/unrar-${UNRAR_COMMIT}.tar.gz"
 
 verify_sha256() {
   local file="$1"
@@ -105,14 +111,15 @@ cleanup() { rm -rf "$workdir"; }
 trap cleanup EXIT
 
 tar -xf "$tarball" -C "$workdir"
-src="${workdir}/unrar"
+# GitHub archive/<sha>.tar.gz extracts to unrar-<sha>/ (files at repo root).
+src="${workdir}/unrar-${UNRAR_COMMIT}"
 if [ ! -f "${src}/makefile" ]; then
-  echo "install-rarlab-unrar: tarball did not contain unrar/makefile" >&2
+  echo "install-rarlab-unrar: expected makefile at ${src}/makefile" >&2
   exit 1
 fi
 
 jobs="$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 2)"
-echo "install-rarlab-unrar: compiling UnRAR ${UNRAR_VERSION} with ${jobs} jobs"
+echo "install-rarlab-unrar: compiling UnRAR ${UNRAR_VERSION} (${UNRAR_COMMIT:0:12}) with ${jobs} jobs"
 make -C "$src" -j"$jobs"
 
 # macOS install(1) has no GNU -D; copy the binary ourselves.
--- a/scripts/install-rarlab-unrar.sh
+++ b/scripts/install-rarlab-unrar.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+# Build RARLAB UnRAR (the freeware decompressor, not the trialware writer) from
+# the published source tarball and install the `unrar` binary.
+#
+# Why this exists: Homebrew disabled the `rar` cask on 2026-09-01 because the
+# official macOS binaries fail Gatekeeper notarization. Compiling from source
+# produces a local binary CI will actually run, and the tarball is cached so
+# jobs after the first do not fetch from rarlab. archivey's finder requires
+# the RARLAB banner — `unar` / `unrar-free` / `7z` are not substitutes.
+#
+# Usage:
+#   scripts/install-rarlab-unrar.sh --dest DIR [--cache-dir DIR]
+#
+# Writes DIR/unrar. No-ops if that path is already executable.
+set -euo pipefail
+
+UNRAR_VERSION="7.2.7"
+UNRAR_SHA256="01d903a7dcf413cb2925696d7796e48e38d471f79bfe7ef3ad2aebf6c12dbefd"
+UNRAR_URL="https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz"
+
+DEST=""
+CACHE_DIR="${XDG_CACHE_HOME:-${HOME}/.cache}/archivey/unrarsrc"
+
+usage() {
+  sed -n '2,16p' "$0" | sed 's/^# \?//'
+}
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --dest)
+      DEST="${2:-}"
+      shift 2
+      ;;
+    --cache-dir)
+      CACHE_DIR="${2:-}"
+      shift 2
+      ;;
+    -h | --help)
+      usage
+      exit 0
+      ;;
+    *)
+      echo "install-rarlab-unrar: unknown argument: $1 (try --help)" >&2
+      exit 2
+      ;;
+  esac
+done
+
+if [ -z "$DEST" ]; then
+  echo "install-rarlab-unrar: --dest DIR is required" >&2
+  exit 2
+fi
+
+mkdir -p "$DEST"
+if [ -x "${DEST}/unrar" ]; then
+  echo "install-rarlab-unrar: already present at ${DEST}/unrar"
+  exit 0
+fi
+
+if ! command -v c++ >/dev/null 2>&1; then
+  echo "install-rarlab-unrar: need a C++ compiler (macOS: Xcode CLI tools)" >&2
+  exit 1
+fi
+
+if ! command -v curl >/dev/null 2>&1; then
+  echo "install-rarlab-unrar: curl is required to fetch ${UNRAR_URL}" >&2
+  exit 1
+fi
+
+mkdir -p "$CACHE_DIR"
+tarball="${CACHE_DIR}/unrarsrc-${UNRAR_VERSION}.tar.gz"
+
+verify_sha256() {
+  local file="$1"
+  local got
+  if command -v sha256sum >/dev/null 2>&1; then
+    got="$(sha256sum "$file" | awk '{ print $1 }')"
+  elif command -v shasum >/dev/null 2>&1; then
+    got="$(shasum -a 256 "$file" | awk '{ print $1 }')"
+  else
+    echo "install-rarlab-unrar: need sha256sum or shasum" >&2
+    return 1
+  fi
+  if [ "$got" != "$UNRAR_SHA256" ]; then
+    echo "install-rarlab-unrar: sha256 mismatch for $file" >&2
+    echo "  got  $got" >&2
+    echo "  want $UNRAR_SHA256" >&2
+    return 1
+  fi
+}
+
+if [ -f "$tarball" ] && verify_sha256 "$tarball"; then
+  echo "install-rarlab-unrar: using cached tarball $tarball"
+else
+  echo "install-rarlab-unrar: fetching $UNRAR_URL"
+  curl -fsSL "$UNRAR_URL" -o "$tarball"
+  if ! verify_sha256 "$tarball"; then
+    rm -f "$tarball"
+    exit 1
+  fi
+fi
+
+workdir="$(mktemp -d "${TMPDIR:-/tmp}/unrarsrc.XXXXXX")"
+cleanup() { rm -rf "$workdir"; }
+trap cleanup EXIT
+
+tar -xf "$tarball" -C "$workdir"
+src="${workdir}/unrar"
+if [ ! -f "${src}/makefile" ]; then
+  echo "install-rarlab-unrar: tarball did not contain unrar/makefile" >&2
+  exit 1
+fi
+
+jobs="$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 2)"
+echo "install-rarlab-unrar: compiling UnRAR ${UNRAR_VERSION} with ${jobs} jobs"
+make -C "$src" -j"$jobs"
+
+# macOS install(1) has no GNU -D; copy the binary ourselves.
+cp "${src}/unrar" "${DEST}/unrar"
+chmod +x "${DEST}/unrar"
+echo "install-rarlab-unrar: installed ${DEST}/unrar"
--- a/scripts/install-rarlab-unrar.sh
+++ b/scripts/install-rarlab-unrar.sh
@@ -4,13 +4,12 @@
 #
 # Why this exists: Homebrew disabled the `rar` cask on 2026-09-01 because the
 # official macOS binaries fail Gatekeeper notarization. Compiling from source
-# produces a local binary CI will actually run. The tarball is fetched from
-# GitHub (not rarlab) and cached so later jobs do not re-download. archivey's
-# finder requires the RARLAB banner — `unar` / `unrar-free` / `7z` are not
-# substitutes.
+# produces a local binary CI will actually run. archivey's finder requires the
+# RARLAB banner — `unar` / `unrar-free` / `7z` are not substitutes.
 #
-# Source: https://github.com/pmachapman/unrar (mirror of rarlab UnRAR source).
-# Pin both the commit and the archive sha256; bump both together.
+# Source: https://github.com/pmachapman/unrar (one-person mirror of rarlab
+# UnRAR source). We fetch a pinned commit with git so the SHA is
+# content-addressed; GitHub's generated archive tarballs are not used.
 #
 # Usage:
 #   scripts/install-rarlab-unrar.sh --dest DIR [--cache-dir DIR]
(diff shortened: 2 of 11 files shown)


scripts/setup-dev-env.sh
```diff
@@ -49,21 +49,27 @@ install_linux_packages() {
...
 }
 
 install_macos_packages() {
+  # Homebrew disabled the `rar` cask on 2026-09-01 (Gatekeeper). Build RARLAB
+  # UnRAR from a pinned GitHub mirror of the source — same binary the macOS CI
+  # job installs. archivey's finder requires the RARLAB banner (unar / 7z do
+  # not satisfy it). This does not need Homebrew; p7zip below still does.
+  if ! command -v unrar >/dev/null 2>&1; then
+    mkdir -p "${HOME}/.local/bin"
+    "${REPO_ROOT}/scripts/install-rarlab-unrar.sh" --dest "${HOME}/.local/bin"
```
davitf: **F1 · 🟡 important · CONFIRMED — macOS dev installs `unrar` where the dev's shell can't see it, and verification still says "ok"**

Before this PR, `brew install rar` put `unrar` in `$(brew --prefix)/bin` — on every mac dev's PATH by construction. Now it goes to `${HOME}/.local/bin`, which macOS's default zsh does not add.

The trap is that this looks fine when you run the script. Line 21 (`export PATH="${HOME}/.local/bin:${PATH}"`) puts that directory on PATH **for the script's own process**, so the closing verification block's `shutil.which("unrar")` finds it and prints `ok   unrar: …`. The developer then opens a normal shell, runs `uv run pytest`, and `unrar` is gone.

Why it matters: this is precisely the hazard this script's own header (lines 11–14) is written to prevent. RAR member-data tests and the benchmark `rar_*` cases skip, the suite still reports green, and `benchmarks/harness.py:409-435` lists `unrar` as a baseline requirement — so `--update-baselines` on such a box would rewrite `structural.json` without those cases. The false `ok` is what makes it silent.

**Trigger:** macOS with `uv` installed via Homebrew (so `~/.local/bin` never joined PATH) → run `scripts/setup-dev-env.sh` → green verification → `uv run pytest tests/test_rar_reader.py` → skips.

**Fix direction**, either:
- **(a)** install into `$(brew --prefix)/bin` when brew is present, restoring the always-on-PATH property the cask had; or
- **(b)** keep `~/.local/bin` but make the verification block honest — resolve `unrar` the way the finder does, and warn when the resolved directory is absent from the user's login-shell PATH.

(a) is smaller and matches what the old code guaranteed. CI is unaffected either way — `.github/workflows/ci.yml:256-264` covers it there.

---

scripts/install-rarlab-unrar.sh
```diff
@@ -0,0 +1,128 @@
...
+#   scripts/install-rarlab-unrar.sh --dest DIR [--cache-dir DIR]
+#
+# Writes DIR/unrar. No-ops if that path is already executable.
+set -euo pipefail
+
+# Mirror commit "Updated to 7.2.7" — same tree as rarlab unrarsrc-7.2.7.
+UNRAR_VERSION="7.2.7"
+UNRAR_COMMIT="d8612461815f7feffcf7f1bf9c5942125b25c1b4"
+UNRAR_SHA256="3d385eb009bbd657981ef317fc6625be9b627e699095019349ffec48da7fcf04"
+UNRAR_URL="https://github.com/pmachapman/unrar/archive/${UNRAR_COMMIT}.tar.gz"
```
davitf: **F2 · 🟡 important · PLAUSIBLE — macOS RAR coverage rests on a personal mirror plus a byte-stability assumption**

Two independent single points of failure, neither called out in the comment block above:

1. `pmachapman/unrar` is a one-person mirror. Deletion, rename, or a history rewrite → 404 → both macOS `all` legs fail.
2. `https://github.com/<owner>/<repo>/archive/<sha>.tar.gz` is **generated on demand**, not a stored artifact. GitHub changed its gzip output in early 2023 and invalidated pinned checksums across Homebrew, Nix and Bazel. They have since committed to advance notice, but the assumption is load-bearing here and undocumented.

Both surface identically — `verify_sha256` prints `sha256 mismatch` — which reads as a supply-chain attack when the likely cause is neither.

**What I verified**, so this is a fragility finding and not a correctness one: I cloned the mirror. Commit `d8612461815f7feffcf7f1bf9c5942125b25c1b4` exists, is the mirror's current HEAD, has message `Updated to 7.2.7`, carries a `makefile` at repo root (which line 116 requires), and the tree is genuine UnRAR source (`version.hpp`: MAJOR 7, MINOR 23, dated 2026-06-27). **The pin is correct today.**

**Fix direction** — the cheapest robust option is to fetch the commit with git rather than the archive endpoint. A commit SHA is content-addressed, so the separate checksum becomes unnecessary and failure mode (2) disappears entirely:

```sh
git -c advice.detachedHead=false clone --filter=blob:none --no-checkout "$UNRAR_REPO" "$src"
git -C "$src" fetch --depth 1 origin "$UNRAR_COMMIT"
git -C "$src" checkout --detach FETCH_HEAD
```

If you would rather keep curl + checksum, then at minimum extend the mismatch message to name the three causes in likelihood order (archive regenerated → mirror rewrote history → actual tampering) and to say what to do: re-pin after diffing the tree against rarlab's own `unrarsrc-7.2.7.tar.gz`. Whoever hits this at 2am should not be debugging an imaginary attack.

This one needs a maintainer call — see decision 1 in the review body.

---

tests/test_review_simplicity_consistency.py
```diff
@@ -518,23 +518,15 @@ def test_streaming_mode_is_uniform_across_formats(key: str, tmp_path: Path) -> N
...
+
+    ``unrar`` is enough to *read* RAR. The corpus RAR column is committed
+    (ADR 0016) and runs without the writer. What still skips on an unrar-only box
+    is the handful of tests that shell out to ``rar a`` at runtime — listed in
+    ``tests/fixtures/rar/README.md``.
+
+    This test skips where the writer *is* present (Linux ``setup-dev-env.sh``).
+    On CI / macOS it pins that the backend still registers as readable without it.
     """
     rar_is_readable = format_availability(ArchiveFormat.RAR).support is not (
```
davitf: **F3 · 🟡 important · CONFIRMED — this test asserts a tautology, and the new docstring claims otherwise**

The docstring this PR writes says: *"On CI / macOS it pins that the backend still registers as readable without it."* It does not.

`RarReader` declares `OPTIONAL_DEPENDENCY = None` (`src/archivey/internal/backends/rar_reader.py:1017`), so `_backend_available()` (`registry.py:242-243`) is unconditionally true; `ArchiveFormat.RAR.stream` is uncompressed, so the stream-codec branch is skipped; and the RAR container declares no optional member codecs. Support is always `FULL`.

**Verified empirically** — with *neither* `unrar` nor `rar` on PATH:

```
$ PATH=<empty dir>:… uv run python -c "…format_availability(ArchiveFormat.RAR)…"
unrar: None | rar: None
avail: FormatSupport.FULL
```

So the assert on line 536 cannot fail on any platform, with or without any binary. Its message — `"unrar present: RAR reads fine, yet no RAR fixture is built"` — is stale too: it describes the pre-ADR-0016 coupling, not anything the test now checks.

The tautology is **pre-existing** (this PR changed only the docstring), but the rewrite is the moment it acquires a false claim, so it's better fixed here than inherited.

**Fix direction**, either:
- **(a)** delete the test. ADR 0016 removed its reason to exist, and `tests/fixtures/rar/README.md` now carries the inventory in prose — better than an assertion could; or
- **(b)** give it a real, environment-sensitive assertion, e.g. when `shutil.which("unrar")` is not None, `find_rarlab_unrar()` must succeed. That *would* catch a non-RARLAB `unrar` on PATH — a live risk now that macOS compiles its own — and would give the test a body that matches its name.

I lean (b): the new macOS build path makes that check worth having. Either way, drop the stale assertion message.

---

scripts/install-rarlab-unrar.sh
```diff
@@ -0,0 +1,128 @@
...
+
+DEST=""
+CACHE_DIR="${XDG_CACHE_HOME:-${HOME}/.cache}/archivey/unrarsrc"
+
+usage() {
+  sed -n '2,19p' "$0" | sed 's/^# \?//'
```
davitf: **F4 · 🟢 nit · CONFIRMED — `--help` prints `set -euo pipefail` as if it were documentation**

The range ends one line too late: line 19 is `set -euo pipefail`, and `sed 's/^# \?//'` passes a non-comment line straight through.

```
$ bash scripts/install-rarlab-unrar.sh --help | tail -1
set -euo pipefail
```

Also brittle by construction — any edit to the header comment silently reshapes the help text.

**Fix:** a `cat <<'EOF' … EOF` heredoc in `usage()`, or `sed -n '2,18p'` if you want to keep the trick.

---

scripts/install-rarlab-unrar.sh
```diff
@@ -0,0 +1,128 @@
...
+  exit 2
+fi
+
+mkdir -p "$DEST"
+if [ -x "${DEST}/unrar" ]; then
+  echo "install-rarlab-unrar: already present at ${DEST}/unrar"
+  exit 0
+fi
+
+if ! command -v c++ >/dev/null 2>&1; then
```
davitf: **F5 · 🟢 nit — the preflight checks half the tools the script needs**

`c++` and `curl` get a clear up-front message; `make`, `tar`, and the sha256 tool do not.

The sha256 tool is the odd one out: its absence is only discovered inside `verify_sha256` (line 87), *after* a successful download — so the script does the network work and then exits 1. Harmless on macOS (`shasum` is always present), but it defeats the point of having a preflight block at all.

**Fix:** fold `make`, `tar`, and `sha256sum || shasum` into the same loop.

---

scripts/install-rarlab-unrar.sh
```diff
@@ -0,0 +1,128 @@
...
+# Pin both the commit and the archive sha256; bump both together.
+#
+# Usage:
+#   scripts/install-rarlab-unrar.sh --dest DIR [--cache-dir DIR]
+#
+# Writes DIR/unrar. No-ops if that path is already executable.
+set -euo pipefail
+
+# Mirror commit "Updated to 7.2.7" — same tree as rarlab unrarsrc-7.2.7.
+UNRAR_VERSION="7.2.7"
```
davitf: **F6 · 🟢 nit — `UNRAR_VERSION` and the binary's own banner disagree, unremarked**

`7.2.7` is rarlab's *source-tarball* designation (`unrarsrc-7.2.7.tar.gz`), and the mirror's commit series (`7.0.2 … 7.1.10 … 7.2.7`) follows that convention — so the comment on line 21 is correct.

But the binary built from this tree prints **7.23**: `version.hpp` has `RARVER_MAJOR 7` / `RARVER_MINOR 23`, and `cmdmix.cpp:19` formats it as `%d.%02d`. So the log line at 122 (`compiling UnRAR 7.2.7`) sits next to a binary that identifies itself as `UNRAR 7.23`.

That's a trap for whoever bumps the pin next — it took a full clone-and-read of the mirror to establish that these name the same artifact, which is exactly the argument for writing it down.

**Fix:** one parenthetical, e.g. `UNRAR_VERSION="7.2.7"  # unrarsrc tarball name; binary banner reads "UnRAR 7.23"`.

---

scripts/install-rarlab-unrar.sh
```diff
@@ -0,0 +1,128 @@
...
+    echo "  want $UNRAR_SHA256" >&2
+    return 1
+  fi
+}
+
+if [ -f "$tarball" ] && verify_sha256 "$tarball"; then
+  echo "install-rarlab-unrar: using cached tarball $tarball"
+else
+  echo "install-rarlab-unrar: fetching $UNRAR_URL"
+  curl -fsSL "$UNRAR_URL" -o "$tarball"
```
davitf: **F7 · 🟢 nit · PLAUSIBLE — an interrupted download self-heals, but reports itself as tampering**

`curl -fsSL … -o "$tarball"` writes straight to the final cache path.

A clean HTTP error leaves nothing behind — I confirmed this: a 403 through this sandbox's proxy left the cache dir empty. But a transfer cut mid-body leaves a truncated file. The next run's `verify_sha256` then fails, prints `sha256 mismatch`, falls into the else branch and re-downloads. Right outcome, alarming output — and it compounds F2, since a truncated download and a regenerated archive produce the same scary message.

**Fix:** download to `"$tarball.tmp"` and `mv` on success (atomic), or add `--remove-on-error`.

---

.github/workflows/ci.yml
```diff
@@ -208,16 +208,26 @@ jobs:
...
         run: sudo apt-get update && sudo apt-get install -y unrar
+      # Homebrew disabled the `rar` cask on 2026-09-01 (official binaries fail
+      # Gatekeeper notarization). Compile RARLAB UnRAR from a pinned GitHub
+      # mirror of the published source (not rarlab): same unrar Debian/conda-forge
+      # ship, no trialware `rar` writer, and a GHA cache of the binary so later
+      # jobs skip the download. `unar` is not a substitute — archivey's finder
+      # requires the RARLAB banner.
+      - name: Cache compiled RARLAB unrar (macOS)
+        if: (matrix.install == 'all' || matrix.install == 'all-lowest') && runner.os == 'macOS'
+        uses: actions/cache@v4
```
davitf: **F8 · 🟢 nit — the cache stores the compiled binary, so the SHA-256 pin only guards the first build**

On a cache hit the script exits at the `[ -x "${DEST}/unrar" ]` no-op (`install-rarlab-unrar.sh:61`) and the restored executable goes on PATH with no integrity check of its own.

Practical risk is low — GHA cache scoping means poisoning this key requires push access to a branch whose cache the job reads — and the `Verify RARLAB unrar on PATH` step (lines 256-264) at least forces a restored binary to produce a RARLAB banner via `find_rarlab_unrar()`. But it is worth stating plainly that the pin's guarantee ends at the job that populates the key.

If you want it to hold end-to-end: cache only `unrar-<sha>.tar.gz` (re-verified on every use) and rebuild per job, trading roughly a minute of macOS runner time. Reasonable to accept as-is — recording it so it's a decision rather than an oversight.

**Two things I checked here that are fine, for the record:**
- The key hashes `install-rarlab-unrar.sh` and the pin lives *inside* that script, so bumping the commit/sha invalidates the cache automatically. No `restore-keys`, so there's no stale-prefix fallback either. Correct.
- `save-always: true` is still a declared input on `actions/cache` v4/v5 — I checked `action.yml` on `main`; the README simply stopped documenting it. The intent is right too, since the binary is independent of test outcomes. And a failed build caches a DEST holding only the tarball, whose sha is re-verified on the next run before rebuilding — self-healing, not a poisoned key.

---

davitf (commented):
## Maintainer briefing

Reviewed **`8b1d083`** (`cursor/macos-unrar-source-build-a93e`). Round 1 — no prior review to reconcile.

**What this change is.** Homebrew disabled the `rar` cask, which was how macOS CI got RARLAB `unrar`. This PR replaces it with `scripts/install-rarlab-unrar.sh` (new, 128 lines): fetch a pinned commit of the `pmachapman/unrar` GitHub mirror, verify a pinned SHA-256, build with the upstream unix makefile, drop the binary in a GHA-cached dir. `setup-dev-env.sh` uses the same script on Darwin. The rest is documentation — an ADR 0016 follow-up, an IDEAS entry, and a `tests/fixtures/rar/README.md` inventory of the four tests that still shell out to the `rar` writer — plus one stale `@requires_binary("rar")` dropped from the corpus rarfile oracle.

**Snapshot.** Small–medium (+218 / −35, 10 files, no `src/` Python). **Verdict: 💬 Comment** — nothing blocking, two things I'd want addressed.

**Main points**

- **F1** 🟡 — macOS *dev* boxes now get `unrar` in `~/.local/bin`, which a mac shell usually doesn't have on PATH, and `setup-dev-env.sh` prints `ok unrar` anyway because it exported that dir for its own process. That's the quiet-skip failure the script's own preamble exists to prevent. CI is unaffected.
- **F3** 🟡 — `test_rar_column_is_unmeasured_without_the_rar_writer` asserts a tautology (verified: it holds with *neither* binary on PATH), and this PR's new docstring claims it pins something it does not.
- **F2** 🟡 — macOS RAR coverage now depends on one person's GitHub mirror *and* on GitHub's generated tarballs staying byte-stable. Both failure modes surface as `sha256 mismatch`, which reads like tampering.
- **F4–F8** 🟢 — script and CI nits: `--help` prints `set -euo pipefail`; partial preflight; version-string ambiguity; non-atomic download; the cache holds the built binary rather than the verified tarball.

**What's fine.** The cache key hashes the install script and the pin lives inside it, so bumping the pin invalidates the cache automatically, with no `restore-keys` to fall back to something stale. The existing `Verify RARLAB unrar on PATH` step already calls `find_rarlab_unrar()`, so a binary that builds but fails the RARLAB banner check fails CI loudly rather than skipping quietly — that closes the loop this change could plausibly have opened. Dropping `@requires_binary("rar")` from `test_corpus_rar_matches_rarfile` is correct and is proven by the Linux/macOS `all` legs, which have `unrar` and no `rar`. And the README's four-test inventory is exactly right: I checked every `requires_binary("rar")` site and there are precisely four (`test_volumes.py:101`, `test_sfx.py:386,394,513`).

### What I actually ran

CI is green on all 23 checks including both macOS legs — that is the PR's result, not mine. Separately, on this Linux box I ran `ruff check .` (clean) and `pytest` over the four test files this PR touches → **149 passed, 2 skipped**. Note this box has *both* `rar` and `unrar`, so that run does **not** exercise the unrar-only path; the Linux/macOS `all` legs in CI are the evidence for that. I also verified F3's tautology directly (`format_availability(ArchiveFormat.RAR)` → `FULL` with an emptied PATH), reproduced F4, and cloned `pmachapman/unrar` to check the pinned commit's provenance for F2. I did not re-run the full three-config gate.

---

### F9 · 💡 suggestion — "CI does not download from rarlab" is true only for macOS

No diff line to anchor this to, so it lives here.

`.github/workflows/ci.yml:236-255` — the Windows leg still does `Invoke-WebRequest 'https://www.rarlab.com/rar/unrarw64.exe'` with **no checksum and no version pin**, then executes whatever rarlab served that day as a self-extracting archive.

Pre-existing and genuinely out of scope for this PR. But this PR establishes the pinned-and-verified pattern, which makes Windows the weakest link by a wide margin — so it's the natural moment to notice. Not asking for a fix here; the PR body's blanket claim could just be softened to "the macOS leg does not download from rarlab."

---

## Maintainer decisions

1. **Is a single third-party personal mirror an acceptable dependency for macOS RAR coverage?** (F2)
   Your alternatives table rejected the four obvious options for defensible reasons, so the question isn't *whether* to use a mirror but how much to invest in the failure path.
   - **A** — Accept as-is. Cost: when the mirror or the archive endpoint changes, both macOS legs fail with a message that looks like a supply-chain attack.
   - **B** — Switch the fetch to git (`clone --filter=blob:none` + `fetch --depth 1 <sha>`). A commit SHA is content-addressed, so the pinned checksum becomes unnecessary and the "did GitHub re-gzip?" mode disappears. Roughly a ten-line change; mirror deletion still breaks it.
   - **C** — Keep curl, but rewrite the mismatch message to name the likely causes and the re-pin procedure.
   *Recommendation: B, with C's message as a fallback if you prefer the tarball path.*

2. **Delete the tautological guard test, or give it a real assertion?** (F3)
   ADR 0016 removed the coupling this test existed to pin, and `tests/fixtures/rar/README.md` now documents the residue better than an assert can. Keeping it means writing a *new* assertion (e.g. `find_rarlab_unrar()` must succeed when `shutil.which("unrar")` does), which would newly guard against a locally-compiled `unrar` that fails the RARLAB banner check — a risk this PR itself introduces.
   *Recommendation: replace the assertion rather than delete, since the new macOS build path makes that check worth having.*

3. **Track the unpinned Windows rarlab download?** (F9)
   Not this PR's job, but this PR is what makes it visible. Your call whether it goes in `dev-docs/IDEAS.md` alongside the live-`rar a` entry or stays untracked.

---

scripts/install-rarlab-unrar.sh
davitf: Correction — GitHub's sanitizer stripped my angle-bracket placeholders in point 2 above, which mangled the load-bearing bit. It should read:

> `https://github.com/{owner}/{repo}/archive/{sha}.tar.gz` is **generated on demand**, not a stored artifact.

i.e. the point is about the `/archive/{commit}.tar.gz` endpoint shape in general, not a URL with empty path segments.

---

tests/test_review_simplicity_consistency.py
davitf: Correction — GitHub's sanitizer stripped an angle-bracket placeholder in my repro block. The first line should read `PATH={an empty dir}:{uv's dir}`, i.e. a PATH containing neither `rar` nor `unrar`. The output (`FormatSupport.FULL`) is as posted.

---

.github/workflows/ci.yml
davitf: Correction — GitHub's sanitizer stripped an angle-bracket placeholder. The fix suggestion should read: cache only `unrar-{sha}.tar.gz` (i.e. the source tarball alone, re-verified on every use) and rebuild per job.

---

davitf:
## Maintainer decisions — resolved

Answers to the three decisions in the review above. Recording them here so the implementing agent has them without a scrollback.

### Decision 1 (F2) → **B: fetch with git, drop the SHA-256 pin**

Replace the `curl` + pinned-checksum fetch in `scripts/install-rarlab-unrar.sh` with a git fetch of the pinned commit:

```sh
git -c advice.detachedHead=false clone --filter=blob:none --no-checkout "$UNRAR_REPO" "$src"
git -C "$src" fetch --depth 1 origin "$UNRAR_COMMIT"
git -C "$src" checkout --detach FETCH_HEAD
```

A commit SHA is content-addressed, so `UNRAR_SHA256` and `verify_sha256` go away with it — and so does the "GitHub regenerated the archive" failure mode. Knock-on effects for the implementor:

- The `--cache-dir` tarball cache no longer has a tarball to cache; decide whether it becomes a source-tree cache or is dropped.
- **F7** (non-atomic `curl -o`) is resolved by this change rather than separately — don't fix it twice.
- **F5**'s preflight list changes: `curl` is no longer required, `git` now is. `make`, `tar` and the sha256 tool were the gap; `tar` may also fall away.

### Decision 2 (F3) → **B: replace the assertion with a RARLAB environment guard**

```python
def test_unrar_on_path_is_the_rarlab_build() -> None:
    """Environment guard: an `unrar` the finder rejects makes RAR data tests fail confusingly."""
    if shutil.which("unrar") is None:
        pytest.skip("no unrar on PATH — RAR data tests skip, the documented state")
    find_rarlab_unrar()  # raises PackageNotInstalledError when it is not the RARLAB build
```

Scope notes, so this is not over-claimed:

- **It is a diagnostic win, not a coverage win.** On CI it duplicates the `Verify RARLAB unrar on PATH` step (ci.yml:256-264). Its unique value is developer machines, and turning "~109 tests fail with a confusing `PackageNotInstalledError`" into one named failure.
- **It is not the finder's unit behaviour** — `tests/test_rar_reader.py::test_non_rarlab_unrar_rejected` already covers a fake non-RARLAB `unrar`. This pins the box the suite is running on, which nothing currently does.
- **It does not cover F1.** F1's trap is `unrar` installed but invisible on PATH, where `shutil.which("unrar")` is `None` and this guard skips like every other RAR test. F1 still needs its own fix.
- **Rename, and probably relocate.** `test_rar_column_is_unmeasured_without_the_rar_writer` names the purpose ADR 0016 deleted. `test_review_simplicity_consistency.py` currently imports only the public `archivey` surface plus `archivey.types`; `tests/test_rar_reader.py` already imports `rar_unrar` and is the more natural home for an environment guard.
- Checked and clear: `find_rarlab_unrar()` populates the module-global `_cached_unrar`, which is benign here (it caches the correct real path), and the two tests that care already `monkeypatch.setattr(..., None)`.

### Decision 3 (F9) → **yes, track it in `dev-docs/IDEAS.md`**

The unpinned, unverified `Invoke-WebRequest 'https://www.rarlab.com/rar/unrarw64.exe'` on the Windows leg gets an IDEAS entry alongside the live-`rar a` one. Not fixed in this PR.

---

### Still open, unchanged by these decisions

- **F1** 🟡 — macOS dev installs `unrar` to `~/.local/bin` where the dev's shell may not see it, and the verification block reports `ok` anyway. Recommendation stands at option (a): install into `$(brew --prefix)/bin` when brew is present.
- **F4** 🟢 — `--help` prints `set -euo pipefail`.
- **F6** 🟢 — `UNRAR_VERSION="7.2.7"` vs. the binary's own `UnRAR 7.23` banner.
- **F8** 🟢 — accepted as-is, unless decision 1's rework makes caching the source tree cheap to verify.

---

scripts/setup-dev-env.sh
```diff
@@ -49,21 +49,36 @@ install_linux_packages() {
...
 install_macos_packages() {
+  # Homebrew disabled the `rar` cask on 2026-09-01 (Gatekeeper). Build RARLAB
+  # UnRAR from a pinned GitHub mirror of the source — same binary the macOS CI
+  # job installs. archivey's finder requires the RARLAB banner (unar / 7z do
+  # not satisfy it). Install into brew's prefix when brew is present so a
+  # normal login shell can see `unrar` (macOS zsh does not put ~/.local/bin
+  # on PATH). p7zip below still needs Homebrew.
+  if ! command -v unrar >/dev/null 2>&1; then
+    if command -v brew >/dev/null 2>&1; then
+      dest="$(brew --prefix)/bin"
```
davitf: **N1 · 🟢 nit · CONFIRMED — a broken-but-present `brew` makes `dest` become `/bin`**

`command -v brew` succeeding does not mean `brew --prefix` succeeds — a partially-installed brew, or one that errors before printing, gives an empty substitution and `dest="/bin"`. The script then does `mkdir -p /bin` (a no-op, it exists) and asks the installer to `cp` a compiled binary to `/bin/unrar`.

Normally `set -e` would catch the failing substitution. **It does not here**, because this function is invoked as `install_macos_packages || system_packages_ok=0` (line 99) — the left side of `||` runs with errexit suppressed. Verified:

```
$ bash -c 'set -euo pipefail; f() { d="$(false)/bin"; echo "reached: $d"; }; if f; then echo continued; fi'
reached: /bin
continued
```

Practical impact is small — SIP makes `/bin` unwritable, so the `cp` fails, the `|| system_packages_ok=0` catches it, and the verification block reports `unrar` missing. So it surfaces rather than silently writing to `/bin`. But the message the developer gets is `cp: /bin/unrar: Operation not permitted`, which points at the wrong thing.

**Fix:** capture the prefix first and treat empty as "no brew", so the fallback branch handles it:

```sh
prefix="$(brew --prefix 2>/dev/null || true)"
if [ -n "$prefix" ]; then
  dest="${prefix}/bin"
else
  ...existing ~/.local/bin fallback, warning included...
fi
```

That also gets the no-brew warning in front of the developer in the case where they most need it.

---

davitf (commented):
## Re-review — round 2

Reviewed **`7232034`** (was `8b1d083`). Two new commits: `0d15168` (git-SHA fetch + macOS PATH install) and `7232034` (guard moved next to the finder tests).

### Status of round 1

| ID | Round 1 | Now |
| --- | --- | --- |
| **F1** | 🟡 macOS dev `unrar` lands where the shell can't see it | **Fixed** — installs into `$(brew --prefix)/bin` when brew is present; explicit warning on the fallback. One residual, N2 below. |
| **F2** | 🟡 personal mirror + archive byte-stability | **Fixed** — `git init` + `fetch --depth 1 {sha}` + `checkout --detach`. Mirror-availability risk remains but is now stated in the script header and the PR body. |
| **F3** | 🟡 tautological assertion | **Fixed** — deleted, replaced, renamed, relocated. Red–green verified below. |
| **F4** | 🟢 `--help` printed `set -euo pipefail` | **Fixed** — heredoc; output is clean. |
| **F5** | 🟢 partial preflight | **Fixed** — `need git` / `need c++` / `need make`; `curl`, `tar` and the sha256 tool are no longer needed at all. |
| **F6** | 🟢 version string vs. banner | **Fixed** — both the pin comment and the compile log now name the `7.23` banner. |
| **F7** | 🟢 non-atomic `curl -o` | **Superseded by F2** — no curl path left. The git path self-heals: a partial or wrong-commit checkout fails `rev-parse HEAD != $UNRAR_COMMIT` and is `rm -rf`'d and refetched. |
| **F8** | 🟢 cache holds the built binary | **Accepted, and improved** — CI dropped `--cache-dir "$DEST"`, so the cache now holds only the binary rather than binary + source tree, and the PR body states the pin-vs-cache-hit boundary explicitly. That was the ask: a decision, not an oversight. |
| **F9** | 💡 unpinned Windows rarlab download | **Tracked** — `dev-docs/IDEAS.md` entry, with the macOS approach named as the pattern to copy. |

All three maintainer decisions are implemented as agreed. Nothing from round 1 is still open.

### What I ran

CI is green on all 23 checks at `7232034`, both macOS legs included — that is the PR's result. Mine, on this Linux box:

- `ruff check .` — clean.
- `pytest` over `test_rar_reader.py`, `test_review_simplicity_consistency.py`, `test_sfx.py`, `test_volumes.py`, `test_rar_oracle.py` → **217 passed, 1 skipped**.
- **F2, live:** ran the script's exact fetch sequence against the mirror. `fetch --depth 1` of the pinned SHA succeeds in 1.2s, `rev-parse HEAD` returns `d861246…` exactly, `makefile` and `version.hpp` are present, `.git` is 472K. GitHub does serve reachable SHAs this way, as the comment claims. I did not compile (blocked in this sandbox) — the macOS CI legs cover that.
- **F3, red–green:** with a fake `unrar-free` shadowing the real binary on PATH, `test_unrar_on_path_is_the_rarlab_build` **fails** with `PackageNotInstalledError`; with an empty PATH it **skips** with the intended reason. It is a real guard, not another tautology. Given what F3 was, that check felt worth doing rather than assuming.
- `bash scripts/install-rarlab-unrar.sh --help` → clean, no stray `set -euo pipefail`.

### New this round

**N1 · 🟢** — a broken-but-present `brew` makes `dest` become `/bin`. Inline on `scripts/setup-dev-env.sh:69`; `set -e` does not catch it because the function runs under `|| system_packages_ok=0`. Low impact (SIP blocks the write, the failure surfaces), confusing error message.

**N2 · 🟢 — the no-brew fallback contradicts itself within one run.** `setup-dev-env.sh:73` warns *"add that directory to your login PATH or RAR data tests will skip"*, but the verification block at the end of the same run still prints an unqualified `ok   unrar: /Users/…/.local/bin/unrar` — because line 21 put `~/.local/bin` on PATH for the script's own process. The warning scrolls past; the `ok` is the last word the developer sees, which is the same false-reassurance shape F1 was about.

This only bites the no-brew branch, so it is much narrower than F1 was — most mac devs have brew and get the honest answer. Fix, if you want it closed: have the verification block flag when the resolved binary's directory is absent from the login PATH, rather than printing a bare `ok`. Reasonable to leave as-is given the explicit warning.

### Verdict

**✅ Approve** — everything from round 1 is resolved, the two mechanisms I was least sure about (fetch-by-SHA, the replacement guard) are now verified working rather than assumed, and CI is green across the matrix. N1 and N2 are both 🟢 and neither blocks; take them or leave them.

Not casting a formal GitHub approval — you are the PR author, so the API would reject it. This is the review verdict, not a merge signal.

---

PR #277, opened 2026-09-01
Detect zipapp and shebang-prefixed ZIP (prefixed-archive-detection, PR 1)

davitf:
Block 1 of `prefixed-archive-detection` (four PRs, in order — see `design.md` §Implementation decisions). **Cue + zipapp only.** No `prefix_kind`, no ZIP tail probe, no compressor needles.

## What this PR does

Detection now finds archives behind a `#!` launcher or a Mach-O stub, which is the live macOS SFX gap `#254` deferred and the reason a stdlib `zipapp` `.pyz` currently raises `FormatDetectionError` while `open_archive(format=ZIP)` already lists its members.

- **Cue widens** from `MZ` / ELF to also accept a shebang (`WEAK`) and a **Mach-O header that parses** (`STRONG`). Mach-O magic alone is not a cue: `ca fe ba be` is the Java class-file magic, and a weak cue would still put every `.class` file through the `SFX_MAX` scan.
- **ZIP scan hits are validated** with a cheap local-header check (`version_needed` 10–99, non-empty name, reserved GP flags, known method including legacy 20, name/extra in bounds **of the remaining scan window**) owned by the ZIP backend, returning internal `HitOutcome`. Four bytes of `PK\x03\x04` — including a zero-filled stub decoy — are `NOT_THIS_FORMAT`; the scan continues. EOCD confirmation is Block 2.
- **Shebang cue searches only formats with a hit validator** (F3 A, F10). Derived from `registry.sfx_hit_validators()`, so 7z/RAR join automatically when tasks 2.3–2.4 register on `SFX_HIT_VALIDATOR`. A script that mentions those magics stays `FormatDetectionError` rather than a wrong-format `CorruptionError`. `MZ` / ELF / Mach-O keep the full needle set — do not key this off `ExecutableCue.WEAK` alone (unconfirmed `MZ` stubs are the live 7z/RAR SFX path).
- **A budget-truncated validator records `sfx_scan` / `BUDGET_EXHAUSTED`** (F9). The clamp still returns `NOT_THIS_FORMAT` (identity inside the validator cannot tell cost-gate from EOF); the receipt distinguishes "searched the window, found nothing" from "hit the ceiling while confirming a candidate".
- **ZIP-family extensions** `.jar` / `.pyz` / `.whl` / `.apk` are registered so the extension fallback can rescue them.
- **Cost:** a shebang non-archive reads at most `min(size, SFX_MAX)`. A hostile local header at the scan edge cannot grow the prefix past that window (`candidate_view(..., limit=scan_limit)`). A seekable zipapp is answered by the cued scan under `BALANCED`; the tail probe does not run (`zip_tail` / `NOT_ENABLED_BY_POLICY`).
- The cue is commented as a **cost gate, not a correctness gate**. The stale “~2×” peek-cost comment is corrected to 1.66× (pre-workspace) / 1× unique I/O now.

JPEG + appended ZIP stays `FormatDetectionError` under `BALANCED` — that is Block 2 (`THOROUGH` tail).

## Review rounds

| Round | Head | Outcome |
| --- | --- | --- |
| 1 (`47257d4` → `88442fa` / `f5fd7c1`) | F1–F8 | F1–F6 fixed; F3 as A; F7 deferred to backlog; F8 no action |
| 2 (`f5fd7c1` → `347d80c`) | F9–F10 | Both fixed (nits; F9 as the recommended skip record) |

## Out of scope (later blocks)

- ZIP tail probe, `prefix_kind`, `ArchiveyConfig.detection_budget`, makeself / compressor needles.
- 7z/RAR hit validators (tasks 2.3–2.4) — slim follow-up after this PR, or with Block 3. Same `SFX_HIT_VALIDATOR` functions; they self-enable on the shebang cue.
- OpenSpec archive stays for the finishing PR (task 4.12). Task 4.10 is left unchecked because later blocks still run it.

## Gates (this revision)

- `./scripts/check.sh` — clean
- `./scripts/test.sh` (`[all]`) — 2638 passed / 24 skipped / 3 deselected
- Did not re-run `--all-configs` (behaviour is core, not extra-dependent)

<div><a href="https://cursor.com/agents/bc-20174351-08c2-49ce-85ee-0111cb248b34?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-20174351-08c2-49ce-85ee-0111cb248b34&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


src/archivey/internal/zip_detect.py
```diff
@@ -0,0 +1,88 @@
...
+    ) = struct.unpack_from("<HHHHHIIIHH", header, 4)
+    if version_needed > _VERSION_NEEDED_MAX:
+        return HitOutcome.NOT_THIS_FORMAT
+    if flags & _GP_RESERVED_MASK:
+        return HitOutcome.NOT_THIS_FORMAT
+    if method not in _KNOWN_METHODS:
+        return HitOutcome.NOT_THIS_FORMAT
+    needed = _LOCAL_HEADER_SIZE + name_len + extra_len
+    rest = peek_more(needed)
+    if len(rest) < needed:
```
davitf: ## F1 🔴 `[blocking]` · `CONFIRMED` — the validator's peek is unbounded by the cost gate, and breaks `within_budget`

`name_len` and `extra_len` come straight from the candidate bytes, so `needed` reaches `30 + 65535 + 65535` = 131 100. `peek_more` is `PrefixWorkspace.candidate_view(...)`, so `peek_range` → `ensure()` does a real forward read on the source. A candidate near the far edge of the scan window therefore grows the prefix buffer well past `scan_limit`. `_scan_for_sfx_payload` then charges `charge_scanned(min(workspace.buffered_length, scan_limit))`, which **clamps** — so `scanned_bytes` looks compliant while `unique_bytes_read` is not.

**Why it matters.** `DetectionCostReceipt.within_budget` is the repo's own encoding of the detection budget and is asserted at `tests/atheris_fuzz/targets.py:107`. The delta spec is explicit that validation "does **not** license scanning more sources than the cost gate allows" (`specs/format-detection/spec.md:549-551`), and task 4.8a — checked off in this PR — promises a `#!` non-archive source reads "no more than `min(size, SFX_MAX)`". All three are false on the input below. It also raises the worst-case in-memory prefix on a non-seekable pipe by the same 128 KiB.

**Trigger (red–green candidate).** Verified on `47257d4`:

```python
import io, struct
from archivey.internal.detection import detect_format
from archivey.detection_cost import BALANCED_BUDGET

SCAN = 2 * 1024 * 1024
hdr = b"PK\x03\x04" + struct.pack("<HHHHHIIIHH", 20, 0, 8, 0, 0, 0, 0, 0, 0xFFFF, 0xFFFF)
data = bytearray(b"#!/bin/sh\n" + b"\x90" * (SCAN + 200_000))
data[SCAN - 40 : SCAN - 40 + len(hdr)] = hdr

r = detect_format(io.BytesIO(bytes(data))).cost_receipt
print(r.unique_bytes_read, r.scanned_bytes, r.within_budget(BALANCED_BUDGET))
```

Output: `2228212 2097152 False` — the ceiling is `2097344`. Under `FAST` (`max_scan_bytes=256 KiB`) the proportional overrun is far worse.

**Fix direction.** Task 2.2a words the check as "`filename_length` / `extra_field_length` in-bounds **against the remaining source**" — a length question, not a bytes question. Cheapest first:

- **(a)** Clamp at the caller: have `_scan_for_sfx_payload` hand the validator a candidate view limited to the remaining scan budget (`workspace.candidate_view(origin, limit=scan_limit - origin)`), so an over-long name/extra reads nothing extra and grades `NOT_THIS_FORMAT`. Charge whatever it does read.
- **(b)** Answer from length rather than bytes: expose an availability query on the view (or pass `workspace.remaining_known()` in) and compare against `needed` without materialising it. Note `peek_range` returns `bytes(self._buf[origin:end])` — two full copies of whatever it fetches — so (b) also removes a per-candidate 128 KiB copy.
- **(c)** Drop the availability arm entirely. It only rejects candidates within ~128 KiB of EOF, and Block 2's EOCD confirmation is the real bound check.

Either way, add a cost assertion alongside `test_shebang_non_archive_reads_at_most_min_size_sfx_max` pinning `within_budget(BALANCED_BUDGET)` for a hostile-header source — the existing test only exercises a benign script.

---

src/archivey/internal/zip_detect.py
```diff
@@ -0,0 +1,88 @@
...
+        _usize,
+        name_len,
+        extra_len,
+    ) = struct.unpack_from("<HHHHHIIIHH", header, 4)
+    if version_needed > _VERSION_NEEDED_MAX:
+        return HitOutcome.NOT_THIS_FORMAT
+    if flags & _GP_RESERVED_MASK:
+        return HitOutcome.NOT_THIS_FORMAT
+    if method not in _KNOWN_METHODS:
+        return HitOutcome.NOT_THIS_FORMAT
```
davitf: ## F2 🟡 `[important]` · `CONFIRMED` — the cheap check accepts a zero-filled `PK\x03\x04` decoy, which is the common stub filler

`PK\x03\x04` followed by 26 zero bytes decodes as `version_needed=0`, `flags=0`, `method=0` (STORED — in `_KNOWN_METHODS`), `name_len=0`, `extra_len=0`, so `needed == 30` is always satisfied and the function returns `VALID`.

The PR body's claim "Four bytes of `PK\x03\x04` in a stub are `NOT_THIS_FORMAT`" holds only for fillers whose 6th byte is non-zero — and both new decoy tests happen to use such fillers (`\x90` in `test_a_decoy_zip_needle_is_skipped_for_the_real_payload`, ASCII text in `test_shebang_decoy_pk_bytes_are_not_a_zip`). Zero padding, which is what real ELF/PE stubs are full of, sails through.

**Why it matters.** It defeats the validator on the most likely input, and it is a spec scenario: `specs/format-detection/spec.md:561` — "`#!` stub whose text contains `PK\x03\x04` but no valid local header | Not reported; scan continues". `payload_offset` lands on the decoy, which matters directly for Block 3's `prefix_kind` and for any caller that slices the source at `payload_offset` rather than letting `zipfile` re-derive the base from the EOCD.

**Trigger (red–green candidate).** Verified on `47257d4`:

```python
stub = b"\x7fELF\x02\x01\x01" + b"\x00"*200 + b"PK\x03\x04" + b"\x00"*300
data = stub + zip_bytes()          # real ZIP starts at 511
detect_format(io.BytesIO(data)).payload_offset
```

Returns `207` (the decoy), not `511`. It still opens — `zipfile` finds the EOCD from the tail — which is precisely why the wrong offset is silent.

**Fix direction.** Two one-line additions that stay inside the "cheap" budget and kill the zero case:

- reject `name_len == 0` — a ZIP local header always names its member; no writer emits an empty name;
- reject `version_needed < 10` — 1.0 is the floor; 0 is not a value any writer produces.

Then parametrise `test_a_decoy_zip_needle_is_skipped_for_the_real_payload` over `b"\x90"` **and** `b"\x00"` filler, and add the zero-filled case to `test_zip_local_header_validator_accepts_a_real_header_and_rejects_a_decoy`.

---

src/archivey/internal/sfx.py
```diff
@@ -111,24 +145,34 @@ def executable_cue(prefix: bytes) -> ExecutableCue:
...
+    the ordinary weak-cue path (scan, then probes on a miss), never a wrong answer.
     """
     if prefix.startswith(_MZ):
         return ExecutableCue.STRONG if _is_pe(prefix) else ExecutableCue.WEAK
     if prefix.startswith(_ELF):
         return ExecutableCue.STRONG if _is_elf(prefix) else ExecutableCue.WEAK
+    if prefix.startswith(b"#!"):
+        return ExecutableCue.WEAK
+    if _is_macho(prefix):
+        return ExecutableCue.STRONG
```
davitf: ## F3 🟡 `[important]` · `CONFIRMED` — the `#!` cue lands a block before the 7z/RAR hit validators that make it safe

ZIP gets a validator in this PR; 7z and RAR get `None` (`registry.sfx_hit_validators()` simply omits them), so any needle match is reported and stops the scan. Until now the cue required `MZ`/ELF — binary prefixes where the magic rarely appears as literal text. `#!` covers every shell and Python script on the system, where a 6- or 7-byte magic can easily appear as a literal in source.

**Why it matters.** `specs/format-detection/spec.md:558,565` require "The 6 magic bytes appear in unrelated data | CRC fails; not reported; scan continues", and this module's own docstring cites `format-detection`'s "Executable-looking prefixes must not silently become a wrong stream format". Between this PR and Block 3 the opposite holds: an honest `FormatDetectionError` becomes a wrong-format `CorruptionError`. `design.md` §Implementation decisions settles the tail probe, `OTHER_FORMAT`, and the makeself ordering, but says nothing about this particular intermediate state — so under CONTRIBUTING's "unspoken deferrals are debt" it wants an explicit call either way.

**Trigger.** Verified on `47257d4`:

```python
s = b"#!/bin/sh\necho 'magic 7z\xbc\xaf\x27\x1c here'\n" + b"x"*100
detect_format(io.BytesIO(s))      # -> SEVEN_Z, payload_offset=22, PROBABLE
open_archive(io.BytesIO(s))       # -> CorruptionError: 7z signature header CRC mismatch
```

The same file without the shebang raises `FormatDetectionError` (correct). RAR behaves identically: `#!/usr/bin/env python3\nMARKER = b'Rar!\x1a\x07\x00'\n` → `RAR`, `payload_offset=34`.

**Fix direction.** This is a scoping call (see decision 1 in the review body), not something to patch silently. Cheapest in-PR options if it should close now: gate needles by cue kind so 7z/RAR needles are searched only under `MZ`/ELF/Mach-O in Block 1, or pull forward just the 7z `StartHeaderCRC` and RAR5 main-header CRC checks as `SFX_HIT_VALIDATOR`s — the seam already exists. If it stays open, it wants a recorded deferral (a note under `tasks.md` 2.1 or `review/backlog.md`).

---

src/archivey/internal/base_reader.py
```diff
@@ -172,6 +173,12 @@ class ReadBackend(ABC):
     # ``MAGIC``: ZIP declares only its local-file header here, because scanning a stub
     # for the EOCD or spanned markers would claim any file containing those four bytes.
     SFX_MAGIC: tuple[MagicSignature, ...] = ()
+    # Format-owned check the SFX scan calls on a candidate-relative view. ``None``
+    # means "needle match is enough" (today: RAR / 7z). ZIP supplies a cheap
+    # local-header sanity check so four ``PK\\x03\\x04`` bytes in a stub are not a ZIP.
+    # Returns :class:`~archivey.internal.sfx.HitOutcome`; the detector treats
+    # anything other than ``VALID`` as "skip and continue".
+    SFX_HIT_VALIDATOR: Callable[[Callable[[int], bytes]], HitOutcome] | None = None
```
davitf: ## F4 🟡 `[important]` · `CONFIRMED` — a bare function as a class attribute binds `self` on instance access

`ZipReadBackend.SFX_HIT_VALIDATOR = validate_zip_local_header` (`src/archivey/internal/backends/zip_reader.py:1470`) stores a plain function, which is a descriptor. `registry.sfx_hit_validators()` reads it as `cls.SFX_HIT_VALIDATOR`, which returns the function unbound — correct today. Any instance-level read does not:

```python
ZipReadBackend.SFX_HIT_VALIDATOR
# <function validate_zip_local_header at 0x...>
instance.SFX_HIT_VALIDATOR
# <bound method validate_zip_local_header of <ZipReadBackend object ...>>
instance.SFX_HIT_VALIDATOR(view)
# TypeError: validate_zip_local_header() takes 1 positional argument but 2 were given
```

**Why it matters.** Every other detection table on `ReadBackend` (`MAGIC`, `EXTENSIONS`, `SFX_MAGIC`, `CONTENT_PROBES`) is inert data that reads identically through class or instance. This one is not, and the annotation here (`Callable[...] | None`) advertises a plain callable, so nothing warns the next editor. Neither pyrefly nor ty flags it — I re-ran both, clean.

**Fix direction.** `SFX_HIT_VALIDATOR = staticmethod(validate_zip_local_header)` on the backend, and annotate the base as `staticmethod(...) | None` (or keep the `Callable` annotation and add a one-line comment here saying subclasses must wrap in `staticmethod`). A `test_registry` assertion that `instance.SFX_HIT_VALIDATOR is cls.SFX_HIT_VALIDATOR` pins it cheaply.

---

src/archivey/internal/sfx.py
```diff
@@ -264,10 +401,22 @@ def find_magic_in_prefix(
             index, needle = hit
             origin = candidate_origin_for_hit(index, needle.offset)
             if origin is not None:
-                return MagicHit(origin, needle.magic, needle.offset)
+                yield MagicHit(origin, needle.magic, needle.offset)
```
davitf: ## F5 🟢 `[nit]` · `CONFIRMED` — `iter_magic_in_prefix` yields duplicate hits at peek-step boundaries when needles differ in length

`overlap` (line 389) is `max(len(needle)) - 1` across *all* needles — 7 today, from RAR5's 8-byte `Rar!\x1a\x07\x01\x00` — and `search_from = max(0, searched - overlap)` (line 396). A **shorter** needle that already fitted entirely inside the previous peek — anything starting in `[searched - 7, searched - 4]` for ZIP's 4-byte `PK\x03\x04` — is re-found and re-yielded on the next step:

```python
data = bytearray(65536*2); data[65532:65536] = b"PK\x03\x04"
list(iter_magic_in_prefix(lambda n: bytes(data[:n]),
                          [ScanNeedle(b"PK\x03\x04"), ScanNeedle(b"Rar!\x1a\x07\x01\x00")],
                          limit=len(data)))
# [MagicHit(candidate_origin=65532, ...), MagicHit(candidate_origin=65532, ...)]
```

The docstring promises "every in-window match, earliest first"; duplicates are not that. **No correctness impact today** — a duplicate is either preceded by the caller's `break` on the valid case or fails its validator identically — but it costs a redundant validator run (and, given F1, a redundant 128 KiB peek), and it is the kind of thing the later evidence-ledger scheduler will trip on.

**Fix direction.** Track per-needle overlap (`searched - (len(needle.magic) - 1)`) rather than one global `overlap`, or dedupe on `candidate_origin` before yielding.

---

src/archivey/internal/zip_detect.py
```diff
@@ -0,0 +1,88 @@
...
+        16,
+        18,
+        19,
+        93,
+        95,
+        96,
+        97,
+        98,
+        99,
+    }
```
davitf: ## F6 🟢 `[nit]` · `PLAUSIBLE` — `_KNOWN_METHODS` omits legacy method 20

Method **20** is the pre-reassignment Zstandard id, still emitted by 7-Zip ZS and other older builds; APPNOTE deprecates it in favour of 93, which *is* listed here. A shebang- or stub-prefixed ZIP whose **first** local entry uses method 20 grades `NOT_THIS_FORMAT`, and the scan walks past the real payload.

Narrow, but the cost of including it is one integer, and the comment two lines up already makes the argument for it: "a method this library cannot decode can still identify as ZIP."

No repro built — I have no method-20 fixture, hence `PLAUSIBLE` rather than `CONFIRMED`.

---

src/archivey/internal/zip_detect.py
```diff
@@ -0,0 +1,88 @@
...
+
+from __future__ import annotations
+
+import struct
+from collections.abc import Callable
+
+from archivey.internal.sfx import HitOutcome
+
+_LOCAL_HEADER_MAGIC = b"PK\x03\x04"
+_LOCAL_HEADER_SIZE = 30
```
davitf: ## F7 🟢 `[nit]` · `CONFIRMED` — ZIP local-header layout is now known in a fourth place

`zip_reader.py` already has three near-identical blocks — `read_exact(fp, 30)` → magic check → `struct.unpack_from("<HH", fheader, 26)` — at lines **796**, **817** and **946**. This module adds its own `_LOCAL_HEADER_MAGIC` / `_LOCAL_HEADER_SIZE = 30` and its own struct format, so the literal `30` and the magic bytes now live in five spots (plus `MAGIC` / `SFX_MAGIC`).

The three in `zip_reader.py` are pre-existing, but `zip_detect.py` is the natural home for the constants and a small `name_and_extra_len(header)` helper the reader sites could call — and addendum §3 asks for duplication adjacent to a change to be paid now or recorded as keep-with-reason.

Not blocking. A line in `review/backlog.md` is a fine alternative to doing it here.

---

src/archivey/internal/sfx.py
```diff
@@ -67,19 +75,45 @@ def candidate_origin_for_hit(hit_offset: int, needle_offset: int) -> int | None:
...
+    ``VALID`` — identity and cheap structure both hold.
+    ``DAMAGED`` — identity holds, structure does not (a 7z whose ``StartHeaderCRC``
+    fails). This change's first-match scan treats ``DAMAGED`` like
+    ``NOT_THIS_FORMAT`` (skip, continue). The later evidence-ledger scheduler may
+    treat ``DAMAGED`` as a still-identified candidate without changing this enum.
+    """
+
+    NOT_THIS_FORMAT = "not_this_format"
+    VALID = "valid"
+    DAMAGED = "damaged"
```
davitf: ## F8 📚 — `HitOutcome.DAMAGED` is declared but never produced (no action)

Nothing on this branch returns `DAMAGED`: `validate_zip_local_header` says so explicitly in its own docstring, and `_scan_for_sfx_payload` treats it identically to `NOT_THIS_FORMAT`. That is a deliberate forward-declaration for the evidence-ledger scheduler and the docstring here says exactly that, so this needs no change.

Noting it only so a later reader doesn't mistake it for dead code and delete the member out from under Block 3's 7z `StartHeaderCRC` validator, which is what will first produce it.

---

davitf (commented):
Review of **`47257d4`** (`cursor/prefixed-archive-pr1-8b34`), `/code-review-skill` + archivey addendum. Findings F1–F8 are posted as inline comments on the lines they touch; blocks 1 and 3 are here. IDs are stable across re-reviews.

**Verdict: 🔄 Request Changes.** (Submitted as a comment review — GitHub does not accept `REQUEST_CHANGES` from a PR's own author, and this posts through @davitf's account.)

# 1. Maintainer briefing

**What this change is.** Block 1 of the four-PR `prefixed-archive-detection` change. It widens the SFX cost gate (`executable_cue`) from `MZ`/ELF to also accept a `#!` shebang (weak) and a Mach-O header that actually parses (strong, thin + fat); adds a format-owned hit-validator hook so a ZIP scan hit gets a cheap local-header sanity check before being reported; makes the cued scan skip rejected hits instead of stopping at the first needle; and registers `.jar` / `.pyz` / `.whl` / `.apk` for the extension fallback. Net effect: stdlib `zipapp` `.pyz` files and Spring-Boot-style `#!/bin/sh` + ZIP now detect and list members instead of raising `FormatDetectionError`. Touches `sfx.py`, new `zip_detect.py`, `detection.py`, `registry.py`, `base_reader.py`, `zip_reader.py`.

**Snapshot.** Medium (628+/86−, 9 files). Gates **re-run locally on this HEAD**, not copied from the PR body: `ruff check` clean; `ruff format --check` clean apart from 4 pre-existing `review/archive/` files that fail identically on `main` (verified); `pyrefly check` 0 errors; `ty check` clean; `pytest` on `[all]` in this container — **2628 passed / 24 skipped**, matching the PR body. I did **not** re-run `[all-lowest]` or `[core-only]`, and I did not reproduce the macOS CI failure (agreed it is the Homebrew `rar` cask, not this change).

**Main points**

| ID | Severity | Confidence | Gist |
| --- | --- | --- | --- |
| **F1** | 🔴 blocking | CONFIRMED | The ZIP validator peeks `30 + name_len + extra_len`, both fields attacker-controlled, so a crafted source reads ~128 KiB **past** the SFX scan budget. `cost_receipt.within_budget(BALANCED_BUDGET)` goes `False` — the invariant `tests/atheris_fuzz/targets.py:107` asserts. |
| **F2** | 🟡 important | CONFIRMED | The "cheap local-header check" accepts `PK\x03\x04` followed by **zeros**, the most common stub filler. An ELF stub with a zero-padded decoy reports `payload_offset` at the decoy, not the real payload. Both new decoy tests use fillers that happen to fail, so this is untested. |
| **F3** | 🟡 important | CONFIRMED | Widening the cue to `#!` lands *before* Block 3's 7z/RAR hit validators. An ordinary script that merely contains the 7z or RAR magic now detects as `SEVEN_Z`/`RAR` and raises `CorruptionError` where it previously raised `FormatDetectionError`. |
| **F4** | 🟡 important | CONFIRMED | `SFX_HIT_VALIDATOR` is a bare function on the class body; any instance-level access silently binds `self` as the first argument. |
| F5 | 🟢 nit | CONFIRMED | `iter_magic_in_prefix` yields duplicate hits at peek-step boundaries for needles shorter than the longest. |
| F6 | 🟢 nit | PLAUSIBLE | `_KNOWN_METHODS` omits legacy compression method 20. |
| F7 | 🟢 nit | CONFIRMED | ZIP local-header layout now known in a fourth place. |
| F8 | 📚 | — | `HitOutcome.DAMAGED` declared but not yet produced — documented, no action. |

**What's fine.** The Mach-O work is the strongest part of this PR: requiring a parsing header rather than bare `ca fe ba be` is the right call, the Java class-file collision is reasoned **in the code** and backed by a cost assertion (`test_class_file_does_not_enter_the_sfx_scan`), and both entropy shapes of the thin stub are covered — the low-entropy one being the silent-wrong-answer shape. The `HitOutcome` / `SFX_HIT_VALIDATOR` seam correctly keeps ZIP parsing out of `detection.py` while leaving room for the 7z/RAR validators. And the corrected 1.66× cost comment plus the "cost gate, not a correctness gate" docstring are exactly the doc-debt repayment tasks 2.2 / 5.4 asked for — a future editor reading only the tree now gets the *why*.

# 3. Maintainer decisions

1. **Does Block 1 ship the `#!` cue before Block 3's 7z/RAR validators?** (F3) — Right now a plain shell or Python script containing the 7z or RAR magic bytes detects as that format and raises `CorruptionError` instead of `FormatDetectionError`. That inverts the spec scenario at `specs/format-detection/spec.md:558` and the "must not silently become a wrong stream format" line the code itself cites, for however long Blocks 2–3 take. `design.md` §Implementation decisions settles several other block-ordering questions but not this one, so under CONTRIBUTING's "unspoken deferrals are debt" it needs an explicit call.
   - **A** — gate needles by cue kind: search 7z/RAR needles only under `MZ`/ELF/Mach-O in Block 1, lift the restriction when their validators land. Small, local, no spec change.
   - **B** — pull the 7z `StartHeaderCRC` and RAR5 main-header CRC checks forward into this PR. The `SFX_HIT_VALIDATOR` seam already exists; it grows the PR by maybe 60 lines and closes the gap permanently.
   - **C** — accept the exposure and record it (`tasks.md` under 2.1, or `review/backlog.md`) so the deferral is deliberate rather than unspoken.
   - *Recommendation:* B if the CRC checks are as small as the spec implies, otherwise A. Both leave the zipapp / Spring-Boot wins intact, which are the point of this PR.

2. **How should a hit validator be bounded by the cost gate?** (F1) — The delta spec says validation "does not license scanning more sources than the cost gate allows", but the `peek_more` view handed to validators has no ceiling, and `charge_scanned` clamps rather than reports. This is an interface decision Blocks 2–4 inherit: the makeself compressor needles will want a bounded structural read and the inner-TAR probe a bounded decode, both through this same seam.
   - **A** — clamp in `candidate_view` (an explicit `limit=`), so every present and future validator is bounded by construction and the clamp is visible at the call site.
   - **B** — pass remaining-length information to validators and let each answer bounds questions arithmetically, never by fetching.
   - **C** — leave the seam unbounded and drop ZIP's availability arm (Block 2's EOCD confirmation subsumes it).
   - *Recommendation:* A — the only option that also constrains the validators Blocks 3–4 will add, and it makes `charge_scanned` honest rather than clamped.

3. **Must `payload_offset` be right for a zero-filled decoy in Block 1, or can it wait for Block 2's tail probe?** (F2) — The decoy still *opens* correctly today because `zipfile` re-derives the base from the EOCD, so the only visible damage is a wrong `payload_offset`. That becomes user-visible in Block 3, when `prefix_kind` is reported alongside it. The F2 fix is two comparisons, so my read is "fix it now" — but if you would rather Block 2's EOCD confirmation be the single place that settles the offset, that is coherent and worth writing down, since the PR body currently claims the stronger property.

---

src/archivey/internal/detection_workspace.py
```diff
@@ -203,17 +203,29 @@ def peek_prefix(self, length: int) -> bytes:
...
         """A ``peek_more(n)``-shaped callable relative to ``candidate_origin``.
 
         ``peek_more(n)`` returns the first ``n`` bytes of the *candidate*, which are the
         absolute range ``[candidate_origin, candidate_origin + n)``. Served from the
         shared buffer — never a second fetch of bytes already retrieved.
+
+        ``limit`` is an exclusive archive-origin ceiling (the SFX scan passes
+        ``scan_limit``). A validator that asks for more than remains gets a short
+        read — identity fails cheaply — and must not grow the prefix past the cost
+        gate. ``None`` leaves the view unbounded.
```
davitf: ## F9 🟢 `[nit]` · `CONFIRMED` — a budget-truncated validation is indistinguishable from "not this format", and records no skip

New this round; a consequence of the F1 fix (option A, which I recommended — this is the residue, not a regression I'd trade back).

"A validator that asks for more than remains gets a short read — identity fails cheaply" conflates two different situations. When the short read is because the *source* ended, identity really did fail. When it is because the *cost gate* ended, identity may well have held and we simply could not afford to confirm it — but `validate_zip_local_header` returns `NOT_THIS_FORMAT` either way, whose own docstring in `sfx.py` says "identity **never held**".

The practical effect is a dead zone of exactly `30 + name_len + extra_len - 1` bytes at the end of the scan window, in which a genuine prefixed ZIP is silently missed. Pinned on `f5fd7c1` under `FAST` (`max_scan_bytes` 256 KiB), an ELF stub + real ZIP with a 40-byte member name (`needed == 70`):

```
header at scan_limit-68: FormatDetectionError   <-- real ZIP missed
header at scan_limit-70: ZIP off=262074         <-- found
header at scan_limit-72: ZIP off=262072
```

**Why it is only a nit.** Under `BALANCED` the gate is 2 MiB while real stubs are ~250 KB, so the dead zone sits nowhere near real data; you need a stub that ends within ~130 bytes of the ceiling. And the alternative (F1) was a genuine budget break, so the trade is right.

**Why it is still worth closing.** Nothing in the receipt says the scan ran out mid-validation — `FormatDetectionError` carries no `unavailable_tiers` entry for it, so a caller who raises `max_scan_bytes` and succeeds has no way to have known why the first attempt failed. That is the same cost-honesty axis F1 was on.

**Fix direction.** Record a skip when the clamp actually truncates a validator read — `workspace.record_skip("sfx_scan", TierSkipReason.BUDGET_EXHAUSTED)` — so the receipt distinguishes "searched the whole window, found nothing" from "hit the ceiling while confirming a candidate". Cheapest shape: have the clamped `peek_more` set a flag the scan reads after the loop, rather than plumbing a return channel through `HitOutcome`. Leaving `NOT_THIS_FORMAT` as the outcome is fine if the skip is recorded; alternatively this is what an unused `DAMAGED`-adjacent outcome could carry later, but that is Block 3's call, not this PR's.

---

src/archivey/internal/detection.py
```diff
@@ -421,39 +424,65 @@ def _warn_on_conflict(
...
+    ZIP has a local-header validator in this PR; 7z and RAR do not yet. A ``#!``
+    script that merely mentions those magics would otherwise detect as
+    ``SEVEN_Z`` / ``RAR`` and raise ``CorruptionError``. ``MZ`` / ELF / Mach-O
+    keep the full needle set. Lift this filter when the 7z/RAR
+    ``SFX_HIT_VALIDATOR``s land (tasks 2.3–2.4 — a slim follow-up, not the rest
+    of Block 3).
+    """
+    if prefix.startswith(b"#!"):
+        return [entry for entry in entries if entry.format == ArchiveFormat.ZIP]
+    return entries
```
davitf: ## F10 🟢 `[nit]` / 💡 — the filter hardcodes what it could derive, so lifting it is a manual step someone must remember

New this round; on the F3 fix, which works — both script repros are back to `FormatDetectionError`, verified on `f5fd7c1`.

Two small things, one root cause: this function re-states facts that already exist as data elsewhere.

1. **`entry.format == ArchiveFormat.ZIP`** encodes "ZIP is the format with a validator". That is exactly what `registry.sfx_hit_validators()` already returns, and `_scan_for_sfx_payload` already receives it as `validators`. Filtering on `entry.format in validators` expresses the real invariant — *only search needles we can confirm under the weakest cue* — and becomes a no-op automatically the moment 2.3/2.4 register their validators. As written, the docstring and `design.md` both have to remind someone to delete the filter by hand, and a stale filter would silently keep 7z/RAR off the shebang cue after their validators land.

2. **`prefix.startswith(b"#!")`** re-derives the shebang predicate from raw bytes, while `sfx.executable_cue` already owns "what counts as a script prefix". Two places now know it. `_detect_format_body` has `cue` in hand at the call site; passing the cue (or the `PrefixKind` Block 3 introduces) instead of the prefix bytes keeps that fact in one module.

**Fix direction.** Fold the filter into `_scan_for_sfx_payload`, which already has `entries` and `validators`, and gate it on the cue rather than the bytes — something like `if cue_is_script: entries = [e for e in entries if e.format in validators]`. Keep the current docstring's *why* (it is good, and explains a non-obvious ordering decision); it just stops being a maintenance instruction.

Not blocking — the behaviour is right today and both the design record and tasks 2.3/2.4 point at the lift.

---

davitf (commented):
Re-review of **`f5fd7c1`** (previous round was `47257d4`), `/code-review-skill` + archivey addendum. Round-2 findings F9–F10 are inline; the status table and briefing are here. IDs continue from round 1.

## Status of round 1 (F1–F8)

Every finding re-checked by re-running its own repro against `f5fd7c1`, not by reading the diff.

| ID | Was | Status | Verified how |
| --- | --- | --- | --- |
| **F1** | 🔴 blocking | **Fixed** | `candidate_view(..., limit=scan_limit)` clamps the validator's peek. The hostile-header source now raises `FormatDetectionError` instead of reading 2 228 212 bytes; `within_budget(BALANCED_BUDGET)` holds. Pinned by the new `test_hostile_zip_local_header_stays_inside_the_scan_budget`. |
| **F2** | 🟡 important | **Fixed** | `_VERSION_NEEDED_MIN = 10` plus the `name_len == 0` rejection. `PK\x03\x04` + zero-fill is now `NOT_THIS_FORMAT`; the ELF zero-filled decoy reports `payload_offset == 511` (the real payload) instead of 207. Covered by the `zero-fill` parametrisation and `test_elf_zero_filled_zip_decoy_skips_to_the_real_payload`. |
| **F3** | 🟡 important | **Fixed** (option A) | `_sfx_entries_for_prefix` keeps 7z/RAR needles off the `#!` cue. Both repro scripts — the one mentioning `MAGIC_7Z`, the one mentioning `RAR_ID` — are back to `FormatDetectionError` rather than a wrong-format `CorruptionError`. Recorded in `design.md` §Implementation decisions and in tasks 2.1 / 2.3 / 2.4, so the deferral is explicit rather than unspoken. See F10 for a nit on the filter's shape. |
| **F4** | 🟡 important | **Fixed** | `staticmethod(...)` + `ClassVar`. `inst.SFX_HIT_VALIDATOR is ZipReadBackend.SFX_HIT_VALIDATOR` is now `True` and calling it through an instance works. The added comment on the base class tells the next backend author why the wrapper is required — that is the part that keeps it fixed. |
| **F5** | 🟢 nit | **Fixed** | Per-needle floor in `_find_earliest` (`searched - (len(magic) - 1)`) replaces the single global overlap. The mixed-length case yields `[65532]` once. I re-traced the boundary in both directions: a 4-byte needle at 65532 is emitted only in step 1, at 65533 only in step 2, and an 8-byte RAR5 needle at 65528/65529 likewise — no duplicates and no misses. `scan_for_magic` still passes the default `searched=0`, so its behaviour is unchanged. |
| **F6** | 🟢 nit | **Fixed** | Method 20 accepted, with the APPNOTE reason inline. |
| **F7** | 🟢 nit | **Recorded** | Kept with reason in `review/backlog.md`, scoped to when Block 2's EOCD work touches `zip_reader.py`. That is a valid disposition under addendum §3 — the point was that it not be an unspoken deferral. |
| **F8** | 📚 | **No action needed** | Correct — it was a note, not a request. |

## Briefing

**What changed since `47257d4`.** Three commits, +182/−20 across 11 files: bound the SFX validator view and harden the ZIP local-header check (F1, F2, F6), `staticmethod` the validator hook (F4), fix the duplicate-yield boundary (F5), keep 7z/RAR needles off the shebang cue (F3), and record F7 in the backlog. Five new tests plus a parametrisation, each pinning a specific finding.

**Snapshot.** Small delta on a medium PR. Gates **re-run locally on `f5fd7c1`**, not copied from the PR body:

- `ruff check` — clean
- `ruff format --check` — clean apart from the same 4 pre-existing `review/archive/` files that fail identically on `main`
- `pyrefly check` — 0 errors; `ty check` — clean (the `ClassVar[Callable[...] | None]` + `staticmethod` combination is accepted by both)
- `./scripts/test.sh --all-configs` — **`[all]` 2636 passed / 24 skipped; `[all-lowest]` 2636 / 24; `[core-only]` 2126 / 477.** I ran all three this round specifically because the new `tests/test_sfx.py` imports `RAR_ID` at module scope; `rar_parser` is stdlib-only, so `[core-only]` collects it fine.

**Verdict: ✅ Approve.** F1–F6 are genuinely fixed rather than papered over, F7's deferral is recorded, and each fix arrived with a test that would catch its regression. F9 and F10 are 🟢 nits and do not block.

**Main points**

- 🟢 **F9** — the `limit` clamp makes a budget-truncated validation indistinguishable from "identity never held", and records no skip. A real prefixed ZIP whose header starts inside the window but whose name/extra straddle `scan_limit` is silently missed; dead zone is exactly `30 + name_len + extra_len - 1` bytes at the ceiling. Narrow, but it is the residue of the F1 fix and sits on the same cost-honesty axis.
- 🟢 **F10** — `_sfx_entries_for_prefix` hardcodes `ArchiveFormat.ZIP` and re-derives `#!` from raw bytes. Filtering on `entry.format in validators` would self-lift when 2.3/2.4 land, instead of leaving a manual deletion for someone to remember.

**What's fine.** The F5 fix is the one I'd single out: moving the overlap floor per-needle into `_find_earliest` is a smaller and more honest change than deduping origins downstream, and the docstring now states the invariant rather than the mechanism. The `version_needed` floor is inclusive of 10, which is the Info-ZIP stored value, so it does not clip real writers. And the F3 fix landed as a design-record entry *and* a task relabel *and* two tests — the deferral is now legible to someone reading only the tree, which is the pass-1 bar from addendum §8.

## Maintainer decisions

Round 1's decisions 1–3 are all settled and recorded (option A on the shebang cue, option A on the validator bound, fix-now on the zero-fill decoy) — nothing to re-open.

1. **Does F9 get closed here or in Block 2?** The dead zone is unreachable under `BALANCED` with realistic stubs, so shipping as-is is defensible. But Block 2's tail helper touches exactly this seam, and a skip record is a few lines. Options: **A** — record the `BUDGET_EXHAUSTED` skip now, while the clamp is fresh; **B** — leave it and fold it into Block 2's tail work, with a `review/backlog.md` row so it is not an unspoken deferral. *Recommendation:* A, since the "recorded skip" is the part that makes the receipt honest and it is smaller than the note explaining why it was postponed.

2. **F10 is a shape preference, not a defect** — if you would rather keep the filter explicit and format-named so it is obvious what is switched off, say so and I will not raise it again on the next round. The one thing worth weighing is that a hardcoded filter left in place after 2.3/2.4 fails silently, whereas a derived one cannot.

---

PR #278, opened 2026-09-01
docs: RAR decompressor alternatives and macOS unrar install

davitf:
Homebrew’s removal of the `rar` cask (PR #276) makes RARLAB `unrar` harder to install on macOS. This work records which other CLIs can supply RAR **member data**, then publishes install guidance aimed at pip users.

Write-up: `dev-docs/investigations/alternative-rar-decompressors.md`
User guidance: `docs/install.md`
Re-run: `python3 scripts/exploration/rar_decompressor_matrix.py`

## Verdict (backends)

No silent fallback to whatever is on `PATH` (C1).

- **`unar`** — kept **open**. Stdout concat works; RAR5 solid + any empty FILE SIGSEGVs (1.10.1) or returns rc=0 empty (1.10.7 / Homebrew XADMaster 1.10.8), including disk extract. A listing-only early-fail gate can refuse those archives. Blocked on that gate, an upstream XADMaster report, and a Homebrew-bottle matrix. Not implemented in this PR.
- **`7z`** — **closed**. Works only with the non-free Rar5 codec; no gain over RARLAB `unrar`. Homebrew `7zz` is compiled with `DISABLE_RAR_COMPRESS=1`.
- **`bsdtar` / `unrar-free`** — dead ends.

## macOS install (pip users)

`docs/install.md` leads with `brew install gromgit/new-life/unrar` (unofficial tap: bottles from the tap’s GHCR, or a compile of the formula’s RARLAB tarball — both trust the tap; no lockfile) plus a copy-paste rarlab `make` into `~/.local/bin`. The repo script is contributor/CI only.

User docs name `unar` / `7z` once because `rarfile` accepts them; archivey does not.

The rar_add **UnRAR for Mac OS X 64 bit** link is a 2018 Intel user addon (unsigned). Current official **RAR for macOS** packages include `unrar` but are ad-hoc signed (ARM) or unsigned (Intel) — Gatekeeper blocks quarantined downloads; local compiles still run on Tahoe.

## CI

Unchanged: macOS compiles `pmachapman/unrar@d861246` via `scripts/install-rarlab-unrar.sh`. Does not install `unar` or the gromgit tap.

<div><a href="https://cursor.com/agents/bc-d3ec0fd1-7f42-403a-9b1b-e3765e371082?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d3ec0fd1-7f42-403a-9b1b-e3765e371082&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


docs/install.md
```diff
@@ -17,15 +17,72 @@ containers, so a format name would be the wrong thing to install. On a free-thre
...
+Archivey looks for a binary named `unrar` whose banner contains `UNRAR` plus
+`Alexander Roshal` or `RARLAB`. Run `unrar` with no arguments to check. Common
+lookalikes are not that binary:
+
+| What you installed | What you got | Archivey RAR data |
+| --- | --- | --- |
+| Distro `unrar` / compiled RARLAB UnRAR | RARLAB `unrar` | yes |
+| `brew install 7-zip` (aliases `sevenzip`, `7zip`) | `7zz`, built with `DISABLE_RAR_COMPRESS=1` | **no** — listing a RAR may work; solid / compressed members print `ERROR: Unsupported Method` |
+| `brew install unar` | The Unarchiver | **no** — archivey will not call it (RAR5 solid + empty FILE crashes or returns empty, including extract-to-disk) |
+| `unrar-free` | DFSG unrar | **no** |
```
davitf: This table mixes linux and mac info, which makes it confusing. Move it into the Linux and macOS sections below. and I think we don't need to mention 7z and unar here as we never mention them as options; they belong in dev docs only I think.

docs/install.md
```diff
@@ -17,15 +17,72 @@ containers, so a format name would be the wrong thing to install. On a free-thre
...
+./scripts/install-rarlab-unrar.sh --dest "$(brew --prefix)/bin"
+```
+
+**Unofficial Homebrew taps** rebuild that same RARLAB source. As of 2026-09,
+`brew install gromgit/new-life/unrar` fetches `unrarsrc-7.1.10.tar.gz` from
+rarlab.com (checksummed in the formula) and installs a banner-compatible `unrar`.
+The tap is a one-person resurrection of formulae Homebrew will not ship; archivey's
+CI does not use it; bottles lag current macOS, so a new OS may compile from that
+tarball instead of downloading a bottle. If you use it, run `unrar` with no
+arguments and confirm the banner names RARLAB / Alexander Roshal.
```
davitf: As this doc is intended for end users, who will only pip install archivey or add it to their package dependencies, I think the unofficial casks may be easier to setup as they don't require the archivey repo checkout for the script. I think that could be the main guidance, with the caveat that they're unofficial and offering the direct download/script (we could provide a code block that users can simply copy/paste as an alternative).

Also, for end users, https://www.rarlab.com/rar_add.htm does link to an official `UnRAR for Mac OS X 64 bit` (https://www.rarlab.com/rar/unrar_MacOSX_10.13.2_64bit.gz). Is that the same binary that Gatekeeper will refuse? If that's properly signed, it may be a better alternative for users than compiling from source. Pleae check

docs/install.md
```diff
@@ -17,15 +17,68 @@ containers, so a format name would be the wrong thing to install. On a free-thre
...
+[RARLAB](https://www.rarlab.com/rar_add.htm) (the Windows UnRAR row) and put
+`UnRAR.exe` on `PATH`.
+
+### macOS
+
+Homebrew core does not ship `unrar` (the UnRAR license is not open source). The
+`rar` cask that used to install RARLAB's macOS binaries is disabled: those builds
+are not Apple-notarized.
+
+**If you use Homebrew**, an unofficial formula compiles UnRAR from RARLAB's
```
davitf: ### F1 🟡 `[important]` · CONFIRMED — this overstates what `brew install gromgit/new-life/unrar` does

The page says the formula "compiles UnRAR from RARLAB's published source", and line 79 adds that "the Homebrew formula and the source compile above do not hit that prompt."

I fetched the formula. It carries a `bottle do` block:

```ruby
bottle do
  root_url "https://ghcr.io/v2/gromgit/new-life"
  rebuild 1
  sha256 cellar: :any, arm64_sequoia: "f16e6cf5..."
  sha256 cellar: :any, arm64_sonoma:  "fed380e8..."
  sha256 cellar: :any, ventura:       "67853dcd..."
  sha256 cellar: :any_skip_relocation, x86_64_linux: "a68d0fb1..."
end
```

Homebrew pours a matching bottle by default, so on Sequoia and Sonoma — the two most common current macOS versions — the reader gets a **prebuilt binary from a one-maintainer third-party GHCR namespace**, not a source compile. Only bottle-less OSes (Tahoe today) actually run `make`.

**Why it matters.** This is the trust statement on a published install page: a reader is told they are compiling a checksummed rarlab.com tarball themselves, when most of them are running someone else's binary. The investigation in this same PR has it right — "**Bottle path:** prebuilt binary from a third-party GHCR namespace… the remaining bet is that gromgit's GHCR and the formula stay honest", and "A user who lets brew compile **on a bottle-less OS**". The user page contradicts the evidence file it was derived from.

The distinction is load-bearing, not pedantic, because the two paths have genuinely different trust properties:

- **Source path** — the `sha256` for `unrarsrc-7.1.10.tar.gz` is checkable against rarlab.com, a party the tap maintainer does not control (and which this PR did verify). That is a *stronger* guarantee than any PyPI dependency, where the package author controls both artifact and hash.
- **Bottle path** — the bottle `sha256` lives in the same formula, controlled by the same maintainer who controls the GHCR blob. No independent check at all, and far fewer eyes on it than on a widely-used PyPI package.

**Fix direction.** Say what the command actually does: installs RARLAB UnRAR via an unofficial tap, pouring a prebuilt bottle from the tap's own GHCR where one matches your macOS version, and compiling from RARLAB source otherwise. Mention `brew install --build-from-source gromgit/new-life/unrar` for readers who want the compile-it-yourself guarantee. Keep the (correct) Gatekeeper point on line 79 separate from the (incorrect) provenance one, and state the residual risk in a clause — the maintainer has approved leading with the tap, so this is about honest wording, not dropping the recommendation.

**Trigger:** `brew install gromgit/new-life/unrar` on macOS Sequoia or Sonoma → bottle pour, no compilation.

---

docs/install.md
```diff
@@ -17,15 +17,68 @@ containers, so a format name would be the wrong thing to install. On a free-thre
...
+it). How to get that binary is below; format quirks live on
+[Formats and extras](formats.md).
 
 ## What each format needs
 
 The per-format detail lives on [Formats and extras](formats.md); the short version
 is that every format except RAR is a pip install away, and RAR **member data** needs
-the RARLAB `unrar` binary on `PATH` — not `unrar-free`, `unar`, or `7z`. Listing and
-metadata work without it.
+the RARLAB `unrar` binary on `PATH` — not `unrar-free`. Listing and metadata work
```
davitf: ### F2 🟡 `[important]` · CONFIRMED — restore `unar` / `7z` to the non-substitute list

**Maintainer has decided this one: put the mention back.** Recording the reasoning so the fix carries it.

This PR narrows the negative list in two user-facing places, while the runtime error message widens it:

| Location | Before | After |
| --- | --- | --- |
| `docs/install.md:28` (this line) | not `unrar-free`, `unar`, or `7z` | not `unrar-free` |
| `docs/formats.md:115` | (not `unrar-free` / `unar`) | (not `unrar-free`) |
| `src/archivey/internal/backends/rar_unrar.py:27` | `unrar-free / unar / 7z` | `unrar-free / unar / 7z / 7zz` |

The investigation justifies the removal ("Lookalikes (`unar`, Homebrew `7zz`) are not mentioned in user docs"), so it was deliberate — but it points the wrong way for this PR's own audience. The change exists *because* macOS `unrar` got hard to install, `brew install unar` is the first thing a macOS user will find, and this PR's own evidence is that `unar` returns **0 bytes with rc=0** on solid RAR5 containing an empty member. Removing the warning means the only place a user learns it is after they hit the error at runtime.

**The framing the maintainer wants, and the better justification than "it's a lookalike":** `rarfile` *does* accept `unar` and `7z` as data backends. Someone migrating from `rarfile` to archivey has a working setup with one of those on `PATH` and no reason to suspect it — the mention exists for them, not as a generic warning.

**Fix direction.** One clause on each of the two lines, carrying the migration reason. Something like:

> not `unrar-free`, `unar`, or `7z` — `rarfile` accepts those, archivey does not: they either cannot read solid RAR or fail silently on it.

Forensics stay in `dev-docs/investigations/alternative-rar-decompressors.md`; only the names and the one-line reason belong in user docs.

---

scripts/exploration/rar_decompressor_matrix.py
```diff
@@ -0,0 +1,336 @@
...
+    ver: bool = False,
+) -> bytes:
+    cmd = ["unrar", "p", "-inul"]
+    if ver:
+        cmd.append("-ver")
+    cmd.append(f"-p{password}" if password else "-p-")
+    if member is not None:
+        cmd.append("-n./" + member)
+    cmd.append(str(archive))
+    return _run(cmd)[1]
```
davitf: ### F3 🟡 `[important]` · CONFIRMED — an empty gold makes any failure mode print `match=yes`

`_unrar_p()` returns `_run(cmd)[1]`, discarding the return code. A gold that *failed to build* is then indistinguishable from a gold that is legitimately empty, and `_row()` computes `match = out == gold` → `True` for any tool that also produced nothing.

The `pwd_wrong` rows at lines 243 and 257 hardcode `b""` as the gold, which makes those rows satisfied by **every** failure — missing codec, missing file, wrong password. This is live, not hypothetical. From my run on this branch (Ubuntu, `7zip` 23.01, no `7zip-rar`):

```
case                    tool   rc   len   gold   match  note
pwd_wrong:enc           7z     2    0     0      yes    empty=True ERROR: Unsupported Method : also_secret.txt
```

The row reads `match=yes` while the process never reached password handling at all — it died on the missing Rar5 codec. The investigation's table asserts `7z` gives an honest `ERROR: Wrong password` on a bad password; that is only true *with* `7zip-rar` installed, and the harness as written cannot tell the two apart. Anyone re-running this without the plugin gets a green-looking row that proves nothing about the claim it appears to support.

**Fix direction**, two small changes:

1. Have `_unrar_p` return `(rc, stdout)` and record the gold's rc in the row; set `match` to `None` when the gold itself failed, so a broken oracle can never read as agreement.
2. For the `pwd_wrong` rows, stop comparing against `b""`. Classify on the stderr token (`Wrong password` vs `Unsupported Method`) and emit `n/a` when the `7z_codec_rar5` probe came back `False`.

**Trigger:** run the script on a box without `7zip-rar` and read the `pwd_wrong:enc` row.

---

scripts/exploration/rar_decompressor_matrix.py
```diff
@@ -0,0 +1,336 @@
...
+        except ValueError:
+            return str(rc)
+    return str(rc)
+
+
+def _run(cmd: list[str], *, timeout: int = TIMEOUT) -> tuple[int, bytes, bytes]:
+    try:
+        r = subprocess.run(
+            cmd,
+            capture_output=True,
```
davitf: ### F4 🟡 `[important]` · CONFIRMED (mechanism traced; not reproduced — no `bsdtar` in my container) — the timeout bounds time, not memory

`capture_output=True` makes the parent buffer the child's entire stdout in RAM. `TIMEOUT = 8` does not bound that: the module docstring records `bsdtar --to-stdout` writing **~7 GB in 8 s** (~875 MB/s), so the buffer reaches multiple GB *before* the timeout fires. The timeout stops a hang; it does nothing about a firehose.

The mitigation at line 219 is a hardcoded denylist of the single combination already known to bomb:

```python
# bsdtar on stored nonsolid bombs (~7 GB). Only try solid, which fails fast.
if have["bsdtar"] and key.startswith("solid"):
```

Any other tool/fixture pair that loops has no guard, and the guard encodes one measurement rather than the property being defended.

This is contributor-run tooling, not library code, so it is not a product risk — but it is the one structural hazard in a file whose whole purpose is running assorted CLIs against adversarial fixtures, and it is the hazard the author already tripped over once.

**Fix direction.** Give `_run` a byte budget: redirect stdout to a `tempfile`, or read incrementally and abort past a cap. A few MB is generous — the largest legitimate gold in this matrix is 3840 bytes. Then the bsdtar special-case becomes an ordinary row with `note="output cap exceeded"` (which is *better* evidence than a comment, since it puts the bomb in the matrix instead of in a code comment), and the guard at line 219 can go.

---

scripts/exploration/rar_decompressor_matrix.py
```diff
@@ -0,0 +1,336 @@
...
+    if have["7z"]:
+        _rc, out, err = _run(["7z", "i"])
+        text = (out + err).decode("utf-8", "replace")
+        in_codecs = False
+        codec_rar5 = False
+        for line in text.splitlines():
+            if line.startswith("Codecs:"):
+                in_codecs = True
+                continue
+            if in_codecs and line.startswith("Formats:"):
```
davitf: ### F5 🟢 `[nit]` · CONFIRMED — this break is unreachable

```python
if in_codecs and line.startswith("Formats:"):
    break
```

`7z i` emits its sections in the order `Libs:` → `Formats:` → `Codecs:` → `Hashers:`. Checked against 7-Zip 23.01 on this branch:

```
6:Libs:
9:Formats:
71:Codecs:
93:Hashers:
```

`Formats:` never appears *after* `Codecs:`, so the break never fires and the scan runs to EOF straight through the `Hashers:` block. Harmless today — no hasher is named `Rar5`, and the probe correctly returned `Codecs Rar5=False` in my run — but the section boundary a reader assumes is being enforced is not.

Worth fixing because this is the load-bearing probe of the whole investigation: recommendation 5 hangs on "a backend would need `Codecs:` containing `Rar5`", and this is the code that demonstrates it.

**Fix direction.** Break on `Hashers:` (or on the first blank line after the section opens), and add a one-line comment recording the actual section order so the next editor doesn't reintroduce the wrong terminator.

---

docs/install.md
```diff
@@ -17,15 +17,68 @@ containers, so a format name would be the wrong thing to install. On a free-thre
...
+
+The tap is not Homebrew core. If you would rather not use it, compile the source
+yourself (needs a C++ compiler — Xcode command-line tools):
+
+```bash
+# Download "UnRAR source" (unrarsrc-*.tar.gz) from https://www.rarlab.com/rar_add.htm
+tar xf unrarsrc-*.tar.gz
+cd unrar
+make
+cp unrar "$(brew --prefix)/bin/"   # or another directory on PATH
```
davitf: ### F6 🟢 `[nit]` · CONFIRMED — this branch exists for people avoiding the tap, then requires Homebrew

```bash
cp unrar "$(brew --prefix)/bin/"   # or another directory on PATH
```

Two small problems:

1. The block is introduced with "If you would rather not use it" — i.e. it is the escape hatch from the third-party tap. `$(brew --prefix)` fails outright for a reader who has no Homebrew at all, which is a plausible reason to be reading this block in the first place.
2. Dropping an unmanaged binary into Homebrew's prefix is something Homebrew actively discourages; `brew doctor` reports it under "Unbrewed files were found in …".

**Fix direction.** Lead with `/usr/local/bin` or `~/.local/bin` plus a note to have it on `PATH`, and keep `$(brew --prefix)/bin` out of this block entirely.

---

dev-docs/investigations/unar-as-rar-decompressor.md
```diff
@@ -0,0 +1,7 @@
+# `unar` as a RAR data backend
+
+Superseded by
```
davitf: ### F7 🟢 `[nit]` · CONFIRMED — this stub is circular, and a tombstone for content that never landed

This PR adds a 7-line file whose only content is "superseded by `alternative-rar-decompressors.md`" — and the superseding document points *back* at it, at `alternative-rar-decompressors.md:24`:

> The earlier unar-only notes live in [`unar-as-rar-decompressor.md`](unar-as-rar-decompressor.md) (pointer).

That is the only inbound link in the tree; I grepped. Neither file exists on `main` — the original was created in `7250bdd` and reduced to this pointer in `b571dae`, both inside this unmerged branch. So on merge, `dev-docs/` gains a file that sends a reader back where they came from, promising "earlier notes" that exist nowhere.

**Fix direction.** Delete this file and drop the pointer paragraph at `alternative-rar-decompressors.md:24`. Per §3 clean-as-you-go — or keep it with a one-line reason if there is an external link to the filename that I can't see from here.

---

tests/test_rar_reader.py
```diff
@@ -688,7 +688,7 @@ def test_non_rarlab_unrar_rejected(
     fake.chmod(0o755)
     monkeypatch.setenv("PATH", str(tmp_path))
     monkeypatch.setattr(rar_unrar, "_cached_unrar", None)
-    with pytest.raises(PackageNotInstalledError, match="RARLAB"):
+    with pytest.raises(PackageNotInstalledError, match=r"RARLAB.*7zz"):
```
davitf: ### F8 🟢 `[nit]` · CONFIRMED — this couples two behaviour tests to the copy of the message

`match=r"RARLAB.*7zz"` passes today — I ran these three tests and they're green, and `_NOT_INSTALLED_MSG` is a single line so `.` matching is not an issue. But it now pins both this test and `test_missing_unrar_raises` (line 712) to `7zz` being *last* in the lookalike list. Reordering or trimming that list breaks two tests that are about **rejection behaviour**, not about copy — and F2 above proposes touching the neighbouring wording, so this is live.

**Fix direction.** Keep `match="RARLAB"` in both of these. If you want the `7zz` mention itself covered, assert it once in a dedicated test over `_NOT_INSTALLED_MSG`, where a copy change correctly shows up as a copy test failing.

---

dev-docs/investigations/alternative-rar-decompressors.md
```diff
@@ -0,0 +1,377 @@
...
+
+**User-docs implication:** pointing at the official macOS *package* is honest
+but a worse first-line than compiling source (Gatekeeper prompt, ARM vs Intel
+URL, trialware `rar` sitting next to `unrar`). Mention it as “exists, may be
+blocked, not the add-ons 2018 link.” Lead pip users at a source compile —
+the unofficial formula or a copy-paste `make`.
+
+---
+
+---
```
davitf: ### F9 🟢 `[nit]` — doubled horizontal rule

Lines 370 and 372 are both `---`, giving two rules in a row before `## Open questions`. Drop one.

(Also worth a look while you're here: per the maintainer's decision the 7z branch is being closed, so this section's heading — "Open questions (only if we reopen 7z)" — and the `unar` disposition both want updating. See F10 in the review body.)

---

davitf (commented):
## Review — RAR decompressor alternatives + macOS unrar install

Reviewed at `4c34632`. Nine findings **F1–F9** are posted inline at their `file:line`; **F10** is below (it has no single anchor). Maintainer decisions are recorded in §Decisions — several were made after the review was drafted, so the inline text already reflects them.

*Verdict: 🔄 **Request changes** on **F1** and **F2** — both are small edits to published user docs. Recorded as a comment rather than a formal Request-Changes because GitHub does not allow that on one's own PR.*

---

### What this change is

After #276 removed Homebrew's `rar` cask, this PR records which other CLIs could supply RAR member data and publishes macOS install guidance for pip users. It adds a 377-line investigation, a 336-line stdlib-only evidence script (`scripts/exploration/rar_decompressor_matrix.py`), a new "Getting RARLAB `unrar`" section in `docs/install.md`, and pointers from `formats.md` / `errors-and-diagnostics.md` / `IDEAS.md`. Behaviour delta is essentially nil: one `PackageNotInstalledError` message gains `7zz`, and two tests tighten their regex to match.

### Snapshot

~816 added / 11 removed across 12 files; medium, but only ~340 lines are code.

**Gates I ran myself** (not copied from CI): `ruff check` + `ruff format --check` clean on the new script; `pytest tests/test_rar_reader.py -k "non_rarlab or missing_unrar or rarlab_build"` → 3 passed; the matrix script runs end-to-end and reproduces the investigation's "distro 7z without `7zip-rar`" column exactly. PR CI green 23/23.

**Caveat on coverage:** my container has no `unar`, `bsdtar`, or `unrar-free`, so I verified the `7z` rows only. The `unar` SIGSEGV / silent-empty results and the `bsdtar` 7 GB measurement are **unreproduced by me** — I am taking them from the write-up.

### Main points

- 🟡 **F1** — `install.md` tells users the brew formula "compiles UnRAR from RARLAB's published source". The tap ships bottles for Sequoia/Sonoma/Ventura, so most users get a prebuilt binary from a third-party GHCR. The investigation in this PR states this correctly; the user page does not.
- 🟡 **F2** — This PR *removes* `unar` and `7z` from the "not a substitute" list in user docs, while the runtime error message expands to name them.
- 🟡 **F3** — The matrix harness compares against an unchecked gold; an empty gold prints `match=yes`. Live in my run: `pwd_wrong:enc` reports `yes` while actually failing on `Unsupported Method`, never reaching the password path.
- 🟡 **F4** — `capture_output=True` with no byte cap; the timeout bounds time, not memory.
- 🟢 **F5–F9** — dead `7z i` section terminator; `$(brew --prefix)` in the no-Homebrew branch; a circular doc stub; copy-coupled test regexes; a doubled `---`.

### What's fine

- Subprocess hygiene in the script is exactly what the addendum §5 asks for: list args, no `shell=True`, `stdin=DEVNULL`, `start_new_session=True`, timeout on every child.
- The `7z_codec_rar5` row is a genuinely good idea — the matrix records which codec configuration produced it, so a re-run is self-describing rather than silently measuring a different machine.
- The `CONTRIBUTING.md` claim checks out: `scripts/setup-dev-env.sh:87` does invoke `install-rarlab-unrar.sh` on Darwin.
- The C1 conclusion — no silent fallback to whatever is on `PATH` — is well evidenced and correctly refuses to widen the backend surface.

---

## F10 🟡 `[important]` — keep `unar` open as a candidate, with concrete next steps

*No single anchor; touches `dev-docs/IDEAS.md:45`, the investigation's §Recommendation and §Open questions, and `dev-docs/known-issues.md`.*

Per the maintainer's decision below, the disposition changes: **close the `7z` branch, keep `unar` open.** The current text closes both, and `IDEAS.md:45` records the whole idea as investigated-and-rejected. Three things follow.

**1. The "is it even installable" question is unanswered.** The investigation measured Debian's `unar 1.10.1` and a locally built 1.10.8, but never establishes whether `brew install unar` works on macOS with RAR support — which is the entire premise of `unar` as the easy-install alternative. I could not check this from here: the egress proxy blocked `formulae.brew.sh`. **Please run it on a Mac and record the result**: whether homebrew-core's `unar` formula installs, what version it reports, and whether it reads the RAR fixtures at all.

**2. The early-fail gate looks cheap and precise — the trigger predicate is pure listing metadata.** I checked what archivey already knows before touching `unrar`:

```
$ archivey → tests/fixtures/rar/basic_solid__.rar
format: ArchiveFormat.RAR | solid: True
  FILE      size=0      empty_file.txt
  FILE      size=13     file1.txt
  FILE      size=16     subdir/file2.txt
  FILE      size=12     implicit_subdir/file3.txt
  DIRECTORY size=0      empty_subdir/
  DIRECTORY size=0      subdir/
```

So `format == RAR and info.is_solid and any(m.type is FILE and m.size == 0)` is decidable from the native parser alone — no `unrar`, no subprocess, no decompression. A `unar` backend could refuse exactly the affected archives at listing time and serve everything else. That makes the pre-fix option more viable than the write-up's flat "no" implies.

**Two caveats before relying on it.** The predicate is generalized from one fixture family (empty first / mid / last on `basic_solid__.rar`); whether other RAR5 solid shapes also trip the decoder — ANTI members, a member whose packed size is nonzero but unpacked is 0 — is untested, so it may be under-inclusive. And RAR4 solid+empty reportedly *works* in `unar`, so a RAR5-only predicate is tempting; being conservative and gating both costs nothing.

**3. The gate is mandatory, not an optimization — and this is a reportable upstream bug.** Note which way the versions run: 1.10.1 **SIGSEGVs** (loud), 1.10.7 returns **rc=0 with empty output** (silent wrong data). The *newer* behaviour is the dangerous one, and it is what a brew user would get. Without the gate, a `unar` backend on current `unar` silently hands callers empty files — a direct hit on VISION's "honest errors" claim, and much worse than crashing.

That belongs in `dev-docs/known-issues.md`, whose stated job is "upstream bugs, our mitigations, and the evidence behind them". The evidence is already gathered (both versions, stdout *and* disk extract, empty-member position varied, skipping empties ruled out as a workaround) — enough to file against MacPaw/XADMaster. Worth an entry there plus an `IDEAS.md` correction from "investigated, do not add" to "blocked on upstream; gate is designed, predicate is listing-only".

---

## Decisions

Recorded from the maintainer, so the implementor does not re-ask:

1. **Lead with the tap in user docs — approved** (it was the maintainer's own suggestion to the implementer), **with the risk stated**. On "is it any worse than a PyPI dep publishing a malicious version?" — broadly the same *kind* of exposure, with two real asymmetries, both in F1: a malicious release of a widely-used PyPI package gets caught fast where a two-star tap has nobody watching; and the *source* path's checksum is verifiable against rarlab.com, a party the tap maintainer does not control, which is actually **stronger** than PyPI, whereas the *bottle* path has no independent check at all. Hence F1: the wording should not imply everyone gets the source path.
2. **`unar` / `7z` get a one-line mention in user docs — restore it.** Justification is `rarfile` compatibility: `rarfile` accepts those backends, so someone migrating has a working setup and no reason to suspect it. That reason should appear in the doc, not just the name. → **F2**.
3. **Docs-review coordination — deferred, no action here.** The docs review resumes later with a new shape in mind (one document per format carrying details, features, limitations, gotchas and detection approaches, as the main reference for future development in place of scattered investigations); `install.md` gets reviewed then.
4. **`7z` — close it.** No gain for a second non-free RAR decompressor under a different package name. → **F10** (update `IDEAS.md` and the investigation's §Recommendation / §Open questions accordingly).
5. **`unar` — keep open** instead, with the brew check, the early-fail gate, and the upstream report. → **F10**.

---

docs/install.md
davitf: ### F1 correction — the source/bottle trust distinction is weaker than I wrote

Maintainer pushed back: gromgit can change the `sha256` in the formula at any time. That's right, and it undercuts the reasoning above. Correcting it, because the fix direction I gave was wrong in a way that would put a false claim into user docs.

**What I got wrong.** I said the source path's checksum is "checkable against rarlab.com, a party the tap maintainer does not control" and called it "a *stronger* guarantee than any PyPI dependency." But the `sha256` lives *in the formula*, and the maintainer owns the formula — so a malicious update changes `url` and `sha256` together. Homebrew's verification protects against transport/CDN tampering and against rarlab.com swapping a tarball out from under a fixed pin. It does not protect against the tap maintainer. **Both paths root in trusting gromgit.**

**Corrected ranking**, weakest first:

1. **The brew tap, either path** — no consumer-side pin at all. `brew install` takes whatever the tap's HEAD formula says at install time, and again on every `brew upgrade`. (`brew pin` only freezes an already-installed version; there is no lockfile equivalent that pre-commits to a hash.)
2. **A hashed PyPI dep in a lockfile** (`uv.lock`, `--require-hashes`) — the pin lives on the consumer's side, so upstream publishing something new does not change what you install.
3. **`pmachapman/unrar@d861246` via `scripts/install-rarlab-unrar.sh`** — a content-addressed commit pin we control. Strongest, and what CI already does.

So the tap is *weaker* than a normal pinned dependency, not stronger. That is the same reasoning the investigation already used to keep the tap out of CI ("a tap can change the URL tomorrow; CI would inherit that") — it applies to users too, just with a different risk appetite.

**What actually survives** is auditability, not verification: a malicious source-path change is a public commit whose `sha256` won't match what rarlab.com currently serves, so it is falsifiable by anyone who looks. A malicious bottle is an opaque binary with no upstream reference to diff against — unfalsifiable without reproducible builds. Real, but that is "someone could catch it after the fact", not "you are protected".

**Revised fix direction.** The factual correction in F1 stands unchanged — the page must stop implying everyone compiles from source, because most readers get a bottle. What changes is how to frame the alternative:

- **Do not** sell `--build-from-source` as a "compile-it-yourself guarantee". It removes the opaque-binary problem, not the trust-gromgit problem.
- State the actual cost of the one-command install: an unofficial tap, trusted at install time and on every `brew upgrade`. That is the honest version of the decision to lead with it — which stands.
- The contrast worth drawing for a reader who wants more assurance is the **copy-paste source build** further down the page (download from rarlab.com yourself, `make`), where no third party is in the path at all. That is the real "no tap trust" option, not `--build-from-source`.

---

davitf:
## Review addendum — F10 narrowed, F2 context, F1 corrected

Three updates to [the review](https://github.com/davitf/archivey/pull/278#pullrequestreview-5082839787), after maintainer feedback.

### F1 — corrected in the thread

The source-vs-bottle trust argument in F1 was wrong: gromgit controls the formula, therefore controls both `url` and `sha256`, so both paths root in trusting the tap. The factual finding stands (the page implies everyone compiles from source; most readers get a bottle), but the fix direction changed — `--build-from-source` must **not** be described as a compile-it-yourself guarantee. Details and the corrected ranking are in [the F1 thread](https://github.com/davitf/archivey/pull/278#discussion_r3908536020).

### F10 — narrowed: install is confirmed, version and fixture read are not

Maintainer has tested on a Mac and reports it installed fine, so F10's first question is answered — **don't redo it**. Two things it does not settle:

1. **Which `unar` version?** `unar -v` (or `lsar -v`). This is the crux, not a detail: 1.10.1 **SIGSEGVs** (loud, safe-ish), 1.10.7 returns **rc=0 with empty output** (silent wrong data). If Homebrew ships the silent one, a `unar` backend without the early-fail gate hands callers empty files with no error — which is the VISION "honest errors" violation, not a performance nit.
2. **Does it actually read the fixtures?** Compare against `unrar p` on `tests/fixtures/rar/basic_solid__.rar` (should fail) and a non-solid one (should match).

Cheapest way to get both: run this PR's own script on the Mac — it picks `unar` up from `PATH` and fills in the rows.

```bash
python3 scripts/exploration/rar_decompressor_matrix.py
```

Worth also confirming *which* tool was installed — `brew install unar` (XADMaster, the candidate backend) vs `brew install gromgit/new-life/unrar` (RARLAB UnRAR, the docs recommendation). The investigation already records a user confirmation for the latter; F10 is asking about the former.

### F10 — no, none of this is covered by CI, and won't be

Checked `ci.yml` at the PR base (`6492cdb`):

- macOS installs RARLAB unrar from the pinned source build — `scripts/install-rarlab-unrar.sh` into `$RUNNER_TEMP/rarlab-unrar`, cached on the script's hash. It never runs `brew install unar`, and never installs the `gromgit/new-life` tap.
- The only occurrences of `unar` in the workflow are two comments. There is no job that would exercise it.
- Nor is there anything to exercise: archivey has no `unar` backend, and `find_rarlab_unrar()` rejects non-RARLAB banners by design.

So **both** macOS install paths this PR publishes — the tap for users, and `unar` as a future candidate — are hand-verified once and untested thereafter. That is normal for install instructions, but worth stating plainly: if the tap changes its URL, drops a bottle for a new macOS, or disappears, nothing in CI turns red. The pinned `install-rarlab-unrar.sh` path is the only one with a guard, and it guards the contributor/CI route, not the documented user route.

Not proposing CI coverage here — a tap-install job would be a macOS-only moving part depending on a third party, which is the same objection #276 already sustained. Flagging it so the gap is a known one rather than an assumed one.

### F2 — note that it reverses an earlier decision on this PR

F2 asks to restore `unar` / `7z` to the user-docs negative list. That reverses [an earlier comment on this PR](https://github.com/davitf/archivey/pull/278#discussion_r3907733873) (now resolved and collapsed) which asked for exactly the opposite — "we don't need to mention 7z and unar here… they belong in dev docs only" — and which the current diff implements.

The reversal is deliberate and the maintainer's own, with a new justification: `rarfile` accepts those backends, so someone migrating from `rarfile` arrives with one on `PATH` and no reason to suspect it. Recording it here so whoever implements F2 doesn't read the collapsed thread as a contradictory instruction and split the difference.

---

PR #279, opened 2026-09-01
Validate 7z and RAR SFX scan hits (prefixed-archive-detection follow-up)

davitf:
Slim follow-up after #277 (`prefixed-archive-detection` Block 1). **7z/RAR hit validators only** — tasks **2.3** (partial), **2.4**, and the scan-validation pins in **4.4**. Not the rest of Block 3.

## What this PR does

#277's shebang cue searches only formats that have an `SFX_HIT_VALIDATOR`. ZIP already had one; 7z and RAR now join that seam automatically (`entry.format in validators`). A script that merely *mentions* those magics still raises `FormatDetectionError`. A shebang in front of a **real** 7z or RAR now detects.

- **7z** (`validate_sevenzip_signature_header`): `StartHeaderCRC` over the 20-byte StartHeader. Failed CRC / oversized next-header / declared end past known remaining → `DAMAGED`. Empty next-header behind a stub → `NOT_THIS_FORMAT`. Trailing bytes stay `VALID`.
- **RAR** (`validate_rar_main_header`): RAR 5 first-block CRC32 (MAIN or ENCRYPTION); RAR 4 parseable MAIN with matching 16-bit CRC. Shares `load_vint` and `rar3_main_crc_end` with the parser.
- The scan is still first-`VALID`. Exact-EOF ranking among several CRC-valid 7z hits is **deferred** (D2(a) withdrawn).
- Validators live on the format module; `detection.py` still does not inline parse. The scan always passes `remaining` (required; `None` is a value meaning unknown length).
- A short `peek_more` after `remaining` already proves the bytes exist is a scan-window clamp, not `NOT_THIS_FORMAT` (ZIP name/extra, RAR header).

## Review (`9b1a7aa`)

D1(b) and D3(a) unchanged. **D2 is (b)**. Re-review nits F7 and F13 closed:

- **F1 / F2:** `remaining` is candidate-relative source length; no `peek_more(declared)`.
- **F3:** post-CRC rejects are `DAMAGED`.
- **F4:** empty-SFX reject shipped. Inexact CRC-valid decoy is an `xfail(strict=True)` red half. Exact-end ranking deferred.
- **F5 / F6:** aligned / shared parser helpers.
- **F7:** short peek is a clamp, not a reject, once `remaining` proves the bytes exist (ZIP name/extra and RAR header).
- **F8–F11:** fixed earlier.
- **F12:** task 2.3 is `[~]`.
- **F13:** `remaining` has no default.

## Out of scope (later blocks)

- ZIP tail probe, `prefix_kind`, `ArchiveyConfig.detection_budget`, exhaustive scan, makeself / compressor needles.
- Exact-EOF multi-hit ranking (task 2.3 remainder) — follow-up after this PR.
- OpenSpec archive stays for the finishing PR (task 4.12). Task 4.10 is left unchecked because later blocks still run it.

## Gates

`./scripts/check.sh --fix` clean. `./scripts/test.sh --all-configs` all three green:

- `[all]` / `[all-lowest]`: 2674 passed / 23 skipped / 1 xfailed
- `[core-only]`: 2161 passed / 479 skipped / 1 xfailed

GitHub CI on `9b1a7aa`: 23/23 green.

_Generated by [Cursor](https://cursor.com)_

<div><a href="https://cursor.com/agents/bc-20174351-08c2-49ce-85ee-0111cb248b34?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-20174351-08c2-49ce-85ee-0111cb248b34&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/openspec/changes/prefixed-archive-detection/tasks.md
+++ b/openspec/changes/prefixed-archive-detection/tasks.md
@@ -11,6 +11,8 @@ use `(context)` / `(every block)` rather than a block number.
    hit validator (`entry.format in validators`); 2.3–2.4 self-enable on that
    seam. No `prefix_kind`, no tail probe, no compressor needles.
    Tasks: **2.1, 2.1a, 2.1b, 2.1c, 2.2, 2.2a, 3.4, 4.1, 4.1a, 4.8a, 5.4.**
+   **Slim follow-up after #277:** 7z `StartHeaderCRC` and RAR main-header CRC
+   on `SFX_HIT_VALIDATOR` (tasks **2.3, 2.4, 4.4**). Not the rest of Block 3.
 2. **ZIP tail probe** — budget-gated (`THOROUGH.max_tail_bytes`, not `BALANCED`).
    Workspace tail helper. Both ZIP offset conventions. JPEG+ZIP as an enabled-path
    test. EOCD + central-directory confirmation of a scan hit can fold in here.
@@ -157,16 +159,22 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
       backend (candidate-relative view in, `HitOutcome` out); the scan calls it. A
       decoy of four `PK\x03\x04` bytes is `NOT_THIS_FORMAT`. Do **not** require an
       EOCD seek here — that is Block 2. Red–green alongside 4.1a.
-- [ ] 2.3 **(Block 3, or a slim follow-up after Block 1)** Validate a 7z hit: `StartHeaderCRC` over the 20-byte StartHeader,
+- [x] 2.3 **(Block 3, or a slim follow-up after Block 1)** Validate a 7z hit: `StartHeaderCRC` over the 20-byte StartHeader,
       and `offset + 32 + NextHeaderOffset + NextHeaderSize <= source length`, preferring
       an exact end match when several candidates validate — an intra-format tie-break
       inside this validator, not the cross-format ranking 0.6 defers. Named function on
       the 7z backend returning `HitOutcome`; the scan calls it. Failed CRC with a
       plausible header is `DAMAGED`; this change still skips it (scan continues). Ledger
       task 4.7 is the later policy that reports `SEVEN_Z` from that outcome.
-- [ ] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
+      **Landed as the slim follow-up after #277.** Exact-EOF among several CRC-valid
+      hits is still earliest-VALID (the generic first-match loop); trailing bytes after
+      a CRC-valid header stay `VALID` (task 4.4). A declared end past `SFX_MAX` is a
+      large archive, not an overrun — CRC passing is enough; do not grow the prefix.
+- [x] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
       parseable main header. Named function on the RAR backend, same `HitOutcome` split
       as 2.3.
+      **Landed as the slim follow-up after #277.** RAR 5 encrypted-headers archives
+      start with a plaintext ENCRYPTION block; that CRC is the identity check.
 - [ ] 2.5 **(Block 3)** Continue scanning past a candidate that fails validation rather than giving up
 - [x] 2.5a ~~**Prerequisite for 2.6–2.8: a bounded candidate-relative read.**~~ — shipped by `#273` as `PrefixWorkspace.peek_range` / `candidate_view` and `ScanNeedle` / `MagicHit.candidate_origin`. Nothing to invent here.
 - [ ] 2.5b **(Block 4)** Report `payload_offset` as the **candidate origin**, not the needle hit. A TAR hit reported at `H` rather than `H - 257` is wrong by 257 bytes and hands the backend a misaligned source; add a red–green for exactly that off-by-257. **Waits on 2.6**, which waits on the ledger.
@@ -219,7 +227,7 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
 - [x] 4.1a **(Block 1)** Red–green the ZIP scan validator: a `#!` stub whose text contains `PK\x03\x04` but whose following bytes fail local-header sanity SHALL raise `FormatDetectionError`, not report `ZIP` / `CorruptionError`. Pin the ELF-cued form of the same bytes too — today that path already claims a damaged ZIP, which is the live defect Block 1 must not widen.
 - [ ] 4.2 **(Block 4)** Red–green: a makeself-style `#!/bin/sh` + tar.gz detects as `TAR_GZ` and opens its members. Cover the negatives that define the scoping too: an `MZ` stub containing the same gzip bytes yields no scan claim, a `#!` stub whose own text contains `1f 8b 08` is rejected by the **structural header check**, and a `#!` + non-tar gzip stream reports `GZIP` (inner-TAR is resolution, not identity)
 - [ ] 4.3 **(Block 3)** SFX matrix across stub kinds: PE, ELF (a real `rar a -sfx`), Mach-O, and shebang, for 7z and RAR. Needs Block 1's cue already landed.
-- [ ] 4.4 **(Block 3)** Scan validation: the 6 magic bytes embedded in unrelated data are not claimed; a 7z whose declared end overruns the source is not claimed; a 7z with trailing bytes appended still is
+- [x] 4.4 **(Block 3)** Scan validation: the 6 magic bytes embedded in unrelated data are not claimed; a 7z whose declared end overruns the source is not claimed; a 7z with trailing bytes appended still is. **Landed with the 2.3–2.4 slim follow-up after #277.**
 - [ ] 4.5 **(Block 2)** Non-seekable prefixed ZIP falls through rather than crashing or buffering the stream
 - [ ] 4.6 **(Block 3)** Exhaustive scan: off by default leaves a beyond-window archive undetected and unread past the window; on, it finds it with `prefix_kind == UNKNOWN`
 - [ ] 4.7 **(Block 3)** `prefix_kind` values for each fixture in 4.1–4.3
--- a/openspec/changes/prefixed-archive-detection/tasks.md
+++ b/openspec/changes/prefixed-archive-detection/tasks.md
@@ -169,12 +169,14 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
       **Landed as the slim follow-up after #277.** The declared-end check compares
       against the known source length (`remaining` from the candidate origin), not the
       peek window / `scan_limit` / `SFX_MAX`. An unreachable end or oversized next-header
-      after a passing CRC is `DAMAGED`. Exact-EOF among several CRC-valid hits is still
-      earliest-VALID (the generic first-match loop) — deferred: a genuine SFX with a
-      trailing config blob has no exact match, so the tie-break is a preference among
-      validated hits, not a filter, and needs this remaining plumbing first. Trailing
-      bytes after a CRC-valid header stay `VALID` (task 4.4). The pin
-      `test_crc_valid_empty_7z_decoy_still_wins_over_a_later_payload` records the hole.
+      after a passing CRC is `DAMAGED`. Exact-EOF among several CRC-valid hits is
+      `VALID_EXACT` (7z-only: declared end equals known remaining); the scan returns
+      immediately on that outcome and otherwise remembers the earliest `VALID`.
+      `NextHeaderSize == 0` behind a stub is `NOT_THIS_FORMAT`. Trailing bytes after a
+      CRC-valid header stay `VALID` (task 4.4). Pins:
+      `test_inexact_7z_decoy_loses_to_a_later_exact_payload` (the tie-break) and
+      `test_inexact_7z_decoy_still_wins_when_the_real_payload_has_trailing_bytes`
+      (the residual).
 - [x] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
       parseable main header. Named function on the RAR backend, same `HitOutcome` split
       as 2.3.
--- a/openspec/changes/prefixed-archive-detection/tasks.md
+++ b/openspec/changes/prefixed-archive-detection/tasks.md
@@ -166,10 +166,15 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
       the 7z backend returning `HitOutcome`; the scan calls it. Failed CRC with a
       plausible header is `DAMAGED`; this change still skips it (scan continues). Ledger
       task 4.7 is the later policy that reports `SEVEN_Z` from that outcome.
-      **Landed as the slim follow-up after #277.** Exact-EOF among several CRC-valid
-      hits is still earliest-VALID (the generic first-match loop); trailing bytes after
-      a CRC-valid header stay `VALID` (task 4.4). A declared end past `SFX_MAX` is a
-      large archive, not an overrun — CRC passing is enough; do not grow the prefix.
+      **Landed as the slim follow-up after #277.** The declared-end check compares
+      against the known source length (`remaining` from the candidate origin), not the
+      peek window / `scan_limit` / `SFX_MAX`. An unreachable end or oversized next-header
+      after a passing CRC is `DAMAGED`. Exact-EOF among several CRC-valid hits is still
+      earliest-VALID (the generic first-match loop) — deferred: a genuine SFX with a
+      trailing config blob has no exact match, so the tie-break is a preference among
+      validated hits, not a filter, and needs this remaining plumbing first. Trailing
+      bytes after a CRC-valid header stay `VALID` (task 4.4). The pin
+      `test_crc_valid_empty_7z_decoy_still_wins_over_a_later_payload` records the hole.
 - [x] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
       parseable main header. Named function on the RAR backend, same `HitOutcome` split
       as 2.3.
--- a/openspec/changes/prefixed-archive-detection/tasks.md
+++ b/openspec/changes/prefixed-archive-detection/tasks.md
@@ -159,24 +159,23 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
       backend (candidate-relative view in, `HitOutcome` out); the scan calls it. A
       decoy of four `PK\x03\x04` bytes is `NOT_THIS_FORMAT`. Do **not** require an
       EOCD seek here — that is Block 2. Red–green alongside 4.1a.
-- [x] 2.3 **(Block 3, or a slim follow-up after Block 1)** Validate a 7z hit: `StartHeaderCRC` over the 20-byte StartHeader,
+- [~] 2.3 **(Block 3, or a slim follow-up after Block 1)** Validate a 7z hit: `StartHeaderCRC` over the 20-byte StartHeader,
       and `offset + 32 + NextHeaderOffset + NextHeaderSize <= source length`, preferring
       an exact end match when several candidates validate — an intra-format tie-break
       inside this validator, not the cross-format ranking 0.6 defers. Named function on
       the 7z backend returning `HitOutcome`; the scan calls it. Failed CRC with a
       plausible header is `DAMAGED`; this change still skips it (scan continues). Ledger
       task 4.7 is the later policy that reports `SEVEN_Z` from that outcome.
-      **Landed as the slim follow-up after #277.** The declared-end check compares
-      against the known source length (`remaining` from the candidate origin), not the
-      peek window / `scan_limit` / `SFX_MAX`. An unreachable end or oversized next-header
-      after a passing CRC is `DAMAGED`. Exact-EOF among several CRC-valid hits is
-      `VALID_EXACT` (7z-only: declared end equals known remaining); the scan returns
-      immediately on that outcome and otherwise remembers the earliest `VALID`.
-      `NextHeaderSize == 0` behind a stub is `NOT_THIS_FORMAT`. Trailing bytes after a
-      CRC-valid header stay `VALID` (task 4.4). Pins:
-      `test_inexact_7z_decoy_loses_to_a_later_exact_payload` (the tie-break) and
-      `test_inexact_7z_decoy_still_wins_when_the_real_payload_has_trailing_bytes`
-      (the residual).
+      **Partially landed as the slim follow-up after #277 (F12).** CRC identity, the
+      remaining-length overrun (`DAMAGED`), and `NextHeaderSize == 0` behind a stub
+      (`NOT_THIS_FORMAT`) are in. The declared-end check compares against the known
+      source length (`remaining` from the candidate origin), not the peek window /
+      `scan_limit` / `SFX_MAX`. Exact-EOF ranking among several CRC-valid hits is
+      **not**: first `VALID` still wins. Trailing bytes after a CRC-valid header stay
+      `VALID` (task 4.4). Pin: `test_inexact_7z_decoy_loses_to_a_later_exact_payload`
+      (`xfail(strict=True)` on the real payload winning). Recorded in
+      `dev-docs/known-issues.md`. `[~]` so this change cannot archive until the
+      tie-break lands (`scripts/check_openspec_archived.py`).
 - [x] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
       parseable main header. Named function on the RAR backend, same `HitOutcome` split
       as 2.3.
--- a/src/archivey/internal/rar_detect.py
+++ b/src/archivey/internal/rar_detect.py
@@ -15,8 +15,8 @@ from archivey.internal.backends.rar_parser import (
     RAR5_ID,
     RAR_ID,
     _crc32,
-    _load_vint,
-    _rar3_main_crc_end,
+    load_vint,
+    rar3_main_crc_end,
 )
 from archivey.internal.sfx import HitOutcome
 
@@ -32,14 +32,6 @@ _RAR5_ENCRYPTION = 4
 _RAR5_HEADER_CAP = 64 * 1024
 
 
-def _try_vint(buf: bytes, pos: int) -> tuple[int, int] | None:
-    """Parser vint, or ``None`` if truncated / too long (does not raise)."""
-    try:
-        return _load_vint(buf, pos)
-    except CorruptionError:
-        return None
-
-
 def validate_rar_main_header(
     peek_more: Callable[[int], bytes],
     remaining: int | None = None,
@@ -52,6 +44,10 @@ def validate_rar_main_header(
     the first block's CRC32 to match; a plausible header whose CRC fails is
     :attr:`HitOutcome.DAMAGED`. RAR 4 requires a parseable MAIN block (type
     ``0x73``) with a matching 16-bit header CRC.
+
+    ``peek_more`` stays outside the parse ``try`` so a workspace ``OSError``
+    propagates. A truncated vint before the CRC is ``NOT_THIS_FORMAT``; after
+    the CRC has matched, a later vint failure is ``DAMAGED``.
     """
     del remaining
     head = peek_more(len(RAR5_ID))
@@ -68,26 +64,28 @@ def _validate_rar5(peek_more: Callable[[int], bytes]) -> HitOutcome:
     if len(prefix) < len(RAR5_ID) + 5:
         return HitOutcome.NOT_THIS_FORMAT
     body = prefix[len(RAR5_ID) :]
-    loaded = _try_vint(body, 4)
-    if loaded is None:
+    try:
+        hdrlen, pos = load_vint(body, 4)
+    except CorruptionError:
         return HitOutcome.NOT_THIS_FORMAT
-    hdrlen, pos = loaded
     if hdrlen > _RAR5_HEADER_CAP:
         return HitOutcome.NOT_THIS_FORMAT
     header_size = pos + hdrlen
     hdata = peek_more(len(RAR5_ID) + header_size)[len(RAR5_ID) :]
     if len(hdata) < header_size:
         return HitOutcome.NOT_THIS_FORMAT
-    header_crc = int.from_bytes(hdata[:4], "little")
-    if header_crc != _crc32(memoryview(hdata)[4:]):
-        return HitOutcome.DAMAGED
-    block = _try_vint(hdata, pos)
-    if block is None:
-        return HitOutcome.NOT_THIS_FORMAT
-    block_type, _pos = block
-    if block_type not in (_RAR5_MAIN, _RAR5_ENCRYPTION):
-        return HitOutcome.NOT_THIS_FORMAT
-    return HitOutcome.VALID
+    identity_held = False
+    try:
+        header_crc = int.from_bytes(hdata[:4], "little")
+        if header_crc != _crc32(memoryview(hdata)[4:]):
+            return HitOutcome.DAMAGED
+        identity_held = True
+        block_type, _pos = load_vint(hdata, pos)
+        if block_type not in (_RAR5_MAIN, _RAR5_ENCRYPTION):
+            return HitOutcome.NOT_THIS_FORMAT
+        return HitOutcome.VALID
+    except CorruptionError:
+        return HitOutcome.DAMAGED if identity_held else HitOutcome.NOT_THIS_FORMAT
 
 
 def _validate_rar3(peek_more: Callable[[int], bytes]) -> HitOutcome:
@@ -103,7 +101,7 @@ def _validate_rar3(peek_more: Callable[[int], bytes]) -> HitOutcome:
     hdata = peek_more(len(RAR_ID) + header_size)[len(RAR_ID) :]
     if len(hdata) < header_size:
         return HitOutcome.NOT_THIS_FORMAT
-    crc_pos = _rar3_main_crc_end(flags)
+    crc_pos = rar3_main_crc_end(flags)
     if crc_pos > header_size:
         return HitOutcome.NOT_THIS_FORMAT
     calc = _crc32(hdata[2:crc_pos]) & 0xFFFF
--- a/src/archivey/internal/rar_detect.py
+++ b/src/archivey/internal/rar_detect.py
@@ -1,0 +1,133 @@
+"""Cheap RAR main-header identity check for the SFX scan.
+
+Seven or eight bytes of ``Rar!\\x1a\\x07`` in a stub are not a RAR. The cued
+scan calls :func:`validate_rar_main_header` with a candidate-relative view.
+"""
+
+from __future__ import annotations
+
+import struct
+import zlib
+from collections.abc import Callable
+
+from archivey.internal.backends.rar_parser import RAR5_ID, RAR_ID
+from archivey.internal.sfx import HitOutcome
+
+# RAR3 MAIN block type and the 7-byte block header (crc16, type, flags, size).
+_RAR3_MAIN = 0x73
+_RAR3_BLK_HDR = struct.Struct("<HBHH")
+# MAIN has 6 reserved bytes after the block header; LONG_BLOCK inserts a 4-byte
+# add_size before those (same walk as the parser); ENCRYPTVER adds one more
+# before the CRC coverage ends. Real MAIN headers are tens of bytes.
+_RAR3_MAIN_MIN = _RAR3_BLK_HDR.size + 6
+_RAR3_HEADER_CAP = 64 * 1024
+_RAR3_MAIN_ENCRYPTVER = 0x0200
+_RAR3_LONG_BLOCK = 0x8000
+
+# RAR5 MAIN / ENCRYPTION (encrypted-headers archive: first block is ENCRYPTION).
+_RAR5_MAIN = 1
+_RAR5_ENCRYPTION = 4
+# Identity only — a MAIN/ENCRYPTION header is small. Cap so a hostile vint
+# cannot force a 2 MiB peek (the F1 analogue).
+_RAR5_HEADER_CAP = 64 * 1024
+
+
+def _crc32(data: bytes | memoryview) -> int:
+    return zlib.crc32(data) & 0xFFFFFFFF
+
+
+def _load_vint(buf: bytes, pos: int) -> tuple[int, int] | None:
+    """RAR5 vint, or ``None`` if truncated / too long."""
+    length = len(buf)
+    if pos >= length:
+        return None
+    b = buf[pos]
+    if b < 0x80:
+        return b, pos + 1
+    limit = pos + 11
+    if limit > length:
+        limit = length
+    res = b & 0x7F
+    ofs = 7
+    pos += 1
+    while pos < limit:
+        b = buf[pos]
+        res += (b & 0x7F) << ofs
+        pos += 1
+        ofs += 7
+        if b < 0x80:
+            return res, pos
+    return None
+
+
+def validate_rar_main_header(peek_more: Callable[[int], bytes]) -> HitOutcome:
+    """Return whether a candidate origin looks like a RAR 4 or RAR 5 archive.
+
+    ``peek_more(n)`` is a view relative to the candidate. A truncated or
+    non-RAR magic is :attr:`HitOutcome.NOT_THIS_FORMAT`. RAR 5 requires the
+    first block's CRC32 to match; a plausible header whose CRC fails is
+    :attr:`HitOutcome.DAMAGED`. RAR 4 requires a parseable MAIN block (type
+    ``0x73``) with a matching 16-bit header CRC.
+    """
+    head = peek_more(len(RAR5_ID))
+    if head.startswith(RAR5_ID):
+        return _validate_rar5(peek_more)
+    if head.startswith(RAR_ID):
+        return _validate_rar3(peek_more)
+    return HitOutcome.NOT_THIS_FORMAT
+
+
+def _validate_rar5(peek_more: Callable[[int], bytes]) -> HitOutcome:
+    # Marker + CRC + a generous vint prefix, then the declared body.
+    prefix = peek_more(len(RAR5_ID) + 4 + 16)
+    if len(prefix) < len(RAR5_ID) + 5:
+        return HitOutcome.NOT_THIS_FORMAT
+    body = prefix[len(RAR5_ID) :]
+    loaded = _load_vint(body, 4)
+    if loaded is None:
+        return HitOutcome.NOT_THIS_FORMAT
+    hdrlen, pos = loaded
+    if hdrlen > _RAR5_HEADER_CAP:
+        return HitOutcome.NOT_THIS_FORMAT
+    header_size = pos + hdrlen
+    hdata = peek_more(len(RAR5_ID) + header_size)[len(RAR5_ID) :]
+    if len(hdata) < header_size:
+        return HitOutcome.NOT_THIS_FORMAT
+    header_crc = int.from_bytes(hdata[:4], "little")
+    if header_crc != _crc32(memoryview(hdata)[4:]):
+        return HitOutcome.DAMAGED
+    block = _load_vint(hdata, pos)
+    if block is None:
+        return HitOutcome.NOT_THIS_FORMAT
+    block_type, _pos = block
+    if block_type not in (_RAR5_MAIN, _RAR5_ENCRYPTION):
+        return HitOutcome.NOT_THIS_FORMAT
+    return HitOutcome.VALID
+
+
+def _validate_rar3(peek_more: Callable[[int], bytes]) -> HitOutcome:
+    start = peek_more(len(RAR_ID) + _RAR3_BLK_HDR.size)
+    if len(start) < len(RAR_ID) + _RAR3_BLK_HDR.size:
+        return HitOutcome.NOT_THIS_FORMAT
+    buf = start[len(RAR_ID) :]
+    header_crc, block_type, flags, header_size = _RAR3_BLK_HDR.unpack_from(buf)
+    if block_type != _RAR3_MAIN:
+        return HitOutcome.NOT_THIS_FORMAT
+    if header_size < _RAR3_MAIN_MIN or header_size > _RAR3_HEADER_CAP:
+        return HitOutcome.NOT_THIS_FORMAT
+    hdata = peek_more(len(RAR_ID) + header_size)[len(RAR_ID) :]
+    if len(hdata) < header_size:
+        return HitOutcome.NOT_THIS_FORMAT
+    # Mirror rar_parser._parse_rar3's MAIN CRC coverage, including LONG_BLOCK.
+    crc_pos = _RAR3_BLK_HDR.size
+    if flags & _RAR3_LONG_BLOCK:
+        crc_pos += 4
+    crc_pos += 6
+    if flags & _RAR3_MAIN_ENCRYPTVER:
+        crc_pos += 1
+    if crc_pos > header_size:
+        return HitOutcome.NOT_THIS_FORMAT
+    calc = _crc32(hdata[2:crc_pos]) & 0xFFFF
+    if header_crc != calc:
+        return HitOutcome.DAMAGED
+    return HitOutcome.VALID
--- a/src/archivey/internal/rar_detect.py
+++ b/src/archivey/internal/rar_detect.py
@@ -6,69 +6,54 @@ scan calls :func:`validate_rar_main_header` with a candidate-relative view.
 
 from __future__ import annotations
 
-import struct
-import zlib
 from collections.abc import Callable
 
-from archivey.internal.backends.rar_parser import RAR5_ID, RAR_ID
+from archivey.exceptions import CorruptionError
+from archivey.internal.backends.rar_parser import (
+    _RAR3_MAIN,
+    _S_BLK_HDR,
+    RAR5_ID,
+    RAR_ID,
+    _crc32,
+    _load_vint,
+    _rar3_main_crc_end,
+)
 from archivey.internal.sfx import HitOutcome
 
-# RAR3 MAIN block type and the 7-byte block header (crc16, type, flags, size).
-_RAR3_MAIN = 0x73
-_RAR3_BLK_HDR = struct.Struct("<HBHH")
-# MAIN has 6 reserved bytes after the block header; LONG_BLOCK inserts a 4-byte
-# add_size before those (same walk as the parser); ENCRYPTVER adds one more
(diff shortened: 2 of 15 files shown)


src/archivey/internal/sevenzip_detect.py
```diff
@@ -0,0 +1,70 @@
...
+        return HitOutcome.NOT_THIS_FORMAT
+    declared = _SIGNATURE_HEADER_SIZE + next_header_offset + next_header_size
+    if declared < _SIGNATURE_HEADER_SIZE:
+        return HitOutcome.NOT_THIS_FORMAT
+    # Only confirm the declared end when it sits inside the scan window. A
+    # next-header past SFX_MAX is a large archive, not an overrun we can see.
+    if declared <= SFX_MAX:
+        got = peek_more(declared)
+        if len(got) < declared:
+            return HitOutcome.NOT_THIS_FORMAT
```
davitf: ### F1 🔴 `[blocking]` · CONFIRMED — the declared-end check uses the peek window, not the source length

`peek_more` here comes from `workspace.candidate_view(hit.candidate_origin, limit=scan_limit)` (`detection.py:465`), and that view clamps to `scan_limit - candidate_origin` — an **archive-origin** ceiling (`detection_workspace.py:231-236`). So the predicate this code actually enforces is not `declared <= SFX_MAX`, it is `candidate_origin + declared <= scan_limit`. The stub's size, which says nothing about whether the payload is a 7z, decides the answer.

Task 2.3 says `offset + 32 + NextHeaderOffset + NextHeaderSize <= source length`. That is the right rule and it needs no bytes: `PrefixWorkspace.remaining_known()` returns the exact source length for a path or seekable stream, and `None` when unknown.

**Trigger** (default `BALANCED` budget, `scan_limit == SFX_MAX == 2 MiB`):

```python
def sig(o, s):
    st = struct.pack("<QQI", o, s, 0)
    return MAGIC_7Z + b"\x00\x00" + (zlib.crc32(st) & 0xFFFFFFFF).to_bytes(4, "little") + st

stub = b"\x7fELF\x02\x01\x01" + b"\x00" * (262144 - 7)
path.write_bytes(stub + sig(1_900_000, 64) + b"\x00" * (1_900_000 + 64))
detect_format(path)
```

Observed:

| case | declared end | stub + declared | result |
|---|---|---|---|
| 256 KiB stub, 1.86 MiB payload | 1,945,696 | 2,207,840 | **`FormatDetectionError`** |
| 256 KiB stub, 1.66 MiB payload | 1,740,896 | 2,003,040 | `SEVEN_Z` @ 262144 |
| 4 KiB stub, same 1.86 MiB payload | 1,945,696 | 1,949,792 | `SEVEN_Z` @ 4096 |

Three ways this is worse than a narrow edge case:

1. **It is a regression.** Setting `SevenZipReadBackend.SFX_HIT_VALIDATOR = None` (pre-PR behaviour) detects the failing file at offset 262144. A 170–500 KB `7zS.sfx` stub in front of a 1.5–1.9 MB payload is an ordinary Windows SFX.
2. **The budget becomes a correctness gate.** `FAST_BUDGET.max_scan_bytes` is 256 KiB, so `scan_limit` is 256 KiB and *every* 7z SFX whose declared end exceeds that is rejected. Confirmed: a 4 KiB stub + 1.5 MiB payload detects under `BALANCED`, raises `FormatDetectionError` under `FAST`. Budgets are meant to be cost gates.
3. **Detector and parser now disagree.** On the same failing file, `find_signature_offset(fp)` returns 262144 and forced `format=SEVEN_Z` opens it. `sfx.py`'s module docstring names exactly this — "a stub the RAR parser accepts but the detector does not would be a file that opens under `format=RAR` and fails under auto-detect" — as the thing the shared bound exists to prevent.

The user-facing error is also dishonest: `"Could not detect archive format: no magic-byte match and no usable file extension"`, when the scan found a CRC-valid 7z signature and discarded it on budget. The `BUDGET_EXHAUSTED` skip lands on the receipt, but the receipt is not the message.

**Fix direction.** Do the length check against the source, not the view. The validator signature (`Callable[[int], bytes]`) has no way to learn the source length, so either (a) `_scan_for_sfx_payload` does the `candidate_origin + declared <= remaining_known()` check after the validator returns, or (b) validators get a second argument carrying `remaining_known()` (`None` = unknown). Either way: "the end is not visible" must not be a reject. See decision packet **D1** in the review body — this one is the maintainer's call on shape, not yours.

Red–green candidate: pin the 256 KiB-stub file above at both `BALANCED` and `FAST`, and pin that the same payload behind a 4 KiB stub gives the same answer.

---

src/archivey/internal/sevenzip_detect.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    if next_header_size > _MAX_NEXT_HEADER_SIZE:
+        return HitOutcome.NOT_THIS_FORMAT
+    declared = _SIGNATURE_HEADER_SIZE + next_header_offset + next_header_size
+    if declared < _SIGNATURE_HEADER_SIZE:
+        return HitOutcome.NOT_THIS_FORMAT
+    # Only confirm the declared end when it sits inside the scan window. A
+    # next-header past SFX_MAX is a large archive, not an overrun we can see.
+    if declared <= SFX_MAX:
+        got = peek_more(declared)
```
davitf: ### F2 🟡 `[important]` · CONFIRMED — confirming the declared end buffers the whole archive

`peek_more(declared)` grows the monotone prefix buffer to the archive's declared end, so detection reads the entire payload just to check that bytes exist there. Measured on a 1.5 MB SFX 7z with the needle at offset 4096:

```
with validator   : SEVEN_Z  unique_bytes_read = 1,504,192
without validator: SEVEN_Z  unique_bytes_read =    65,536
```

23× read amplification on the detection path, before any member is listed. The RAR validator next door peeks at most 64 KiB — that asymmetry is the tell.

Same root cause as **F1**, and the F1 fix removes this finding entirely: comparing two integers reads nothing.

---

src/archivey/internal/sevenzip_detect.py
```diff
@@ -0,0 +1,70 @@
...
+        return HitOutcome.NOT_THIS_FORMAT
+    start_header_crc = int.from_bytes(header[8:12], "little")
+    start_header = header[12:32]
+    if _crc32(start_header) != start_header_crc:
+        return HitOutcome.DAMAGED
+    next_header_offset, next_header_size, _next_header_crc = struct.unpack(
+        "<QQI", start_header
+    )
+    if next_header_size > _MAX_NEXT_HEADER_SIZE:
+        return HitOutcome.NOT_THIS_FORMAT
```
davitf: ### F3 🟡 `[important]` · CONFIRMED — `NOT_THIS_FORMAT` after the CRC has already established identity

Both post-CRC rejects in this function return `NOT_THIS_FORMAT` — this one, and the declared-end short read at line 69. `HitOutcome`'s own docstring (`sfx.py:81-86`) defines that as "identity never held (decoy magic, unparseable header)" and `DAMAGED` as "identity holds, structure does not". A 32-byte header whose `StartHeaderCRC` matches **is** identity — a 1-in-2³² accident. A truncated real 7z SFX, and a 7z with a corrupt next-header size, both get filed here as "never was a 7z".

Today the scan treats both outcomes the same, so nothing observable changes. It stops being free at ledger tasks 4.6/4.7, which report `SEVEN_Z` from `DAMAGED` — that is a lost identification, and it cuts against VISION's "damaged input is first-class".

Contrast `_validate_rar5` in `rar_detect.py`, which gets the ordering right: its `_RAR5_HEADER_CAP` check runs *before* the CRC, so `NOT_THIS_FORMAT` there is accurate.

**Fix direction.** After `_crc32(start_header) == start_header_crc` succeeds, every subsequent reject is `DAMAGED`. Add both cases to the validator unit test.

---

tests/test_sfx.py
```diff
@@ -616,32 +622,17 @@ def test_a_stub_carrying_a_decoy_zip_header_does_not_move_the_answer(
...
-
-    with pytest.raises(CorruptionError):
-        open_archive(path)
-
-    # The real payload is reachable the moment something supplies the right offset.
-    with _open_with(path, start_offset=len(stub)) as archive:
-        assert {m.name for m in archive.members() if m.is_file} == set(_FILES)
+    path = tmp_path / "decoy-7z.exe"
+    path.write_bytes(stub + _7z_bytes(tmp_path))
+    _assert_sfx_opens(path, ArchiveFormat.SEVEN_Z, len(stub))
```
davitf: ### F4 🟡 `[important]` · CONFIRMED — this rewrite drops a pin on a hole that is still open

The old test pinned "earliest match wins, even against a real payload", and its docstring said that was acceptable *only* because the failure was loud. The rewrite narrows the decoy to six bare magic bytes — which the validator does now reject — and deletes the pin. The limitation survives in a narrower, forgeable form: a CRC-valid decoy still wins, because the scan takes the earliest `VALID` hit.

The decoy does not even need crafting. `sig(next_offset=0, next_size=0)` is a **valid, well-formed empty 7z archive** — 32 bytes. Trigger:

```python
stub = b"MZ" + b"\x00" * 510 + sig(0, 0) + b"\x00" * 3000
path.write_bytes(stub + Path("tests/fixtures/sevenzip/lz4.7z").read_bytes())
```

Observed: `detect_format` reports `SEVEN_Z` at offset 512 (the decoy; the real payload is at 3544), and `open_archive` returns **zero members**. The only signal is the `"Archive listed no members (SEVEN_Z)"` diagnostic — a warned wrong answer, not a raised one, against a change whose charter carries "no silent wrong answer" as its own requirement (`archive/2026-08-21-sfx-format-detection/brief.md`, MD2 = A). Confirmed separately that those 32 bytes open standalone as a valid empty archive.

To be fair on attribution: this behaviour is **pre-existing**, identical before the PR, and the CRC does raise the bar from "six bytes" to "a well-formed header". The problems are that the deferral removed the mitigation — task 2.3's "preferring an exact end match when several candidates validate" — and the test that recorded the limitation, in the same diff.

**Fix direction.** Keep a pin, with a CRC-valid decoy instead of a bare needle, asserting today's behaviour and naming it as the known limit — the shape the old docstring had. If the exact-end tie-break stays deferred, record *why* in `tasks.md`, not just *that*. See decision packet **D2** in the review body.

---

openspec/changes/prefixed-archive-detection/tasks.md
```diff
@@ -157,16 +159,22 @@ block. **0.4** is `(context)` — a prerequisite note, not a block.
...
       an exact end match when several candidates validate — an intra-format tie-break
       inside this validator, not the cross-format ranking 0.6 defers. Named function on
       the 7z backend returning `HitOutcome`; the scan calls it. Failed CRC with a
       plausible header is `DAMAGED`; this change still skips it (scan continues). Ledger
       task 4.7 is the later policy that reports `SEVEN_Z` from that outcome.
-- [ ] 2.4 **(Block 3, or a slim follow-up after Block 1)** Validate a RAR 5 hit via the main header's CRC32; RAR 4 via a
+      **Landed as the slim follow-up after #277.** Exact-EOF among several CRC-valid
+      hits is still earliest-VALID (the generic first-match loop); trailing bytes after
+      a CRC-valid header stay `VALID` (task 4.4). A declared end past `SFX_MAX` is a
+      large archive, not an overrun — CRC passing is enough; do not grow the prefix.
```
davitf: ### F5 🟡 `[important]` · CONFIRMED — spec, task amendment, and code state three different rules

- Task 2.3's original text (line 163, unchanged): `... <= source length`
- The amendment this PR adds (line 171): "A declared end past `SFX_MAX` is a large archive, not an overrun"
- The code (`sevenzip_detect.py:64-69`): `candidate_origin + declared <= scan_limit`

All three differ. The amendment reads as documenting the implementation, but it describes a bound the code does not use: `SFX_MAX` is a constant, `scan_limit` is `min(SFX_MAX, budget.max_scan_bytes)` and the comparison is measured from the *archive* origin, not the candidate. The same claim is repeated in the module docstring at `sevenzip_detect.py:37-40`.

Whichever way **F1** lands, this text and that docstring have to state the rule the code enforces. Per addendum §3, a spec that turns out wrong should be revised deliberately — narrating it into agreement is the failure mode.

---

src/archivey/internal/rar_detect.py
```diff
@@ -0,0 +1,133 @@
...
+    # Mirror rar_parser._parse_rar3's MAIN CRC coverage, including LONG_BLOCK.
+    crc_pos = _RAR3_BLK_HDR.size
+    if flags & _RAR3_LONG_BLOCK:
+        crc_pos += 4
+    crc_pos += 6
+    if flags & _RAR3_MAIN_ENCRYPTVER:
+        crc_pos += 1
+    if crc_pos > header_size:
+        return HitOutcome.NOT_THIS_FORMAT
+    calc = _crc32(hdata[2:crc_pos]) & 0xFFFF
```
davitf: ### F6 🟡 `[important]` · CONFIRMED — parser logic and constants copied into the detect modules

One root cause, four sites. The comment on line 121 acknowledges the mirroring; that is the problem, not the mitigation.

| copy | original |
|---|---|
| `rar_detect.py:121-130` RAR3 MAIN CRC walk | `rar_parser.py:934-948` |
| `rar_detect.py:39-60` `_load_vint` | `rar_parser.py:451-473` |
| `rar_detect.py:35`, `sevenzip_detect.py:24` `_crc32` | `rar_parser.py:417` |
| `sevenzip_detect.py:16` `_SIGNATURE_HEADER_SIZE`, `:19` `_MAX_NEXT_HEADER_SIZE` | `sevenzip_parser.py` (`_MAX_NEXT_HEADER_SIZE` at `:62`) |

Both new modules already import from those parser modules (`RAR5_ID` / `RAR_ID`, `MAGIC_7Z`), so there is no zero-dep or import-cost reason for the copies.

The drift is silent and one-directional: raise `_MAX_NEXT_HEADER_SIZE` in the parser and the detector quietly stops claiming archives the parser would open; change `_parse_rar3`'s flag walk and the validator starts reporting `DAMAGED` on healthy RAR4s. Nothing in the suite fails when the two sides diverge. Addendum §3 (zero tech debt): pay it now or record it as keep-with-reason.

`_load_vint` is the one copy with a real justification — returning `None` instead of raising `CorruptionError`. That can stay a thin wrapper over the parser's version rather than a reimplementation, with the comment saying why. See decision packet **D3**.

---

src/archivey/internal/sevenzip_detect.py
```diff
@@ -0,0 +1,70 @@
...
+    declared = _SIGNATURE_HEADER_SIZE + next_header_offset + next_header_size
+    if declared < _SIGNATURE_HEADER_SIZE:
+        return HitOutcome.NOT_THIS_FORMAT
```
davitf: ### F8 🟢 `[nit]` · CONFIRMED — unreachable guard

`declared = 32 + next_header_offset + next_header_size`, both fields unsigned from `"<QQI"`, and Python ints do not wrap — so `declared < _SIGNATURE_HEADER_SIZE` cannot be true.

If it is guarding against a future signed unpack or a refactor that makes these bounded ints, say so in a comment; otherwise drop it.

---

src/archivey/internal/sevenzip_detect.py
```diff
@@ -0,0 +1,70 @@
...
+    ``peek_more(n)`` is a view relative to the candidate, not the source. A
+    truncated or non-7z magic is :attr:`HitOutcome.NOT_THIS_FORMAT`. A 7z
+    magic whose ``StartHeaderCRC`` fails is :attr:`HitOutcome.DAMAGED` — identity
+    held, structure did not. The scan treats that like ``NOT_THIS_FORMAT``
+    (skip, continue); the evidence-ledger scheduler may later report it.
+
+    ``offset + 32 + NextHeaderOffset + NextHeaderSize`` must fall at or before
+    the end of the *view*. A declared end past :data:`~archivey.internal.sfx.SFX_MAX`
+    is a real archive whose next header sits outside the scan window — CRC
+    passing is enough; do not grow the prefix to chase it (the F1 analogue).
```
davitf: ### F9 🟢 `[nit]` — unresolvable finding reference in a shipped comment

"(the F1 analogue)" here, and the same phrase at `rar_detect.py:31`. `F1` is a PR-review ID with no anchor anywhere in the tree — `grep -rn '\bF1\b'` finds a dozen unrelated `F1`s across `review/` and `openspec/`, none of them this one. A future editor sees the tree, not the review.

Both sentences already carry the local *why* ("do not grow the prefix to chase it" / "a hostile vint cannot force a 2 MiB peek"), so the ID adds nothing. Drop it, or qualify it as `#277 F1` so it is at least resolvable.

---

tests/test_sfx.py
```diff
@@ -745,6 +736,83 @@ def _header(
...
+    )
+    huge = _sevenzip_signature(next_size=(64 << 20) + 1)
+    assert (
+        validate_sevenzip_signature_header(_peek_view(huge))
+        is HitOutcome.NOT_THIS_FORMAT
+    )
+    # Trailing bytes after a CRC-valid header are still this format (task 4.4).
+    trailing = _sevenzip_signature() + b"config after payload\n"
+    assert validate_sevenzip_signature_header(_peek_view(trailing)) is HitOutcome.VALID
+
```
davitf: ### F10 🟢 `[nit]` — one test, ten assertions

This packs real-fixture, bare-magic, synthetic-valid, wrong-version, CRC-damaged, declared-end-overrun, oversized-next-header and trailing-bytes cases into one function. The first failure hides the rest, and the name can only describe two of them.

Parametrize over `(payload, expected_outcome)` — the cases are already written as independent one-liners, so it is close to a mechanical change, and it makes **F3**'s `DAMAGED` reclassification a two-line diff instead of a re-read.

`test_rar_main_header_validator_accepts_real_archives_and_rejects_a_crc_fail` (line 797) has the same shape, milder.

---

tests/test_sfx.py
```diff
@@ -913,11 +1054,10 @@ def blob(header_at: int) -> bytes:
...
 def test_sfx_hit_validator_is_not_bound_on_instance() -> None:
-    inst = object.__new__(ZipReadBackend)
-    assert inst.SFX_HIT_VALIDATOR is ZipReadBackend.SFX_HIT_VALIDATOR
-    assert (
-        inst.SFX_HIT_VALIDATOR(_peek_view(b"PK\x03\x04")) is HitOutcome.NOT_THIS_FORMAT
-    )
+    for cls in (ZipReadBackend, SevenZipReadBackend, RarReadBackend):
+        inst = object.__new__(cls)
+        assert inst.SFX_HIT_VALIDATOR is cls.SFX_HIT_VALIDATOR
+        assert inst.SFX_HIT_VALIDATOR(_peek_view(b"")) is HitOutcome.NOT_THIS_FORMAT
```
davitf: ### F11 🟢 `[nit]` — the binding test lost its ZIP-specific assertion

Widening to all three backends is right. But the payload went from `b"PK\x03\x04"` to `b""`, and empty input is the trivial case for every validator — the second assertion now proves only that each one tolerates a zero-length peek.

The binding property (the point of the test) is still pinned by the first assertion. Cheap fix: put a per-class decoy payload in the loop table, so ZIP keeps `b"PK\x03\x04"`, 7z gets `MAGIC_7Z`, and RAR gets `RAR_ID`.

---

davitf (commented):
## Maintainer briefing

**What this change is.** This PR gives 7z and RAR the same scan-hit validator ZIP already had, so the SFX scan stops claiming a file just because six magic bytes appear in a stub. Two new modules (`sevenzip_detect.py`, `rar_detect.py`) check a 7z `StartHeaderCRC` and a RAR 4/5 main-header CRC; the backends register them on `SFX_HIT_VALIDATOR`, which also auto-enables the shebang path for both formats. Behaviour delta: a script that merely mentions those magics no longer detects; a shebang or stub in front of a **real** 7z/RAR now does.

**Snapshot.** ~400 added / 39 deleted across 9 files, small-medium. **Verdict: 🔄 Request Changes** — posted as a `COMMENT` review because GitHub refuses `REQUEST_CHANGES` on a PR from the same account.

**What I ran** (head `60bc90e`): read all 23 check runs off the PR — all green (lint, pyrefly/ty, three dependency configs, PPMd stress, Atheris, docs). Locally, `./scripts/test.sh tests/test_sfx.py` on `[all]` → **63 passed**. I did **not** re-run the full suite or the other two configs; everything below comes from targeted repro scripts run against this head with the project venv, and each finding shows what was run.

**Main points**

- **F1 🔴** The 7z declared-end check measures against the *peek window*, not the source length as task 2.3 specifies. A real 7z SFX is rejected whenever `stub + archive > scan_limit` — reproduced: a 256 KiB stub + 1.86 MiB payload now raises `FormatDetectionError` where it detected before this PR. Under the `FAST` budget every 7z SFX over 256 KiB fails.
- **F2 🟡** The same check reads the whole archive to confirm its end: 65 KiB → 1.5 MiB unique bytes for one 1.5 MB SFX. A length comparison costs zero bytes.
- **F3 🟡** After the CRC passes, identity has held — but an unreachable end or an oversized next-header returns `NOT_THIS_FORMAT`, not `DAMAGED`. That distinction is what the ledger's tasks 4.6/4.7 are built on.
- **F4 🟡** The rewritten decoy test drops a pin on a hole that still exists: a CRC-valid 32-byte decoy (which is exactly a valid *empty* 7z) still wins over a real payload, and `open_archive` returns zero members. Pre-existing, not introduced here — but task 2.3's exact-end tie-break, the specified mitigation, is what this PR defers.
- **F6 🟡** `_load_vint`, `_crc32`, the RAR3 CRC walk and two 7z constants are hand-copied from parser modules the new files already import.

**What's fine.** The RAR validator is genuinely cheap and correct — `VALID` on all 29 RAR fixtures including both RAR4 and RAR5 encrypted-headers, peek bounded to 64 KiB, and its LONG_BLOCK/ENCRYPTVER walk mirrors `_parse_rar3` exactly. Module placement follows the `zip_detect.py` precedent, `detection.py` still does not parse, and importing either new module pulls in no third-party package (checked against `sys.modules`). The `staticmethod` wrapping plus the widened binding test is the right guard.

Findings **F1–F6, F8–F11** are inline on the diff. **F7** has no line in this diff:

### F7 🟢 `[nit]` · CONFIRMED — same short-read-is-a-reject shape in the ZIP validator

`src/archivey/internal/zip_detect.py:91-94` — `peek_more(30 + name_len + extra_len)` short → `NOT_THIS_FORMAT`, with the same inability to tell a budget clamp from EOF that **F1** describes. Bounded to ~128 KiB (both lengths are 16-bit), so it only bites for a candidate within 128 KiB of `scan_limit` — much narrower than F1, and listed so the F1 fix covers both sites rather than converting one of two.

---

## Maintainer decisions

### D1 — How should a validator learn the source length?

1. **Question.** F1's fix needs the declared-end check to compare against the source length instead of how far the peek will go. Should the scan do that check, or should validators receive the length?
2. **Why it matters.** As written, a real 7z SFX with a 256 KiB stub and a 1.86 MiB payload is undetectable at the default budget and detectable behind a 4 KiB stub, and every 7z SFX over 256 KiB is undetectable under `FAST`. It also splits auto-detect from forced `format=SEVEN_Z`, which opens the same file fine.
3. **Options.**
   - **(a) Scan-side check.** `_scan_for_sfx_payload` asks the validator for the declared end and compares it to `workspace.remaining_known()`. Cost: `HitOutcome` alone cannot carry a number, so either a second hook or a small result type. Keeps `detection.py` free of parsing but gives it an arithmetic contract.
   - **(b) Second validator argument.** `validator(peek_more, remaining)` where `remaining` is `remaining_known()` or `None`. Cost: touches the `SFX_HIT_VALIDATOR` type, ZIP, and the base-class comment — about a dozen lines. Fixes F7 in the same edit. Keeps the check where the format knowledge is.
   - **(c) Drop the declared-end check; rely on `StartHeaderCRC`.** Cost: task 4.4's "declared end overruns the source is not claimed" no longer holds, and `test_mz_7z_declared_end_overrun_is_not_claimed` has to go. Cheapest, and loses the least in practice given the CRC — but it walks back a task this PR just marked done.
4. **Evidence.** `detection.py:465` passes `limit=scan_limit`; `detection_workspace.py:231-236` clamps with `remaining = max(0, limit - candidate_origin)`. Repro table in F1, run at both budget presets. `find_signature_offset` returns 262144 on the file auto-detect rejects. Pre-PR behaviour recovered by nulling `SFX_HIT_VALIDATOR`.
5. **Recommendation.** **(b)** — the only option that puts the check where task 2.3 wrote it, kills F2's 23× read amplification, and fixes F7 without a second pass.
6. **Default if you ignore this.** 7z SFX files whose stub plus payload exceed the scan budget stop being detected, silently and with a misleading error, and every `FAST`-budget caller loses 7z SFX detection above 256 KiB.

### D2 — Does the crafted-decoy hole block this PR?

1. **Question.** A CRC-valid 32-byte decoy — a valid empty 7z — still beats a real payload behind it, and `open_archive` returns zero members with only a diagnostic. Does that get fixed here (task 2.3's exact-end tie-break) or recorded and deferred?
2. **Why it matters.** The parent change's charter makes "no silent wrong answer" its own requirement (MD2 = A). This PR deferred the specified mitigation and, in the same diff, rewrote the test that recorded the limitation — so nothing in the tree says it is still open.
3. **Options.**
   - **(a) Ship the tie-break now.** Prefer a candidate whose declared end matches the source end. Cost: needs D1's length plumbing first; a genuine SFX with a trailing config blob has no exact match, so it must be a preference among validated hits, not a filter. Closes the hole.
   - **(b) Defer, but keep the pin.** Restore a test with a CRC-valid decoy asserting today's behaviour, and write the reason for the deferral into `tasks.md`. Cost: ~15 lines; the hole stays open and documented.
   - **(c) Defer silently.** What the PR does now.
4. **Evidence.** Repro in F4: `MZ` stub + `sig(0, 0)` + `tests/fixtures/sevenzip/lz4.7z` → detected at 512, `open_archive` lists zero members, warning `"Archive listed no members (SEVEN_Z)"`. Those 32 bytes open standalone as a valid empty archive. Same result before the PR — this is not a regression.
5. **Recommendation.** **(b)** for this PR, since (a) depends on D1 landing first. Not (c): the pin is what keeps the hole visible after the PR body scrolls away.
6. **Default if you ignore this.** The limitation stays real and nothing in the tree records it — the rewritten test now reads as if decoys are handled.

### D3 — Pay the parser-duplication debt now or record it?

1. **Question.** `_load_vint`, `_crc32`, the RAR3 CRC walk, and two 7z constants are copies of parser code the new modules already import. Paid in this PR or recorded as keep-with-reason?
2. **Why it matters.** Drift is silent and one-directional: the parser gets more permissive, the detector quietly stops claiming archives the parser opens — the exact detector/parser split F1 already shows one instance of.
3. **Options.**
   - **(a) Import what exists now.** Constants directly; a thin non-raising wrapper over `rar_parser._load_vint`; a shared helper for the RAR3 MAIN CRC coverage. Cost: ~30 lines, touches `rar_parser`'s private surface.
   - **(b) Record it.** A note in `review/backlog.md` or `IDEAS.md` saying the copies are deliberate and why. Cost: none, debt stays.
4. **Evidence.** Table in F6. `rar_detect.py:121` already says "Mirror `rar_parser._parse_rar3`'s MAIN CRC coverage" — the drift risk is acknowledged in the comment and not otherwise addressed. No test fails when the two sides diverge.
5. **Recommendation.** **(a)** for the two 7z constants and the RAR3 walk (mechanical, and the walk is the one most likely to drift). `_load_vint` can stay forked — the non-raising return is a real difference — if the comment says that is why.
6. **Default if you ignore this.** Four copies of parser rules that must be edited in lockstep, with nothing that fails when they are not.

---

davitf:
## Maintainer decisions — resolved

Answers to D1–D3 from the review above, with what each one requires from the implementor. Finding IDs are unchanged; nothing here supersedes a finding.

| | Decision | Findings it closes |
|---|---|---|
| **D1** | **(b)** second validator argument carrying the source length | F1, F2, F7 |
| **D2** | **(a)** ship the exact-end tie-break now — **plus** (b)'s pin, see below | F4 (partly) |
| **D3** | **(a)** import what exists; no duplicated code | F6 |

---

### D1 = (b) — validators take the length

`validator(peek_more, remaining)`.

`remaining` is **candidate-relative**: the scan passes `remaining_known() - candidate_origin`, because `PrefixWorkspace.remaining_known()` is measured from the archive origin while the validator's whole world (`peek_more`) is candidate-relative. Mixing the two is how F1 happened in the first place.

`None` means "source length unknown" (non-seekable, abandoned spool) and must **skip** the length check, never fail it. That is the whole of F1: "the end is not visible" is not evidence against the format.

Then in `validate_sevenzip_signature_header`, `declared > remaining` replaces the `peek_more(declared)` round-trip. No bytes are read, which is F2, and `zip_detect.py:91-94` gets the same treatment for F7.

### D2 = (a) — and no, D1 does not cover it

Asked in the packet: is "needs D1's length plumbing" satisfied by choosing D1(b)? **Not on its own.** D1(b) supplies the length; the tie-break needs two more things.

**1. A way to express "exact end".** A validator returns `HitOutcome` for *one* candidate, so it cannot compare candidates — "inside this validator" (task 2.3) can only mean it reports a stronger outcome. That needs a fourth member, `VALID_EXACT`, or `VALID` plus a separate exactness signal. `design.md`'s warning is against inventing a parallel *confidence scale*; one more member on the enum the ledger already wraps is not that, but it does want a line in `design.md` since ledger tasks 3.3 / 4.6 / 4.7 read this enum.

**2. A scan loop that does not stop at the first hit.** `_scan_for_sfx_payload` currently `break`s on the first `VALID`. New rule:

- `VALID_EXACT` → return immediately (nothing can beat it)
- `VALID` → remember the earliest one, keep iterating
- end of window → fall back to the remembered `VALID`

**This does not undo F2.** Measured on archives written by the system `7z`:

```
out_m0Copy.7z     declared=200166  actual=200166  exact=True
out_m0LZMA2.7z    declared=200188  actual=200188  exact=True
out_mheonpfoo.7z  declared=200272  actual=200272  exact=True   (-mhe=on)
```

A real 7z ends exactly at its declared end, so an ordinary SFX returns `VALID_EXACT` on the first candidate and still early-exits. The full-window scan only happens for a payload with trailing bytes after it — the case `test_mz_7z_with_trailing_bytes_is_still_claimed` covers.

**It does fix the reproduced decoy.** Running the scan over F4's file and computing both numbers per candidate:

```
candidate@  512  declared=     32  remaining=   3832  exact=False   <- the empty-7z decoy
candidate@ 3544  declared=    800  remaining=    800  exact=True    <- the real payload
```

**Two limits to write down rather than discover later:**

- **Only 7z can produce the signal.** Neither ZIP's local header nor RAR's main header declares a total size — ZIP's end lives in the EOCD, which is Block 2's tail probe. So `VALID_EXACT` is 7z-only for now, and its docstring should say *"the format could prove the payload ends at EOF"*, not *"this candidate is better"*. Otherwise the next reader files it as a format-parity hole (addendum §2).
- **It cannot fire when `remaining is None`.** On a non-seekable source, and on any archive with trailing bytes, both candidates are plain `VALID` and earliest still wins.

So **take D2(b)'s pin as well, not instead**: a test that asserts the tie-break fixes the CRC-valid decoy *and* pins the residual case, so F4's hole stays visible where it survives.

**Cheap complement, one line, worth having anyway:** `next_header_size == 0` at a *nonzero* candidate origin is an empty archive behind a stub — nobody ships a self-extractor with no files. Rejecting that kills F4's specific decoy without any ranking at all, and it works on non-seekable sources where the tie-break cannot.

### D3 = (a) — the try/except is the better half of the idea

> we should have one parsing code that either raises or returns None, and maybe have the other wrap it. (can't rar detection just be surrounded with a try/except CorruptionError?)

Yes, and skip the wrapper. Import the raising `_load_vint` from `rar_parser` and catch `CorruptionError` in the detector. A `None`-returning variant that the parser then wraps would put an extra frame on **30 call sites** in `rar_parser`, on a function whose own comment calls it a hot path ("most RAR5 vints are a single byte … listing many-member archives"). The detector is the cold caller; it should absorb the cost.

Three things to keep right:

- **Preserve F3's split.** One `try` around the whole function collapses it. Set `identity_held = True` immediately after the CRC match and have the handler return `DAMAGED` if set, `NOT_THIS_FORMAT` otherwise.
- **Leave `peek_more` outside the `try`.** An `OSError` from the workspace must propagate (CONTRIBUTING); it will, since `except CorruptionError` does not catch it — just do not widen the clause to `Exception`.
- **Drop the underscore.** With a second in-tree caller, `_load_vint` becomes `load_vint` rather than a cross-module private import.

**try/except only solves the vint.** The other three copies in F6 still need doing:

- `_SIGNATURE_HEADER_SIZE` / `_MAX_NEXT_HEADER_SIZE` → import from `sevenzip_parser`
- `_crc32` → import once instead of two re-declarations
- **the RAR3 MAIN CRC walk** → extract from `_parse_rar3` (`rar_parser.py:934-948`) into a shared helper both sides call, e.g. `rar3_main_crc_end(flags) -> int`. This is the highest-drift copy of the four, it is duplicated *logic* rather than a duplicated call, and it is the one the try/except cannot touch.

---

### One scoping note

D2(a) is the largest of the three — a new enum member plus multi-candidate ranking in the scan — and it lands on top of D1's signature change in the same PR. If this gets long, the clean split is **D1 + D3 here** (they close F1, F2, F3, F6, F7 and are mechanical once the signature moves) and **D2(a) as its own follow-up**, with D2(b)'s pin landing in this PR so F4's hole is recorded in the tree the moment the rewritten test ships. Your call; both are coherent.

---

src/archivey/internal/sevenzip_detect.py
davitf: **Maintainer decision: D1 = (b)** — validators take a second argument.

`validator(peek_more, remaining)`, where `remaining` is **candidate-relative**: the scan passes `remaining_known() - candidate_origin`, since `remaining_known()` is archive-origin-relative and `peek_more` is not. `None` (non-seekable, abandoned spool) **skips** the length check rather than failing it.

So this line becomes `declared > remaining` → reject, with no `peek_more(declared)` round-trip — which also closes **F2**, and the same treatment closes **F7** at `zip_detect.py:91-94`.

Full resolution, including why D1(b) alone does not cover D2: https://github.com/davitf/archivey/pull/279#issuecomment-5510268822

---

tests/test_sfx.py
davitf: **Maintainer decision: D2 = (a)** — ship task 2.3's exact-end tie-break now. Take **(b)'s pin as well**, not instead.

D1(b) supplies the length but not the tie-break. Two more pieces are needed: a fourth `HitOutcome` (`VALID_EXACT`), because a validator sees one candidate and cannot compare; and a scan loop that returns immediately on `VALID_EXACT` but otherwise remembers the earliest `VALID` and keeps iterating instead of `break`ing.

It fixes the decoy in this thread — running the scan over that file and computing both numbers per candidate:

```
candidate@  512  declared=     32  remaining=   3832  exact=False   <- the empty-7z decoy
candidate@ 3544  declared=    800  remaining=    800  exact=True    <- the real payload
```

It does **not** cost F2 back: archives written by the system `7z` (Copy, LZMA2, `-mhe=on`) all end exactly at their declared end, so an ordinary SFX returns `VALID_EXACT` on the first candidate and still early-exits.

It also cannot fire when `remaining is None` (non-seekable) or when the real payload has trailing bytes — both candidates are plain `VALID` and earliest wins again. That is why the pin stays: assert the fix **and** pin the residual.

Rationale and limits: https://github.com/davitf/archivey/pull/279#issuecomment-5510268822

---

src/archivey/internal/rar_detect.py
davitf: **Maintainer decision: D3 = (a)** — no duplicated code. For the vint, import the raising `_load_vint` from `rar_parser` and wrap the detector body in `try / except CorruptionError`; skip the `None`-returning wrapper, since that would add a frame to **30 call sites** in `rar_parser` on a function its own comment calls a hot path. The detector is the cold caller.

Three things to keep right:

- **Preserve F3's split** — one `try` around everything collapses it. Set `identity_held = True` right after the CRC match; the handler returns `DAMAGED` if set, `NOT_THIS_FORMAT` otherwise.
- **Leave `peek_more` outside the `try`** — an `OSError` from the workspace must propagate, and will, as long as the clause is not widened to `Exception`.
- **Drop the underscore** — with a second in-tree caller it is `load_vint`, not a cross-module private import.

The try/except only covers the vint. Still outstanding from the table above: `_SIGNATURE_HEADER_SIZE` / `_MAX_NEXT_HEADER_SIZE` imported from `sevenzip_parser`, `_crc32` imported once instead of twice, and the RAR3 MAIN CRC walk extracted from `_parse_rar3` (`rar_parser.py:934-948`) into a shared helper both sides call. That walk is duplicated *logic* rather than a duplicated call, and it is the highest-drift copy of the four.

Full resolution: https://github.com/davitf/archivey/pull/279#issuecomment-5510268822

---

davitf:
## D2 revised: (a) deferred, (b) now — and (b) must not assert the wrong behaviour

Raised by the maintainer: *doesn't a test asserting the wrong behaviour become dangerous?* Yes. Agreed, and this repo has already run the experiment.

`test_a_decoy_needle_in_the_stub_wins_and_fails_loudly` was exactly that shape — a plain assertion of the wrong answer (`payload_offset == 514`), with a careful docstring explaining it was a known scanner limit and naming the conditions under which to rewrite it. It survived until the first PR that touched the area, and that PR **deleted** it rather than narrowing it. A green test asserting wrong behaviour reads as the contract; that is how it gets removed instead of fixed. Do not rebuild it.

### Pin the right behaviour and let it fail — the repo already has a name for this

`@pytest.mark.xfail(strict=True)` on an assertion of what *should* happen: the real payload at 3544 is the answer, not the decoy at 512.

`tests/test_review_simplicity_consistency.py` calls these **red halves** — *"assertions of the behaviour the review argued was correct, failing on purpose until a fix landed"*. All sixteen of them eventually landed. Every danger inverts:

| plain assertion of current behaviour | `xfail(strict=True)` on the desired behaviour |
|---|---|
| the file states the wrong contract | the file states the **right** contract |
| reports green — indistinguishable from correct | reports **xfail** — visible as outstanding in every run |
| fixing the bug turns a green test red; looks like a regression | fixing the bug turns it **XPASS → suite failure**, forcing the marker's removal |
| deletable in one PR with no signal | cannot be silently fixed *or* silently rotted |

`xfail_strict` is **not** set in `pyproject.toml`, so `strict=True` has to be on the marker itself — without it, an XPASS passes quietly and the mechanism is lost.

### The enforcing marker is the task ledger, not the test — F12

**F12 🟡 `[important]` · CONFIRMED — task 2.3 is marked `[x]` for work that is not done.**
`openspec/changes/prefixed-archive-detection/tasks.md:163-172`

Task 2.3's text includes *"preferring an exact end match when several candidates validate — an intra-format tie-break inside this validator"*. With D2(a) deferred that clause is unimplemented, so `[x]` is a false completion claim — and it is the claim CI reads.

`scripts/check_openspec_archived.py` decides finished-ness from the checkboxes and documents a third state for exactly this: *"the `[~]` partial marker … which this deliberately does not treat as done."* Mark 2.3 **`[~]`** (or split the tie-break out as its own unchecked `2.3a`). `prefixed-archive-detection` then cannot be archived while it stands, on every PR and on `main`. That is a hard gate rather than a comment, and it is the real "this needs fixing" mechanism — the test only makes the bug visible; the ledger is what stops the change from being declared done around it.

### And one line in `dev-docs/known-issues.md`

Same species as the `unar` RAR5 entry already there: a silent-wrong-answer hazard recorded with its trigger and the gate that has not been written yet. It is the only one of the three markers a person browsing docs will hit.

Three markers, three half-lives: **xfail** fails loudly the moment the behaviour is fixed, **`[~]`** blocks archiving until then, **known-issues.md** is what a human reads. The first two are load-bearing.

### Cheap partial fix worth taking in this PR

With (a) deferred, the one-liner from the previous comment is worth more, not less: reject `next_header_size == 0` at a **nonzero** candidate origin. An empty archive behind a stub is not a thing anyone ships, and it is precisely the decoy in F4.

Verified it costs nothing for genuine empty archives — a real empty `.7z` is claimed by near magic, never by the scan:

```
empty .7z at offset 0 -> ArchiveFormat.SEVEN_Z by magic offset 0
```

That closes the reproduced case on **all** sources, including the non-seekable ones where the tie-break could never fire, and it narrows the residual to a decoy with a nonzero, self-consistent declared size that fits inside `remaining`. The red half still holds the general case until (a) lands.

### So D2(b) is four things, not one

1. Red half: `xfail(strict=True)` asserting the real payload wins over a CRC-valid decoy.
2. Task 2.3 → `[~]` (**F12**).
3. A `dev-docs/known-issues.md` entry.
4. Optional but recommended: reject a zero-size next header at a nonzero candidate origin.

Nothing here changes D1 or D3.

---

tests/test_sfx.py
davitf: **Revised: D2(a) deferred, D2(b) lands here — and the pin must not assert today's wrong behaviour.**

The test this rewrite replaced was that exact shape (a green assertion of `payload_offset == 514` plus a docstring explaining the limit), and it was deleted rather than narrowed by the first PR to touch the area. Rebuilding it repeats the failure.

Use a **red half** instead — the pattern already in `tests/test_review_simplicity_consistency.py`: `@pytest.mark.xfail(strict=True)` asserting that the real payload at 3544 wins over the decoy at 512. The file then states the correct contract, reports as xfail rather than green, and `strict=True` turns the day D2(a) lands into an XPASS suite failure that forces the marker out. `xfail_strict` is not set in `pyproject.toml`, so `strict=True` must be on the marker.

Also worth taking here, one line and independent of the tie-break: reject `next_header_size == 0` at a nonzero candidate origin. That closes this thread's reproduced decoy on all sources, including non-seekable ones. Confirmed it costs nothing for real empty archives — `empty .7z at offset 0 -> SEVEN_Z by magic offset 0`, so the scan never sees them.

Full shape, plus new finding **F12** (task 2.3 is marked `[x]` for the unimplemented tie-break, and `[~]` is what blocks archiving): https://github.com/davitf/archivey/pull/279#issuecomment-5510385917

---

src/archivey/internal/zip_detect.py
```diff
@@ -89,6 +94,8 @@ def validate_zip_local_header(peek_more: Callable[[int], bytes]) -> HitOutcome:
     if name_len == 0:
         return HitOutcome.NOT_THIS_FORMAT
     needed = _LOCAL_HEADER_SIZE + name_len + extra_len
+    if remaining is not None and needed > remaining:
+        return HitOutcome.NOT_THIS_FORMAT
     rest = peek_more(needed)
     if len(rest) < needed:
         return HitOutcome.NOT_THIS_FORMAT
```
davitf: ### F7 🟢 `[nit]` · CONFIRMED — **still open** (narrowed, not closed)

The `remaining` overrun check on line 97 is the right half and it landed. But the short-read reject below it is the original F7 defect, and it survives: when `remaining` has *already proved* the bytes exist, a short `peek_more` can only mean the candidate view was clamped by `scan_limit` — yet it still returns `NOT_THIS_FORMAT`.

Unit level:

```python
full    = lambda n: p[:n]
clamped = lambda n: p[:min(n, 40)]   # 8 bytes short of the 48 needed

validate_zip_local_header(full,    len(p))  -> HitOutcome.VALID
validate_zip_local_header(clamped, len(p))  -> HitOutcome.NOT_THIS_FORMAT
```

End to end, with a real ZIP behind a 4 KiB ELF stub and `max_scan_bytes` tuned so the header straddles the ceiling:

```
scan_limit=    4136  candidate@4096 needs 48 bytes -> FormatDetectionError
scan_limit= 1048576  candidate@4096 needs 48 bytes -> ZIP @4096 by=sfx_scan
```

Same shape in both RAR paths, and there it is unguarded because `validate_rar_main_header` does `del remaining`:

```
stored_m0.rar  clamp@20 -> NOT_THIS_FORMAT      (remaining proves the bytes exist)
stored_m0.rar  unclamped -> VALID
```

The RAR docstring's *"`remaining` is unused: a RAR identity check never needs the source length"* is a shade stronger than true — this is exactly what it would be for.

**Reachability is small**, which is why this stays 🟢: the candidate has to land within `30 + name_len + extra_len` of `scan_limit` for ZIP (typically ~100 bytes at the 2 MiB boundary) or within a MAIN header's length for RAR. Not worth blocking on.

**Fix, when convenient:** make the short-read reject conditional on not having proof — `if remaining is None and len(rest) < needed`. Above that line, `needed > remaining` has already rejected a genuine overrun, so once `remaining` is known a short read carries no information about the format.

---

src/archivey/internal/sfx.py
```diff
@@ -81,16 +81,34 @@ class HitOutcome(Enum):
...
+    source length is unknown. It is a length, not a peek budget — do not confuse
+    it with ``scan_limit``. Default ``None`` keeps unit tests and one-arg calls
+    working.
+    """
+
+    def __call__(
+        self,
+        peek_more: Callable[[int], bytes],
+        remaining: int | None = None,
+    ) -> HitOutcome: ...
```
davitf: ### F13 🟢 `[nit]` — the default on `remaining` makes the load-bearing argument look optional

`remaining: int | None = None` on the Protocol and on all three implementations. The scan always passes it (`detection.py:467-473`), so the default exists only so unit tests and the binding test can call one-arg.

Two small costs. It makes the parameter whose absence *was* F1 read as optional, and it lets a validator quietly ignore the length without any signal — `validate_rar_main_header` already does (`del remaining`), which is the F7 residual next door. `None` and "not passed" also become indistinguishable at the call site, so a future scan path that forgets to plumb it degrades silently into the unknown-length branch rather than failing.

Not worth churn on its own; worth folding into the F7 fix if you touch these signatures again. Dropping the default costs the tests one explicit `None` each, and the parametrized 7z table already passes `remaining` positionally.

---

davitf (commented):
## Re-review — head `465b5a6` (was `60bc90e`)

**Verdict: ✅ Approve.** Posted as a `COMMENT` review because GitHub refuses an approval on a PR from the same account. Two 🟢 nits remain, neither blocking.

### Status of the previous round

| ID | Was | Now |
|---|---|---|
| **F1** | 🔴 declared-end measured against the peek window | **Fixed** — `remaining` is candidate-relative source length; verified below |
| **F2** | 🟡 23× read amplification | **Fixed** — 65,536 bytes, identical to the no-validator baseline |
| **F3** | 🟡 `NOT_THIS_FORMAT` after the CRC held | **Fixed** — post-CRC rejects are `DAMAGED`, pinned in the table |
| **F4** | 🟡 rewrite dropped the pin on a live hole | **Partly fixed, deliberately** — see below |
| **F5** | 🟡 spec / amendment / code said three different things | **Fixed** — tasks.md, design.md and the module docstring now agree |
| **F6** | 🟡 parser logic and constants copied | **Fixed** — `load_vint` + `rar3_main_crc_end` shared, 7z constants and `_crc32` imported, no wrapper |
| **F7** | 🟢 short read from a clamped view is a reject | **Still open** (narrowed) — inline on `zip_detect.py` |
| **F8** | 🟢 unreachable guard | **Fixed** — gone |
| **F9** | 🟢 unresolvable `F1 analogue` reference | **Fixed** — gone from both modules |
| **F10** | 🟢 one test, ten assertions | **Fixed** — parametrized, and the new cases are readable |
| **F11** | 🟢 binding test lost its ZIP decoy | **Fixed** — per-class decoys |
| **F12** | 🟡 task 2.3 marked `[x]` for undone work | **Fixed** — `[~]`, honest task text, `known-issues.md` entry |

New this round: **F13** 🟢 (inline on `sfx.py`).

### What I ran on `465b5a6`

- `./scripts/check.sh` → **all checks passed** (ruff, pyrefly, ty, openspec, docs).
- `./scripts/test.sh` (`[all]`) → **2671 passed, 23 skipped, 3 deselected, 1 xfailed** in 169s. The one xfail is the strict red half. Matches the PR body; I did not re-run the other two configs.
- `python3 scripts/check_openspec_archived.py` → `no finished OpenSpec change is left unarchived`, exit 0. The `[~]` is accepted by the gate and does keep the change unarchivable.
- The four repro scripts from round one, re-run verbatim.

### Verification of the fixes

**F1** — the case that raised `FormatDetectionError` last round now resolves, and the budget is a cost gate again:

```
stub=  262144  declared=1945696  stub+declared=2207840  -> SEVEN_Z @262144 by=sfx_scan
stub=    4096  declared=1500064  FAST budget            -> SEVEN_Z @4096   by=sfx_scan
```

**F2** — read amplification gone, and now pinned by a peek-counting unit test rather than only by measurement:

```
with validator   : SEVEN_Z  unique_bytes_read = 65,536
without validator: SEVEN_Z  unique_bytes_read = 65,536
```

**F4** — the empty-header reject closes the reproduced decoy end to end. `open_archive` on that file now returns the real members instead of an empty archive:

```
detected SEVEN_Z offset 3544 (decoy at 512, real at 3544)
opened: ['scripts/', 'scripts/py7zr', 'setup.cfg']
```

**F6** — the RAR try/except came out the way D3 described: `peek_more` outside the `try`, `identity_held` set immediately after the CRC match so a later vint failure is `DAMAGED` rather than `NOT_THIS_FORMAT`, `load_vint` called directly with no wrapper on the parser's 30 hot-path call sites, and the MAIN CRC walk now has exactly one definition that both sides call.

### On the deferral

The D2(b) shape landed as specified, and it is the part I'd have been most willing to see done badly. `test_inexact_7z_decoy_loses_to_a_later_exact_payload` asserts the **correct** contract and carries `strict=True` on the marker, so the day the tie-break lands it XPASSes and fails the suite. Task 2.3 is `[~]`, verified against the real gate. The `known-issues.md` entry names the exact residual — *a decoy with a nonzero, self-consistent declared size that still fits in `remaining`* — rather than gesturing at "decoys", says that the failure mode is a wrong answer rather than an error, and even records why `strict=True` has to be on the marker. Three markers, three half-lives, all live.

Nothing in the tree now asserts the wrong offset as correct.

### Also fine

The reverted ranking commit (`af44c24` → `465b5a6`) is a clean revert: no `VALID_EXACT` left on `HitOutcome`, no orphaned ranking state in `_scan_for_sfx_payload`. `HitValidator` as a Protocol reads better than the nested `Callable` it replaced, in `base_reader`, `registry` and `detection` alike. And the `next_header_size == 0` reject is correctly scoped to the scan path — a genuine empty `.7z` is claimed by near magic (`SEVEN_Z by magic offset 0`) and never reaches the validator, which the docstring says out loud.

One judgement call worth naming rather than filing: that empty-header reject returns `NOT_THIS_FORMAT` even though the CRC has matched, which is in tension with F3's rule. It is the right call — `DAMAGED` would later have the ledger report `SEVEN_Z` for a decoy — and the docstring flags the tension, so no action.

**Verdict: ✅ Approve.** F7 and F13 are 🟢 and can ride along with whatever next touches these signatures.

---

PR #280, opened 2026-09-02
docs: pair workflow, living handbook stubs, thin grill/writing skills

davitf:
## Summary

Maintainer-facing **pair workflow** + thin skills, with review findings addressed.

## Decisions applied

- **DP1 = C** — OpenSpec main specs stay the authoritative *contract*; handbook/pair-workflow is the primary human reading surface.
- **DP2 = B** — No empty `formats/` / `topics/` trees; create the first real handbook page with the change that needs it.

## Specs → handbook + tests (recorded)

Direction + **thin-as-you-go** anti-junk rule:
[`dev-docs/discussions/2026-09-specs-to-handbook-and-tests.md`](https://github.com/davitf/archivey/blob/cursor/pair-workflow-docs-9573/dev-docs/discussions/2026-09-specs-to-handbook-and-tests.md)

Wired into `CONTRIBUTING.md` §Working with the specs, `openspec/schemas/library/README.md`, `pair-workflow.md`, `openspec/project.md`, `IDEAS.md`, and the adoption discussion. Every future PR that touches specs should leave that capability no denser than it found it (prefer pytest + handbook Verify over new WHEN/THEN).

## Status

All review findings F1–F11 dispositioned. Gate: `./scripts/check.sh` green (incl. internal md links).

## How to start

`dev-docs/pair-workflow.md` + `dev-docs/discussions/2026-09-pair-workflow-adoption.md`.

<div><a href="https://cursor.com/agents/bc-07f2d0c2-c431-45e7-a3da-32fef5519573?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-07f2d0c2-c431-45e7-a3da-32fef5519573&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


dev-docs/pair-workflow.md
```diff
@@ -0,0 +1,186 @@
...
+PR. The maintainer is not the audience for that list unless they ask.
+
+---
+
+## How to start using this (checklist)
+
+1. **Read this file** and the [`formats/`](formats/README.md) / [`topics/`](topics/README.md) stubs.  
+2. **Invoke the thin skills** below on the next medium change (`/grill-with-handbook`,
+   then implement against a thin brief).  
+3. **Review with the other model** via `/code-review` / `/code-review-skill`; tell it
```
davitf: **F1** 🔴 `[blocking]` · `CONFIRMED` — this routes the reader to the *builtin* `/code-review`, not this repo's review skill

Line 105 and line 180 both name `/code-review` as the way to get an archivey review. Three existing places warn that this is a different skill:

- `CLAUDE.md:34` — "a bare `/code-review` is a *builtin* skill, not this repo's"
- `AGENTS.md:296` — "not a bare `/code-review`, which is a *builtin* skill in both Claude Code and Cursor and is not this one"
- `.claude/skills/code-review-skill/SKILL.md:38` — the same warning again

In Cursor it happens to resolve because `.cursor/commands/code-review.md` shadows the builtin. In Claude Code it does not — the skill is `code-review-skill`. So the document that is meant to become the everyday entry point sends you to the wrong reviewer, which is precisely the failure the other three warnings exist to prevent, and it does it in the two places a reader is most likely to copy from (the "How to start" checklist and the entry-points table).

**Fix:** `/code-review-skill` on both lines, with the Cursor exception stated once — e.g. "`/code-review-skill` (in Cursor, `/code-review` resolves to it via the project command)".

---

dev-docs/pair-workflow.md
```diff
@@ -0,0 +1,186 @@
...
+   explicitly: *post the full handoff on the PR; message me only with decision packets.*  
+4. **Address** with `/address-review` — expect packets, not a wall of finding recap.
```
davitf: **F5** 🟡 `[important]` · `CONFIRMED` (first two) / `PLAUSIBLE` (third) — three more command names that will not resolve where the doc implies

1. **`/address-review`** (here and line 181). The Claude Code skill is named `address-review-findings`; `/address-review` exists only as the Cursor project command `.cursor/commands/address-review.md`. `AGENTS.md:306` gets this right by naming both ("`address-review-findings` (Cursor: `/address-review`)"); this doc does not.
2. **`/unslop`**, promised in `.claude/skills/technical-writing/SKILL.md:6` ("when the user invokes /technical-writing or /unslop"). A skill is invocable under its own `name:` only, so `/unslop` resolves to nothing. Either drop it from the description or ship an alias command.
3. **`/add-plugin pstack`** (line 184) — I could not confirm this command exists; plugin management is under `/plugin` as far as I know. Worth verifying before it becomes the documented instruction, since nobody will catch it until they try.

The heading over the line-177 table is "Cursor / Claude entry points" but the rows never say which host each name belongs to, so every one of these reads as valid in both.

**Fix:** name the skill in each row and put the host-specific slash form in parentheses, the way `AGENTS.md` already does.

---

.claude/skills/address-review-findings/SKILL.md
```diff
@@ -211,22 +211,25 @@ onto the person you are asking.
 **Do not escalate** routine implementation choices, anything the specs already settle, or
 questions manufactured to look thorough.
 
-### What "full context" means
+### Decision packet (required escalate shape)
```
davitf: **F3** 🟡 `[important]` · `CONFIRMED` — the packet contract is now written out four times

The six numbered fields appear in full in three files and are summarised in a fourth:

- `dev-docs/pair-workflow.md:83-92` (declared canonical: "same contract as…")
- this file, lines 214-227
- `.claude/skills/code-review-skill/reference/archivey-review-addendum.md:112-119`
- `.cursor/commands/address-review.md:24-27` (compressed to one sentence)

They already differ in wording — this file says "**Default if you ignore this** — what you will ship / leave unchanged", the addendum says "**Default if ignored** — what happens if nobody answers", `pair-workflow.md` says "**Default if you ignore this** — what ships". Harmless today; it is exactly how the next divergence starts, and there is no gate that would notice.

This is the same rule the PR itself introduces two files over: `.claude/skills/technical-writing/SKILL.md:53` ("Prefer one source of truth; link the handbook instead of restating") and `CLAUDE.md:8` ("everything else would drift if it were duplicated here").

**Fix:** keep the six fields in `dev-docs/pair-workflow.md` only. Here and in the addendum, link it and state only the local delta (here: "one packet per turn, via `AskUserQuestion`"; addendum: "block 3 items use packet shape"). The `.cursor/` summary can stay — it is a pointer, not a second definition.

---

CONTRIBUTING.md
```diff
@@ -295,23 +295,36 @@ decided explicitly.
...
+   first (`/grill-with-handbook`).
+4. **Does a contributor need a live register or runbook *today*?** → `dev-docs/`
+   (e.g. threat model, known issues, release checklist).
+5. **Is it finished evidence — an investigation, a superseded design, a lab
    notebook?** → `dev-docs/investigations/`, or `dev-docs/history/` for prose that a
    newer document replaced.
 
 If it is a *review*, it belongs to the `review/` lifecycle. If it is a *proposed
-behaviour change*, it belongs to `openspec/changes/`.
+behaviour change* that must update machine-checkable main specs, it belongs to
```
davitf: **F4** 🔴 `[blocking]` · `CONFIRMED` — this narrows when an OpenSpec change is required, as a side effect of a docs PR

Before: "If it is a *proposed behaviour change*, it belongs to `openspec/changes/`."
After: "…*that must update machine-checkable main specs*, it belongs to `openspec/changes/`."

That is a real policy change — a behaviour change can now ship on a thin brief if the author judges the main specs need not move. Two other edits in this same PR push the same direction:

- `AGENTS.md:53` drops "the **authoritative** capability specs" → "dense capability specs … **Not** the primary maintainer reading surface".
- `dev-docs/index.md` deletes the sentence "Normative behavior remains in `openspec/specs/`."

Meanwhile the published docs still tell users the opposite, on three pages: `docs/index.md:87-88`, `docs/formats.md:4` ("Authoritative detail lives in `openspec/specs/format-*`"), `docs/api.md:5` ("Authoritative contracts: `openspec/specs/`"). That trio is claim **I-12** in `review/docs-content/claims.md:869`, dispositioned **Keep / verified**. Nothing in this PR touches those pages, so on merge the maintainer docs and the published docs disagree about whether the specs bind.

Demoting the specs as your *reading surface* is clearly the intent and is fine. Demoting them as the *contract* is a separate call — see decision packet **DP1** in the review body.

**Fix (if DP1 goes the way I recommend):** restore the old CONTRIBUTING sentence, keep the word "authoritative" in `AGENTS.md:53`, and scope the demotion to reading surface only.

---

.claude/skills/technical-writing/SKILL.md
```diff
@@ -0,0 +1,56 @@
+---
+name: technical-writing
+description: |
+  Diátaxis structure, plain developer sentences, and unslop (cut AI tells). Use when
+  writing or reviewing docs/, handbook prose, PR descriptions, or commit messages; or
+  when the user invokes /technical-writing or /unslop.
+disable-model-invocation: true
```
davitf: **F6** 🟡 `[important]` · `CONFIRMED` — `disable-model-invocation` stops this skill firing exactly where the workflow depends on it

With this flag the skill only runs when a human types `/technical-writing`. But the loop this PR defines has agents writing prose without a human in the turn:

- step 4, "Implement … update handbook/user docs when claims move" (`dev-docs/pair-workflow.md:19`)
- step 7, "User docs if needed — Diátaxis mode + unslop" (line 22)
- `CONTRIBUTING.md:302`, added here: "Use `/technical-writing` (Diátaxis + unslop) for the prose."

An implementing agent updating `docs/` mid-PR will never load it, so the unslop pass lands only when you remember to ask.

The description is written for auto-triggering too — "Use when writing or reviewing docs/, handbook prose, PR descriptions, or commit messages" is trigger phrasing that the flag disables, so the two halves of the frontmatter contradict each other.

**Fix:** drop `disable-model-invocation` here (the description already scopes it well), or keep the flag and cut the trigger list down to "invoke-only". On `.claude/skills/grill-with-handbook/SKILL.md:8` the same flag is correct — a grill starting itself would be worse than one that never starts — so this is not a blanket objection.

---

dev-docs/topics/README.md
```diff
@@ -0,0 +1,17 @@
...
+Cross-cutting living notes (same genre as [`../formats/`](../formats/README.md)):
+short, current, rewritten when claims move. Prefer these over new ADRs for
+decisions that span formats.
+
+Suggested first pages (create when a change needs them):
+
+| Topic | Suggested file | Seeds |
+| --- | --- | --- |
+| Detection / SFX / prefixes | `detection.md` | `openspec/specs/format-detection`, detection investigations |
+| Extraction safety | `extraction-safety.md` | [`threat-model.md`](../threat-model.md), safe-extraction spec |
```
davitf: **F9** 🟡 `[important]` · `CONFIRMED` — this gives the threat model's open-gap register a second home

`extraction-safety.md` seeded from `threat-model.md` collides with two other rules landing in the same PR:

- `CONTRIBUTING.md` Q2 sends "current maintainer truth about a format or cross-cutting topic" to `topics/`, while Q4 sends live registers to `dev-docs/` and names the threat model as an example. Extraction safety satisfies both, and the first `yes` wins — so Q2 claims it.
- `dev-docs/pair-workflow.md:66` lists `threat-model.md` *inside* the handbook table ("Trust boundaries (edit in place)"), i.e. the threat model is itself a handbook page — which makes a `topics/extraction-safety.md` seeded from it a copy of a copy.

The `O*` register is not free-floating prose: `AGENTS.md:52` says each open item becomes an OpenSpec change when tackled, and the review program keys on those IDs. Two pages restating it is how an `O` row gets fixed in one place and left stale in the other.

**Fix:** say explicitly on this page that `threat-model.md` stays the register, and a topic page links to `O*` rows rather than restating them. Same treatment for `detection.md` vs the detection specs.

---

dev-docs/pair-workflow.md
```diff
@@ -0,0 +1,186 @@
...
+6. **Default if you ignore this** — what ships  
+
+If an agent cannot fill these, it is not ready to ask — it should measure first.
+
+Review quality does **not** drop: the implementor still gets the full finding list on the
+PR. The maintainer is not the audience for that list unless they ask.
+
+---
+
+## How to start using this (checklist)
```
davitf: **F10** 🟢 `[nit]` · `CONFIRMED` — the adoption plan is baked into a page that is supposed to be current truth

Lines 100-184 are addressed to one person at one moment: "How to start using this (checklist)", the "What to take from external stacks" matrix comparing pstack/Matt Pocock skills, "Pilot: next change that touches 7z or RAR", and the desktop-plugin aside. Once you have adopted the loop, none of it is true any more, and it is over half the file.

This is the genre the same PR rules against: `.claude/skills/technical-writing/SKILL.md:54` ("Do not dump lab evidence into living pages — link `dev-docs/investigations/`") and `dev-docs/formats/README.md:2` ("Rewrite in place when behaviour changes"). `CONTRIBUTING.md` Q5 sends a proposal like this to `dev-docs/discussions/`, which already has the RESOLVED-header convention for exactly this shape.

**Fix:** keep the loop, the handbook table, the thin brief and the packet in `pair-workflow.md` — that is the durable half, and roughly one screen. Move the checklist and the external-stack matrix to `dev-docs/discussions/` with a link back.

---

dev-docs/index.md
```diff
@@ -5,18 +5,22 @@ site: everything under `docs/` is for users, and everything here is not.
 
 | Doc | Role |
 | --- | --- |
+| [Pair workflow](pair-workflow.md) | **Preferred everyday loop**: investigate → grill into handbook → thin brief → implement → other-agent review → decision packets. What to borrow from external skill stacks |
+| [Format handbook](formats/README.md) | Living per-format pages (pilot): behaviour, consequences, light decisions, verify |
```
davitf: **F11** 🟢 `[nit]` · `CONFIRMED` — two empty directories now outrank the code map in the index

Rows 2 and 3 point at `formats/README.md` and `topics/README.md`, whose own index tables are entirely `— | not started` (7 rows) and "create when a change needs them" (4 rows). They sit above `code-map.md`, which `AGENTS.md:21` designates as the starting point for "where in the source do I make this change".

The reader you are optimising for lands on two stubs before the page that answers their question.

**Fix:** keep `pair-workflow.md` at the top, then `code-map.md`, and move the two handbook rows below it until at least one page exists behind them.

---

davitf (commented):
## 1. Maintainer briefing

**What this change is.** Docs and process only — no `src/` change. It writes down a maintainer-facing pair workflow (investigate → grill → thin brief → implement → other-model review → escalate), adds two stub handbook trees (`dev-docs/formats/`, `dev-docs/topics/`), two new thin skills (`grill-with-handbook`, `technical-writing`) with matching Cursor commands, and tightens the escalation contract in the two review skills into a six-field **decision packet**. Pointers were added in `AGENTS.md`, `CONTRIBUTING.md`, `dev-docs/index.md`, the code map and the decisions index. Behavioural delta for humans: the maintainer is meant to read packets and handbook pages instead of OpenSpec change packs and three-block essays.

**Snapshot.** 16 files, +484/−58, docs-only. Gates: CI is green on `21a1f31` (23/23 checks) **per GitHub — I did not re-run it**. What I did run: `uv run --group docs python scripts/check_docs_nav.py` → green, though it covers `docs/` only and therefore inspects nothing this PR touched. I also hand-verified every relative link the PR adds (~30 across `dev-docs/` and `.claude/skills/`) — all resolve at this HEAD. Full `check.sh` not re-run; no Python changed.

**Verdict: 🔄 Request Changes** — submitted as a Comment review because GitHub refuses a changes-requested review on one's own PR. Everything below is a cheap edit; nothing here argues against the direction.

**Main points**

- 🔴 **F1** — `pair-workflow.md` tells the reader to invoke `/code-review`, the *builtin* skill the repo warns about in three separate places. Inline.
- 🔴 **F4** — `CONTRIBUTING.md` quietly narrows when a behaviour change needs an OpenSpec change, and two sibling edits demote the specs' authority, while three published pages still tell users the specs are authoritative. Inline; decision **DP1** below.
- 🔴 **F2** — `AGENTS.md` §Review workflow, the canonical agent guide, still carries the superseded escalation contract and never mentions packets. Below.
- 🟡 **F3** — the packet's six fields are now written out four times, already with three different wordings. Inline.
- 🟡 **F5** — `/address-review`, `/unslop` and `/add-plugin` will not resolve where the doc implies. Inline.
- 🟡 **F6** — `technical-writing` carries `disable-model-invocation`, so it cannot fire during the implement step that depends on it. Inline.
- 🟡 **F7** — nothing checks the links this PR adds; `check_docs_nav.py` covers `docs/` only. Below.
- 🟡 **F8** — net reading surface goes up, not down. Below; decision **DP2**.
- 🟡 **F9** — `topics/extraction-safety.md` would give the threat model's `O*` register a second home. Inline.
- 🟢 **F10**, **F11** — adoption plan baked into a living page; two empty stubs outrank the code map in the index. Inline.

**What's fine.** The decision-packet contract is the part that actually pays for itself, and the block-3 rewrite in the review addendum is a clear improvement over the old prose list — "Default if ignored" in particular turns a question into something answerable by silence. The "same PR as the code when the change falsifies a doc claim" rule is well stated and belongs where it was put. Every relative link resolves, which for a PR that is mostly cross-references is not nothing.

---

## 2. Implementor handoff

PR #280 · `cursor/pair-workflow-docs-9573` @ `21a1f31` · docs/process review, no code in scope.

Eight findings are posted as **inline threads** on the lines they concern, so they can be dispositioned individually: **F1** (`dev-docs/pair-workflow.md:105`), **F3** (`.claude/skills/address-review-findings/SKILL.md:214`), **F4** (`CONTRIBUTING.md:321`), **F5** (`dev-docs/pair-workflow.md:107`), **F6** (`.claude/skills/technical-writing/SKILL.md:7`), **F9** (`dev-docs/topics/README.md:12`), **F10** (`dev-docs/pair-workflow.md:100`), **F11** (`dev-docs/index.md:9`).

The three below have no single line to anchor to.

### F2 🔴 `[blocking]` · `CONFIRMED` — `AGENTS.md` §Review workflow is left describing the superseded contract

`AGENTS.md:332` still reads:

> **Escalate one question at a time.** The maintainer is usually deciding without having read the diff or the docs, and does not know the identifiers. Expand every name, quote the actual code, show what you measured, give options with consequences and a labelled recommendation.

That is the old six-item "full context" shape this PR replaced in `.claude/skills/address-review-findings/SKILL.md`. It has no "Default if ignored", and the section says nothing about the new split-audience rule (full findings on the PR, packets only in chat) even though that rule changes what step 1 of the same section instructs an agent to post.

Why it matters more than a stale paragraph usually would: `CLAUDE.md:3` designates `AGENTS.md` as **the canonical agent guide** ("Read `AGENTS.md` first"), and `AGENTS.md:291-320` is where the two-agent handoff is defined. An agent that reads the canonical guide and stops — the documented path — gets the superseded contract and never learns packets exist. This is also the fourth copy from F3, and the one most likely to be read.

**Fix:** replace the `AGENTS.md:332` bullet with a one-line pointer to the packet definition in `dev-docs/pair-workflow.md`, and add the split-audience rule to step 1 of that section (it currently just says "It posts the findings to the PR").

### F7 🟡 `[important]` · `CONFIRMED` — nothing checks the links this PR adds

`scripts/check_docs_nav.py` validates `docs/` only. Its own docstring is the argument for extending it:

> Those URLs point at files this repository moves around, and nothing in a docs build looks at them — so they rot silently, which is the exact failure this whole reorganization exists to fix. Checking them offline costs nothing: the target is a path in this very repository.

This PR adds roughly 30 relative links under `dev-docs/` and `.claude/skills/`, none of them covered. Several are `../../../` hops out of a skill directory into `dev-docs/`, which break if a skill directory is ever renamed — and `AGENTS.md:317` already records that `.claude/skills/steward/` must not be renamed for a related reason. The handbook design makes this worse over time, since its whole premise is pages linking each other.

I verified all of them by hand at this HEAD and they resolve. That does not survive the next refactor.

**Fix:** extend `check_docs_nav.py` (or a sibling) with a relative-link existence pass over `dev-docs/**/*.md` and `.claude/skills/**/*.md`, wired into the `docs` leg of `check.sh`. Anchors and absolute URLs are not needed — plain path existence catches the failure mode. Roughly 15 lines against the machinery already there.

### F8 🟡 `[important]` · `PLAUSIBLE` — the change adds reading surface where its goal is to remove it

Taking the stated goal seriously: after this PR the maintainer has *more* to read, not less. It adds a 186-line workflow document, two `README`s whose index tables are entirely "not started", and pointer paragraphs in six existing files; `dev-docs/code-map.md`'s "where the answers live" table grows to seven destinations (`openspec/specs/`, `formats/`, `topics/`, `decisions/`, `investigations/`, `known-issues.md`, `review/`). The mechanism that genuinely shrinks the human's inbox — the packet — is about 30 lines in two skill files and does not depend on any of the handbook scaffolding.

The stubs are the specific risk. `dev-docs/formats/README.md` publishes a seven-row index in which every row is `— | not started`, plus "7z | pilot candidate". A future agent reading `CONTRIBUTING.md` Q2 will route a decision to `dev-docs/formats/7z.md`, find nothing, and either create a page from a template it does not have enough context to fill or drop the decision. An empty structure that documents itself as the preferred destination is load-bearing before it has any load.

Labelled `PLAUSIBLE` because this is a judgement about how the docs will be used, not something I can trace to a failure. See **DP2**.

**Fix direction:** land the packet contract and the correction findings now; hold `formats/` and `topics/` until the first real page exists, and create the README alongside it.

**Verdict: 🔄 Request Changes** (submitted as Comment — GitHub blocks changes-requested on one's own PR).

---

## 3. Maintainer decisions

### DP1 — Do the OpenSpec specs stop being the authoritative contract, or only stop being your reading surface?

**Why it matters.** It decides whether a behaviour change can ship without a spec update. Downstream: three published pages currently promise users that `openspec/specs/` is authoritative, so the answer either holds or those pages need editing. It also changes what a reviewing agent treats as binding when code and spec disagree — today `AGENTS.md:56` says pause and ask.

**Options.**
- **A — reading surface only.** Specs stay authoritative; the handbook becomes what *you* read. Cost: `CONTRIBUTING.md:321`, `AGENTS.md:53` and the `dev-docs/index.md` deletion need reverting; the OpenSpec overhead you are trying to escape stays for contract-touching changes.
- **B — full demotion.** Specs become an optional agent/CI artefact, as the PR currently reads. Cost: `docs/index.md:87-88`, `docs/formats.md:4` and `docs/api.md:5` become false and must change in this PR; claim I-12 in `review/docs-content/claims.md` gets re-opened; agents lose the one unambiguous "this binds you" pointer.
- **C — split it explicitly.** Specs remain the authoritative *contract*, and the docs say in one sentence that they are not the primary human reading surface. Cost: one added sentence in `AGENTS.md`; nothing else moves.

**Evidence.** `CONTRIBUTING.md` went from "If it is a *proposed behaviour change*, it belongs to `openspec/changes/`" to "…*that must update machine-checkable main specs*". `AGENTS.md:53` dropped the word "authoritative". `dev-docs/index.md` dropped "Normative behavior remains in `openspec/specs/`." No published page was touched; `docs/formats.md:4` still reads "Authoritative detail lives in `openspec/specs/format-*`", and `review/docs-content/claims.md:869` has that trio dispositioned **Keep / verified**. Nothing in CI would catch the disagreement — `check_openspec_archived.py` only checks that in-flight changes get archived.

**Recommendation: C.** The reading-surface complaint is real and C fixes all of it; B additionally removes the only statement that makes an agent treat a spec as binding, which is a different problem from the one you set out to solve.

**Default if ignored:** B ships, and the maintainer docs contradict three published pages from the merge commit onward.

### DP2 — Land the whole thing, or split the enforceable half from the aspirational half?

**Why it matters.** Determines whether the empty handbook trees exist in `main` before any page fills them, and therefore whether agents start routing decisions into directories with no content.

**Options.**
- **A — merge as is.** Everything lands together; adoption depends on you remembering the pilot next time you touch 7z or RAR. Cost: F8's empty-destination problem is live from day one.
- **B — split.** PR 1: packets + F1/F2/F5 corrections + `check_docs_nav` extension (all self-contained, all enforceable today). PR 2: `formats/` + `topics/` created together with the first real page. Cost: two PRs and the handbook lands weeks later.
- **C — merge as is, plus a forcing function.** Keep the trees but add the pilot as an explicit item somewhere that gets read again — `review/STATUS.md` or `dev-docs/PLAN.md` — so the first page is not left to memory.

**Evidence.** `dev-docs/formats/README.md` ships a seven-row index in which every entry is `— | not started`; `dev-docs/topics/README.md` lists four suggested files, none created. Neither directory has a page. Against that, the packet edits in the two skills and the addendum are complete and take effect the moment they merge — they need none of the scaffolding.

**Recommendation: B**, but **C** is a reasonable trade if you would rather not carry a second PR — the pilot note is what makes the difference, not the split itself.

**Default if ignored:** A ships, and `formats/`/`topics/` sit empty until something forces the issue.

---

---

PR #284, opened 2026-09-02
docs: dev-docs/formats/zip.md + topics/prefixed-archives.md, and the page shape they settle

davitf:
## Goal

The first two living-handbook pages, plus the page structure they were written to discover
(`pair-workflow.md` §Living handbook has said "TBD" on purpose since #283).

The ZIP page was written from a harvest of what the repo already knows — the
problem-catalogue leads on the unmerged #244 branch, `history/SPEC.md`, the registers,
five archived review rounds, the detection investigation, the four archived ZIP changes
and the in-flight `prefixed-archive-detection`, then `src/` and `tests/` to check the
claims. Structure was chosen from the shape the material actually had, not from a template.

## Non-goals

- No behaviour change. Documentation, plus one spec correction.
- No `formats/README.md` index ahead of a third page.

## What the shape turned out to be

Eight sections, now recorded in `pair-workflow.md` §Format page structure with `zip.md` as
the worked example. Three rules the shape depends on:

- **Structural facts carry their consequences inline.** The strongest grouping force in the
  material was causal, not topical — "the central directory is at the end" generates the
  non-seekable refusal, the prefix story, the tail probe, the decoy self-correction, the
  empty-archive case, multi-volume addressing and the native-parser wishlist. A separate
  "consequences" section breaks that chain and makes the reader re-derive it.
- **The middle of the page is ordered by pipeline stage**, not by topic: identify · open and
  list · member data · extract · write. Each says who does the work, what is format-specific
  rather than general, and what is refused. That generalises the awkward "ZIP uses two
  different libraries" observation into a question every format answers, and it makes
  "nothing here is format-specific" a useful answer — which is exactly ZIP's extract stage.
- **Pitfalls are tagged format / library / archivey**, so a reader can tell an inherent limit
  from one that needs an upstream fix from one that is ours. Status and fix plans stay behind
  the register link; the page states only what a caller observes.

No performance numbers anywhere — they were the most volatile claims in the harvest, and
Verify carries the command instead.

`topics/prefixed-archives.md` took the shared prefix/SFX machinery so `zip.md` §2.1 keeps
only ZIP's residue. Writing the two together settled the split rule, also in pair-workflow:
**a format page keeps what the format's own structure decides; a topic page keeps the shared
machinery.** Topic pages take the conventions rather than the section list — this one came
out looser than the format shape.

## Public contract deltas

`2026-09-02-drop-unshipped-write-claims` (proposed, implemented and archived here):
`format-zip`'s "Support streaming ZIP write via data descriptors" requirement is **removed**,
and the `Write support` cells in `format-zip` and `format-single-file-compressors` go to No.
Nothing implemented them — there is no `archivey.create`, no `ArchiveWriter`, no writer
module — and writing is `PLAN.md` phase 9. Restoring the requirement belongs to the writing
phase rather than to an annotation now.

## Things the harvest verified against the code

Behaviour today, stated in `zip.md` §5 with its tag. The write claim above is fixed here;
these two are not, and are the open questions this PR raises.

- 7-Zip's split-ZIP naming (`name.zip.001…`) is not recognised as a split set. The first
  part reports `CorruptionError`, the rest `FormatDetectionError`. The `.z01` rule is a
  filename pattern that does not cover it. Same shape as P17, unregistered.
- An EOCD declaring a non-zero disk number opens and lists normally. Those fields are parsed
  by stdlib and never checked; only the ZIP64 locator path refuses, by matching stdlib's
  exception text. `format-zip` has a scenario claiming otherwise.

## Verify

```bash
./scripts/check.sh
./scripts/test.sh tests/test_zip.py tests/test_zip_aes.py \
    tests/test_zip_native_codecs.py tests/test_zip_multipassword.py
```

`check.sh` passes, including `openspec validate --all` and the archive gate. The ZIP suite is
108 passed. Every relative link in the changed files resolves.

## Out of scope

The two items above; the `sfx_scan` rename the detection investigation proposes; and the
remaining write claims, deliberately untouched pending a decision — the `archive-writing`
capability (5 requirements, including the `CompressionSpec` model and streaming conversion),
`format-tar`'s write requirement, and `testing-contract`'s
round-trip-for-every-writable-format requirement.

https://claude.ai/code/session_014o1iwDmwTBR52DWFtGRmVE


dev-docs/formats/zip.md
```diff
@@ -0,0 +1,335 @@
...
+The extras are named for what they provide, not for ZIP, because every one of those
+codecs is shared with 7z or TAR. See [`packaging-and-extras`](../../openspec/specs/packaging-and-extras/spec.md).
+
+## 1. Shape
+
+Four properties generate most of this page. Everything below is a consequence of one of
+them.
+
+```
+[ optional prefix ] [ LFH+data ] [ LFH+data ] … [ CD entry ] [ CD entry ] … [ EOCD ]
```
davitf: define LFH and EOCD above or below

davitf: Done in 142143c — table under the diagram naming each record, its signature and its job, and the prose now uses those abbreviations.

Your two questions, answered on the page:

**The prefix is not part of the ZIP structure.** No field records it or its length, and nothing inside the archive changes because it is there. That is exactly why it works: everything is located from the EOCD backwards, so the format never had to know where the file began.

**One CDH per member, and it points back at its LFH** through `header_offset` — that pointer is the pairing. The CDH is authoritative for metadata; the LFH's copy of the name, method and flags is a duplicate, and the sizes and CRC in it are zeros when bit 3 is set. archivey reads the LFH only to find where the data starts, and cross-checks the two copies of the name (`_local_data_region`, mirroring stdlib) — a mismatch is `BadZipFile` → `CorruptionError`.

Two records the original diagram omitted are now in the table: the **DD** (data descriptor, only when bit 3 is set — and the same four bytes at offset 0 mean the split-archive spanning marker instead), and the **ZIP64 EOCD + locator**, which is the pair the multi-disk check in §5 actually reads.

---

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,168 @@
...
+| --- | --- | --- |
+| Self-extracting archive | `rar a -sfx`, `7z -sfx7zCon.sfx`, Windows installer stubs | A real executable that unpacks the payload |
+| Script launcher | `zipapp` / `.pyz`, pex, shiv, Spring Boot executable JAR, makeself `.run` | A `#!` line and usually a few lines of shell or Python |
+| Concatenation / polyglot | `cat stub payload.zip`, a JPEG with a ZIP appended | Anything at all |
+| Embedded | An archive inside a firmware or disk image | Anything at all, usually far in |
+
+The first two are meant to be *run*; the last two are not. Nothing in the bytes says which,
+so the machinery below reports what it found and does not classify intent.
+
+## 2. Three ways to find one
```
davitf: one more: prefix analysis. might be possible for a few formats, e.g. Make self, where we could read the script content to find the archive format and location. I think this is listed in ideas.md

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+                  ▲                       │  ▲                                      │
+                  └── header_offset ──────┘  └────── offset_cd ─────────────────────┤
+                                                     search backwards ≤ 65 557 B ◄──┘
+```
+
+| Part | Signature | What it is |
+| --- | --- | --- |
+| **prefix** | — | Arbitrary leading bytes. **Not part of the ZIP structure**: no field records it or its length, and nothing inside the archive has to change for it to be there. See §3 |
+| **LFH** — local file header | `PK\x03\x04` | One per member, immediately before that member's compressed bytes. Carries the name, extra fields, method and flags, and — unless bit 3 is set — the sizes and CRC. Its metadata is a *copy*; the CDH is authoritative |
+| **DD** — data descriptor | `PK\x07\x08`, optional | Present only when general-purpose bit 3 is set: the CRC and both sizes, written *after* the data because a streaming writer did not know them in advance. The same four bytes at offset 0 mean something else — the split-archive spanning marker |
```
davitf: "The same four bytes" -> "The same signature". it took me a while to understand which bytes were being referred to

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+**Building fixtures.** Stdlib `zipfile` cannot write encryption, so encrypted fixtures shell
+out to `7z` (`-mem=AES256` for WinZip AES, ZipCrypto by default) and skip when it is absent —
+one of the ~109 tests that vanish quietly on an unprovisioned container. `-mm=Deflate64`,
+`-mm=PPMd`, `-mm=BZip2` and `-mm=LZMA` produce the extended methods. The backslash fixtures
+are committed rather than generated because `zipfile` rewrites `ZipInfo.filename` on Windows;
+the reader uses `orig_filename` for the same reason.
+
+## 8. References
+
+- APPNOTE: §4.3.6 overall format · §4.3.9 data descriptor · §4.3.16 EOCD · §4.4.4
```
davitf: link to it? I found it at https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT (seems to be the official location, it's linked from Wikipedia)

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+in the format and refused here (§6); and a scan that lands on the wrong `PK\x03\x04` is
+usually self-correcting, because the reader still finds the real EOCD from the tail
+(§2.1).
+
+The backwards search bound is derived, not chosen: `comment_length` is a `uint16`, so the
+record cannot begin more than 65535 + 22 bytes before the end. A larger bound cannot find
+a valid EOCD and a smaller one rejects legal archives, so it is not configurable.
+
+**Members are independent.** Each member has its own LFH and its own compressed byte range, with no cross-member state. So random access is `DIRECT`, seeking within a
+member and holding two member streams open are both cheap, and there is no solid-block
```
davitf: "seeking within a member" is not necessarily cheap, it will typically require re-decompressing from the beginning of the member. exceptions are stored (uncompressed) members, and for some formats (specifically deflate) it's possible to use the accelerator library which would allow seeking to seek points.

Re two member streams: not sure that's relevant. Would that be different from one format to another?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+cost model. Both capabilities are still off by default — see §6.
```
davitf: "Both capabilities": not immediately understandable to me. You mean seeking and multiple open streams, right? I think it's better to state them explicitly.

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+
+### 2.1 Identify
+
+ZIP declares three magics at offset 0 — `PK\x03\x04` (local header), `PK\x05\x06` (empty
+archive) and `PK\x07\x08` (spanned marker) — and five extensions: `.zip`, `.jar`, `.pyz`,
+`.whl`, `.apk`.
+
+Only the local header is offered as a scan needle for prefixed archives. The other two are
+legitimate ZIP magic *at offset 0* but as needles inside a 2 MiB stub window they would
+claim any executable containing those four bytes. A needle hit is confirmed by
```
davitf: "they would claim...": that's also true of the local header, so this justification is weird. is it reducing false positives and/or because they're rare / unlikely to appear in a real prefixed archive ?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+zero-fill an ELF or PE stub pads with.
+
+The validator's method table is deliberately **wider** than the set archivey can decode: a
+member using a method we refuse to read is still a ZIP, and identity is not a support
+claim.
+
+Two things are ZIP-specific rather than general in the shared detector:
+
+- Prefixed ZIPs are found by the cued forward scan and reported as `ZIP` with a
+  `payload_offset`, never as a stream codec. The cue set, the scan window, the budget tiers
```
davitf: "never as a stream codec": why would they be reported as streams? that seemed weird. is it intended to indicate that zip matching is stronger than the on-magic streams which need probes?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+name+extra within the source. That last pair is what rejects `PK\x03\x04` followed by the
+zero-fill an ELF or PE stub pads with.
+
+The validator's method table is deliberately **wider** than the set archivey can decode: a
+member using a method we refuse to read is still a ZIP, and identity is not a support
+claim.
+
+Two things are ZIP-specific rather than general in the shared detector:
+
+- Prefixed ZIPs are found by the cued forward scan and reported as `ZIP` with a
```
davitf: what's the cue? does it mean the executable/script prefix?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+  [`topics/prefixed-archives.md`](../topics/prefixed-archives.md). What is ZIP's alone: the
+  needle and its validator above, the offset conventions in §3, and the fact that ZIP is
+  the only format that can locate itself from the end.
+- A **tail probe** — locating the EOCD directly instead of scanning forward — is designed
+  and not shipped, and ZIP is the only format it can serve. Until it lands, a prefixed ZIP behind bytes that fire no cue (a JPEG
+  polyglot, a plain concatenation) is not detected, though `open_archive(..., format=ZIP)`
+  reads it. Under `open_archive` the probe would be nearly free, since the backend reads
+  the EOCD anyway; under a bare `detect_format()` it is a tail read on every file,
+  including the overwhelming majority that are not ZIPs. That asymmetry, not false
+  positives, is why it is gated.
```
davitf: not sure I understand the justification. this seems to assume open_archive is called only with valid archives, while detect_format is not. if I try opening every file with open_archive, the cost is nearly the same as it uses detect_format internally, isn't it?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+**Metadata mapping.**
+
+| `ArchiveMember` field | Source | Absent when |
+| --- | --- | --- |
+| `name` | CDH name bytes, decoded as above | — |
+| `raw_name` | The stored bytes, verbatim | — |
+| `mode` | `external_attr >> 16` | The producer was not Unix-like, or `external_attr` is 0 — then `None`, never a substituted default |
+| `modified` / `accessed` / `created` | CDH DOS date-time (naive local, 2-second granularity) ← NTFS extra `0x000A` (UTC) ← Extended Timestamp `0x5455` (UTC), later overriding earlier | 1980 sentinel, or every layer invalid — with `MEMBER_TIMESTAMP_INVALID` |
+| `type` | Unix mode where present, else the directory marker and symlink hints | — |
+| `link_target` | The member's **data**, not its metadata | The archive is encrypted and no password is available — `SYMLINK_TARGET_UNAVAILABLE(reason="password_required")`, with no secret in the payload |
```
davitf: in what payload, of the extension? how would a secret even be able to be added there, if there wasn't a password to decrypt it?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+| `compression` | `compress_type` → `CompressionMethod` | — |
+| `is_encrypted` | `flag_bits & 0x1` | — |
+| `hashes["crc32"]` | CDH CRC, as four big-endian bytes | WinZip AE-2 members — the format zeroes the field and the HMAC is the integrity signal |
+| `comment`, `create_system` | CDH | Not recorded |
```
davitf: "Not recorded" -> "Unset in the archive"? "not recorded" gave me the impression that archivey might not record them when they're actually available.

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+| `extra` | `zip.compress_type`, plus `zip.aes_vendor_version` / `zip.aes_strength` / `zip.aes_actual_method` on AE members | — |
+
+Raw extra-field blobs are not surfaced. Duplicate names are legal and are not merged:
+members keep a positional `member_id`, name lookup is last-wins, and currency is computed
+in the reader spine so ZIP behaves like every other format.
+
+### 2.3 Member data
+
+**Stdlib's decoders are not used.** archivey locates the member's raw compressed bytes with
```
davitf: "Stdlib" is ambiguous: I think you mean zipfile's internal decoding. For deflate/bzip2/lzma, we likely end up delegating to the stdlib stream decoders

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,349 @@
...
+a bounded LFH parse (the fixed 30 bytes plus the local name and extra lengths, with
+absurd-length rejection and a name cross-check against the CDH), slices the source, and dispatches on the ZIP method id to the
+shared codec layer.
+
+The reason is coverage and uniformity, in that order. Stdlib decodes STORED, DEFLATE, BZIP2
+and LZMA; the codec layer adds Deflate64 (9), Zstd (93) and PPMd (98), which the registry
+already advertised for ZIP. It also puts ZIP member reads on the same `VerifyingStream` and
+the same exception translation as every other backend, so a corrupt body raises
+`CorruptionError` and a cut-short one raises `TruncatedError` through shared code.
```
davitf: also being able to use the accelerator libraries if the user has chosen to use them

dev-docs/formats/zip.md
davitf: Done in 1a9e196 — "The same signature at offset 0 means something else — the split-archive spanning marker".

---

dev-docs/formats/zip.md
davitf: Linked in 1a9e196, at the URL you found.

One caveat: I could not verify it from this container — `pkware.cachefly.net` is blocked by the egress proxy (403 on CONNECT), so the link rests on your check rather than mine. It is the URL Wikipedia and PKWARE's own developer page both point at, so I'm confident, but I did not fetch a byte of it.

---

dev-docs/formats/zip.md
davitf: Named explicitly in 1a9e196: "**Seekable member streams** and **concurrent member streams** are both off by default even here".

---

dev-docs/formats/zip.md
davitf: You're right on both counts, and the paragraph was wrong. Fixed in 1a9e196.

**Seeking within a member.** Member independence buys random access to member *starts*, not within one. A compressed member decodes sequentially, so a backward seek re-decodes from the member's start — `archive_stream.py:427` `_maybe_warn_rewind` exists precisely because that is O(n), and past `REWIND_REDECODE_WARN_BYTES` (1 MiB) it reports `STREAM_REWIND_REDECOMPRESSES`. The exceptions are the ones you name: STORED, and deflate under `use_rapidgzip`, whose comment at `config.py:147` says "gzip / zlib / **raw deflate**". I confirmed the accelerator does reach ZIP members rather than only bare streams — with `use_rapidgzip=ON` the inner stream under `ArchiveStream` changes from `DecompressorStream` to the accelerator's `VerifyingStream`. The page now says all of that.

**Two member streams.** It is format-dependent, so it earns its place — but the page asserted it instead of explaining it, which is why it read as filler. It follows from members being independent byte ranges over a seekable source: neither decoder depends on the other's state. A solid 7z or RAR block gives you none of that, and TAR on a non-seekable source cannot do it either. That contrast is now in the sentence, rather than the bare claim.

The §6 decision row had inherited the same overclaim ("Both are free here") and now matches [ADR 0003](https://github.com/davitf/archivey/blob/main/dev-docs/decisions/0003-member-streams-opt-in.md), which never said seeking was free on ZIP — its context line calls out "O(n) rewinds" explicitly.

---

dev-docs/formats/zip.md
davitf: You're right — the justification was circular. `PK\x03\x04` appears inside executables too, so "they would claim any executable containing those four bytes" does not distinguish the three.

The real reason is confirmability, not rarity: a local-header hit gets `validate_zip_local_header`, and the other two have nothing to be checked against — an EOCD found mid-window has fields that could be anything, and the spanned marker is four bytes with no structure at all. On top of that the payoff is nil: what a hit would identify is an *empty* archive or a spanning marker.

Your "reducing false positives" guess is the measured half, and the evidence was already sitting on the topic page: of 3 320 ELF/PE files under `/usr`, all six `PK\x05\x06` matches were false — `zip`, `zipnote`, `zipsplit`, `zipcloak`, `libzip.so`, `librevenge-stream.so`, each carrying the signature as a string constant and parsing to nonsense. Both pages now give that reason and cite the survey; fixed in 1a9e196.

---

dev-docs/formats/zip.md
davitf: Yes — the executable or script prefix. The cue is `MZ`, `\x7fELF` or `#!` at offset 0 (weak tier), which is what makes the forward scan worth running at all; the topic page has the weak/strong split and why Mach-O raises no weak cue. The page said "cued" without ever saying by what, so 1a9e196 names it inline.

---

dev-docs/formats/zip.md
davitf: It was compressed to the point of being cryptic. What it meant: the **content probes** are the tier that reports headerless stream codecs (Brotli, LZMA-alone), and they run *after* the SFX scan in the detector order — near magic → SFX scan → far magic → content probes → extension (`detection.py:25`). So a needle hit settles the file as a ZIP before a probe gets a chance to read the stub's own bytes as a headerless stream, which matters because the Brotli probe accepts a lot of arbitrary binary (`open-issues.md` P12) and a weak cue does not suppress the probes.

Your reading is close to right, but the mechanism is ordering rather than strength — the scan is not "stronger", it just runs first. 1a9e196 says that instead.

---

dev-docs/formats/zip.md
davitf: You're right, and the asymmetry I described does not exist. `open_archive` runs the same detection as `detect_format()`, so the wasted tail read falls on exactly the same files either way.

The split is between *outcomes*, not entry points: on a hit the read is free, because whatever opens the archive reads the EOCD anyway; on a miss it is pure waste, and most sources are not ZIPs. That is true under both APIs. What actually gates the tier is that nobody has measured what the wasted seek costs on a cold cache or a remote source — which is what the topic page's tier table already says ("pending a seek-cost measurement"), so the two now agree. Fixed on both pages in 1a9e196.

---

dev-docs/formats/zip.md
davitf: Fair — the clause was answering a question nobody asked. I meant the *diagnostic's* payload: `SymlinkTargetContext` carries `archive_name`, `member_name`, `member_id` and `reason="password_required"` (`zip_reader.py:1379`) and nothing out of the member. You're right that there is no secret available to leak in the first place, so saying it withheld one implied a risk that cannot exist.

The row now states the mechanism instead — a ZIP symlink's target is the member's *data*, so an encrypted member with no password has nothing to read it from — and says what the diagnostic carries. 1a9e196.

---

dev-docs/formats/zip.md
davitf: Your reading was the right one to worry about, and checking it turned up a second error: the two fields do not belong in one row.

- `comment` — set only when the entry stores one (`zip_reader.py:703`), so absent is the common case and it is the archive's doing, not ours.
- `create_system` — **never** absent. `_CREATE_SYSTEM_BY_VALUE.get(info.create_system, CreateSystem.UNKNOWN)` cannot return `None`, so the `if create_system is not None` guard at line 705 is always true; an unrecognised value maps to `UNKNOWN` rather than to nothing.

Split into two rows in 1a9e196, with "the entry stores none" and "never" respectively.

---

dev-docs/formats/zip.md
davitf: Right, and it is the vocabulary trap the code map now warns about — I wrote "stdlib" for `zipfile` specifically. What is bypassed is `ZipExtFile`; the codec layer still ends at `zlib`, `bz2` and `lzma` for the methods they cover, so the standard library does the actual decoding for DEFLATE, BZIP2 and LZMA either way.

1a9e196 makes the heading "`zipfile`'s own decoders are not used" and says so in the same breath, and the coverage sentence below now reads "`ZipExtFile` handles STORED, DEFLATE, BZIP2 and LZMA" rather than "stdlib decodes".

---

dev-docs/formats/zip.md
davitf: Added in 1a9e196, and verified rather than assumed: with `use_rapidgzip=AcceleratorMode.ON` and `seekable_members=True`, the stream under `ArchiveStream` for a DEFLATE ZIP member changes from `DecompressorStream` to the accelerator's `VerifyingStream`, so the accelerator really does reach ZIP members and not only bare streams. `config.py:147` agrees — the tri-state covers "gzip / zlib / **raw deflate**".

The uniformity sentence now ends with it, since it is the same argument: the codec layer is what puts ZIP on the shared machinery, and the accelerators are part of that machinery.

---

dev-docs/topics/prefixed-archives.md
davitf: It is in `IDEAS.md` — "Makeself-aware payload location", parked from `prefixed-archive-detection`. Added to the topic page in 1a9e196 as a fourth row in §2's table (the heading is now "Ways to find one", and "only the second is shipped" became "only the cued forward scan is shipped", so the sentence survives reordering).

It reads as a non-tier on purpose: a makeself `.run` exports `SKIP` (header line count) and `COMPRESS`, so the offset is `head -n $SKIP | wc -c` and the compressor is named outright — no window, no needle, and no false `1f 8b 08` picked up from the script's own text. It also names `bzip3` / `lzo`, which archivey cannot read, instead of reporting a needle miss. But it only works for stub families somebody taught it, so it can short-circuit the scan and never replace it. The `IDEAS.md` entry's promotion conditions (a corpus of current and old `.run` installers, plus a definition of "looks like makeself" that does not become a second shell parser) are pointed at rather than restated.

---

openspec/specs/format-zip/spec.md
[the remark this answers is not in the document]
davitf: **Fixed** in `1c8af96`, and the audit turned up more than the scenario.

`archivey.create` is not exported — confirmed. But neither is anything else on the write side: `register_writer` is called nowhere, so `_writers` is empty and `writer_for_format` raises `UnsupportedOperationError` for *every* format, not just read-only ones:

```
RAR: UnsupportedOperationError: No write backend registered for format RAR
ZIP: UnsupportedOperationError: No write backend registered for format ZIP
TAR: UnsupportedOperationError: No write backend registered for format TAR
```

So the requirement's second branch — `UnsupportedFormatError` with an install hint for optional write formats — is unreachable too, not just the scenario row. The scenario now names something checkable (`writer_for_format(RAR)`, or any format), and the requirement says plainly that the write half is ABC scaffolding with no registered writer, pointing at the parked design in `investigations/archive-writing-design.md`. The `WriteBackend` ABC itself stays: it is scaffolding this PR deliberately kept.

---

dev-docs/pair-workflow.md
```diff
@@ -62,14 +62,61 @@ Organised, **no-fluff**, rewritten in place — not an append-only log.
...
+| **9. References** | External spec sections *with numbers*, our investigations, upstream issues |
+
+Four rules the shape depends on:
+
+- **Never separate a structural fact from its consequence.** The strongest grouping force
+  in the ZIP material was causal — one property generated eight downstream behaviours. A
+  separate "consequences" section breaks the chain and makes the reader re-derive it.
+- **No performance numbers.** They are the most volatile thing on the page and they rot
+  into a fourth disagreeing source. Verify carries the command instead.
+- **Behaviour here, status behind the link.** The page says what a caller sees and how
```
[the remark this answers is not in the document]
davitf: **Fixed** in `1c8af96` — the lead-in now reads "create a page when a change needs one, not before", keeps the no-page-yet advice for formats that still have none, and points at `formats/zip.md` and `topics/prefixed-archives.md` as the worked examples.

On the F2 tie-in: see that thread. The short version is that the register row would be born stale — #285 and #286 sit directly on this branch and change that behaviour twice.

---

openspec/specs/format-zip/spec.md
```diff
@@ -28,7 +30,7 @@ The ZIP backend SHALL expose these properties for every opened ZIP archive:
 | Access cost | `AccessCost.DIRECT` — independent local file offsets |
 | Stream capability | `StreamCapability.SEEKABLE` |
 | Read source | Seekable only; no implicit buffering/spooling |
-| Write support | Yes, including streaming write |
+| Write support | No — writing is not shipped for any format (`PLAN.md` phase 9) |
```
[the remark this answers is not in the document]
davitf: **Confirmed, and already fixed one PR up.** I reproduced it on this HEAD — patching `this_disk` / `cd_start_disk` to 1 on a valid ZIP gives `opened, members = ['a.txt']` — so the SHALL is false here, exactly as you say.

Not fixed *in* this PR, though, and deliberately (maintainer's call). [#285](https://github.com/davitf/archivey/pull/285) is stacked directly on this branch and implements the check — `_disk_field_is_split` in `zip_reader.py`, with `0xFFFF` treated as the ZIP64 sentinel — and rewrites this same requirement and scenario. Patching it here would edit the lines #285 already rewrote, guaranteeing a conflict on the next restack for zero behaviour change, and the false SHALL predates this PR rather than being introduced by it.

What this PR owes is honesty about the gap, and §5 pays it: "An EOCD declaring a non-zero disk number opens and lists normally — **archivey** — … `format-zip` has a scenario claiming otherwise."

---

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+
+| What you see | Where it lives | More |
+| --- | --- | --- |
+| A ZIP on a pipe or socket cannot be opened at all, in either access mode, and is never buffered for you | **format** | The index is at the end (§1). A native forward-walking reader would change this |
+| One member name whose UTF-8 flag lies makes the **whole archive** unlistable | **library** | Stdlib decodes flagged names strictly while parsing the central directory, so the failure is archive-wide rather than confined to the bad entry. [`open-issues.md`](../open-issues.md) P4 |
+| A `.z01`…`.zip` split set is refused with "rejoin first" | **format** for now | ZIP addresses entries by (disk, offset), which stdlib cannot resolve and naive concatenation cannot fake. [`open-issues.md`](../open-issues.md) P2 |
+| A `.zip.001`…`.zip.00N` split set — 7-Zip's naming — is not recognised as split at all: the first part reports `CorruptionError`, the rest report `FormatDetectionError` | **archivey** | Valid input, wrong error. The `.z01` rule is a filename pattern that does not cover this convention. Unregistered as of this writing; same shape as `open-issues.md` P17 |
```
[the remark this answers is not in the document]
davitf: **Not registering it — closed by the stack instead** (maintainer's call, having weighed the alternative).

The rule you cite is the right one, and the register row would still be the wrong move here: [#285](https://github.com/davitf/archivey/pull/285) and [#286](https://github.com/davitf/archivey/pull/286) are stacked directly on this branch and change this behaviour twice. #285 refuses `.zip.NNN` by name, so the wrong errors you reproduced (`CorruptionError` on part one, `FormatDetectionError` on the rest) become one rejoin-first message; #286 goes further and joins complete sets, so the pitfall stops describing anything at all — that row is rewritten in both PRs. A P-row added here would be opened and closed inside one merge train, and would need two downstream edits to retire it.

Worth noting the "same shape as P17" pointer in the row is what P17 turned out to mean in practice: P17 is the SFX case, and #286 kept that blind spot deliberately rather than widening sibling discovery, so it stays registered where it belongs.

---

davitf:
## Round summary @ `1c8af96`

| ID | Disposition |
| --- | --- |
| F1 EOCD disk-field SHALL is false | **Closed by #285**, not here — see thread. Reproduced on this HEAD first |
| F2 `.zip.001` unregistered | **Closed by #285 + #286**, not registered here — see thread |
| F3 `backend-registry` scenarios `archivey.create()` | **Fixed** — and the requirement's other branch was unreachable too |
| F4 stale SFX needle comment | **Fixed** |
| F5 stale "until the first page exists" | **Fixed** |
| F6 `docs/formats.md` omits `.zip.001` | **Closed by #286**, which rewrites that section for both split families |
| F7 AE-1 evidence | Noted, thank you |

**F3 was bigger than the scenario row.** `archivey.create` is not exported, as reported — but `register_writer` is never called either, so `writer_for_format` raises `UnsupportedOperationError` for *every* format, which makes the requirement's second branch (`UnsupportedFormatError` plus an install hint for optional write formats) unreachable as well. Both now say what is true, pointing at the parked design; the `WriteBackend` ABC stays as the scaffolding this PR meant to keep.

**F4 was the same error as a maintainer finding earlier in this PR.** The `SFX_MAGIC` comment still justified excluding `PK\x05\x06` / `PK\x07\x08` by "they would claim any executable containing those four bytes" — circular, since `PK\x03\x04` does too. It now gives the real reason (a local-header hit has a validator; those two can be confirmed by nothing) and cites the survey: six false `PK\x05\x06` hits across 3 320 ELF/PE files, zero true ones.

**On D1 — taking neither A nor B.** Both gaps are real and both are closed by the two PRs stacked directly on this branch, so registering them here would open and close a row inside one merge train, and rewriting the `format-zip` scenario here would edit the exact lines #285 already rewrote. This PR's obligation is to tell the truth about behaviour, and §5 does for both. That is the maintainer's decision, not mine to change without them.

**Gates:** `./scripts/check.sh` all green; `./scripts/test.sh tests/test_zip.py tests/test_registry.py` → 111 passed. Docs and spec prose only, so I did not re-run `--all-configs` — the last full three-config run on this branch was at `263344c` and passed all three legs.

---

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+Each stage: who does the work, what is ZIP-specific rather than general, what is refused.
+
+### 2.1 Identify
+
+ZIP declares three magics at offset 0 — `PK\x03\x04` (local header), `PK\x05\x06` (empty
+archive) and `PK\x07\x08` (spanned marker) — and five extensions: `.zip`, `.jar`, `.pyz`,
+`.whl`, `.apk`.
+
+Only the local header is offered as a scan needle for prefixed archives. Not because the
+other two are rarer inside an executable — `PK\x03\x04` appears in stubs too, which is what
```
davitf: "Not because the other two are rarer": if they were rarer, that would actually be a reason to use them in addition or instead of the local header.

And I think "scan needle" is too technical, it sounds like a project-internal term. So I think we can simplify this justification. Something like: (please refine)

When searching for a ZIP header in a prefixed archive, we only look for the local header. `PK\x05\x06` would appear before `PK\x03\x04` only for empty archives, which are unlikely. `PK\x07\x08` would appear only in multivolume archives, which are rarely concatenated; we haven't seen them in the wild. [note: worth checking this claim; what happens when we try to create a spanned sfx archive? IIRC we found that the SFX stub gets added as a non-concatenated standalone exe, right?] Neither of these two have extra fields that could help validate them as described below.

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+the validator below is for — but because a local-header hit can be confirmed and theirs
+cannot: there is no cheap structure to check `PK\x05\x06` or `PK\x07\x08` against, and what
+a hit would buy is an empty archive or a spanning marker. A survey of 3 320 ELF and PE files
+found six `PK\x05\x06` matches, every one a string constant parsing to nonsense
+([`topics/prefixed-archives.md`](../topics/prefixed-archives.md) §4).
+
+A needle hit is confirmed by `validate_zip_local_header` (`internal/zip_detect.py`) before it
+is reported: version-needed in range, no reserved general-purpose bits, a known method id, a
+non-empty name, and name+extra within the source. That last pair is what rejects
+`PK\x03\x04` followed by the zero-fill an ELF or PE stub pads with.
```
davitf: do we see the local header followed immediately by a bunch of zeros in stubs? That's slightly suspicious (wouldn't it be in the middle of other strings?), so if real, it's worth stating explicitly.

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+
+The validator's method table is deliberately **wider** than the set archivey can decode: a
+member using a method we refuse to read is still a ZIP, and identity is not a support
+claim.
+
+Two things are ZIP-specific rather than general in the shared detector:
+
+- Prefixed ZIPs are found by the cued forward scan — the cue is the prefix itself, an `MZ`,
```
davitf: Simplify the language, what's a cued forward scan? Maybe clearer:

By default, we only scan for prefixed ZIPs if an executable (MZ, ELF or #! header) is detected at offset 0...  (I said "by default" because the THOROUGH scan would always scan, right?)

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+`ArchiveyConfig.zip_unflagged_fallback_encoding` (default `cp437`, which decodes every
+byte, so no `UnicodeDecodeError` escapes).
+
+Choosing UTF-8 for an unflagged name emits `MEMBER_NAME_ENCODING_INFERRED`; a pure-ASCII
+name does not, because ASCII decodes identically under both and nothing was overridden.
+The sniff is validation rather than guessing: UTF-8 is self-checking, so a clean decode is
+near-conclusive evidence. No equivalent is possible for the legacy tail — see §5.
+
+Backslashes are decoded by origin, not globally. A DOS/FAT-origin entry's `\` is a path
+separator; a Unix-origin entry keeps it as a literal filename character.
```
davitf: we replace \ by / on `name` when it's a path separator but not when it's a literal character, right? worth checking and stating here and/or in the table below

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+
+**Metadata mapping.**
+
+| `ArchiveMember` field | Source | Absent when |
+| --- | --- | --- |
+| `name` | CDH name bytes, decoded as above | — |
+| `raw_name` | The stored bytes, verbatim | — |
+| `mode` | `external_attr >> 16` | The producer was not Unix-like, or `external_attr` is 0 — then `None`, never a substituted default |
+| `modified` / `accessed` / `created` | CDH DOS date-time (naive local, 2-second granularity) ← NTFS extra `0x000A` (UTC) ← Extended Timestamp `0x5455` (UTC), later overriding earlier | 1980 sentinel, or every layer invalid — with `MEMBER_TIMESTAMP_INVALID` |
+| `type` | Unix mode where present, else the directory marker and symlink hints | — |
```
davitf: Worth stating that zipfile doesn't identify symlinks, so we have to do that checking baed on the exposed values

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+separator; a Unix-origin entry keeps it as a literal filename character.
+
+**Metadata mapping.**
+
+| `ArchiveMember` field | Source | Absent when |
+| --- | --- | --- |
+| `name` | CDH name bytes, decoded as above | — |
+| `raw_name` | The stored bytes, verbatim | — |
+| `mode` | `external_attr >> 16` | The producer was not Unix-like, or `external_attr` is 0 — then `None`, never a substituted default |
+| `modified` / `accessed` / `created` | CDH DOS date-time (naive local, 2-second granularity) ← NTFS extra `0x000A` (UTC) ← Extended Timestamp `0x5455` (UTC), later overriding earlier | 1980 sentinel, or every layer invalid — with `MEMBER_TIMESTAMP_INVALID` |
```
davitf: Worth stating (here or above) that zipfile doesn't read these additional date formats, so we have to do the parsing

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+**Producers disagree about encryption defaults.** 7-Zip's `-tzip` default is ZipCrypto and
+`-mem=AES256` selects WinZip AES; stdlib `zipfile` writes neither. That is why the
+encrypted corpus rows shell out to `7z` and skip silently without it. Reading is uneven
+too: Info-ZIP `unzip` 6.00, still the default `unzip` on Debian, cannot read WinZip AES at
+all and exits 81 ("unsupported compression method") on any method-99 member — an archive
+archivey reads may be one the system CLI cannot.
+
+**And about the AES vendor version.** AE-1 stores the plaintext CRC in the headers; AE-2
+zeroes it, because for a very small file the CRC narrows the plaintext independently of the
+cipher. 7-Zip writes AE-2, which is the modern default — but `pyzipper`, the
```
davitf: Do we correctly expose the crc in the hashes for AE-1 and omit for AE-2?

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+  entries, or entries with enormous names and comments, with nothing decompressed. Capped
+  by `ListingLimits`; see [`threat-model.md`](../threat-model.md) O1.
+- **ZIP is the canonical bomb.** Declared sizes are attacker-controlled and so is the
+  stored CRC, so neither bounds the output; the caps are enforced against bytes actually
+  written, plus a live ratio measured from bytes consumed. Nesting is not tracked — a
+  zip-of-zips amplifies one level at a time (O6).
+- **Overlapping entries** are a distinct crafted shape, caught by stdlib's open-time
+  overlap guard and translated to `CorruptionError`.
+- **ZipCrypto password confirmation is a measurable side channel by design**: the STORED
+  path reads the member to disambiguate candidates. Documented cost, not a leak.
```
davitf: what is a "measurable side channel"? I didn't understand this part

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+  [ADR 0013](../decisions/0013-cross-platform-name-safety-policies.md) and O3.
+
+## 5. Pitfalls
+
+*Where it lives*: **format** — inherent, no implementation fixes it · **library** — stdlib
+`zipfile`'s behaviour, fixable only upstream or by replacing it · **archivey** — ours.
+
+| What you see | Where it lives | More |
+| --- | --- | --- |
+| A ZIP on a pipe or socket cannot be opened at all, in either access mode, and is never buffered for you | **format** | The index is at the end (§1). A native forward-walking reader would change this |
```
davitf: I'd classify it as format / library. And expand as something like "A native forward-walking reader would allow reading members in order, but some metadata lives only on the end directory"

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+| One member name whose UTF-8 flag lies makes the **whole archive** unlistable | **library** | Stdlib decodes flagged names strictly while parsing the central directory, so the failure is archive-wide rather than confined to the bad entry. [`open-issues.md`](../open-issues.md) P4 |
+| A `.z01`…`.zip` split set is refused with "rejoin first" | **format** for now | ZIP addresses entries by (disk, offset), which stdlib cannot resolve and naive concatenation cannot fake. [`open-issues.md`](../open-issues.md) P2 |
```
davitf: library, right? Our native parser would be able to read these

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+| archivey's codec layer for member data | Stdlib decodes four methods; the codec layer decodes seven, and unifies CRC verification and error translation with the other backends | Staying on `ZipExtFile`, which left ZIP advertising codecs it could not decode |
+| Refuse a non-seekable source rather than spool it | Silent buffering hides an unbounded memory or disk cost the caller did not ask for | Transparent `SpooledTemporaryFile`; still possible later as an explicit opt-in ([ADR 0010](../decisions/0010-no-silent-buffer-nonseekable.md)) |
+| Seeking and concurrent member streams off by default, on ZIP too | Concurrent opens genuinely are free here, and a rewind re-decodes one member rather than a whole solid block — which is the trap, not the reassurance: leaving them on lets someone test on ZIP and ship a footgun on TAR or 7z. The strict default is reversible before 1.0; the permissive one is not | Enabling them where they are cheap ([ADR 0003](../decisions/0003-member-streams-opt-in.md)) |
```
davitf: Concurrent is not always free; they're free only if the zip is opened via a file. If opened via a stream, it would require stream multiplexing (like any other format).

And the trap is that rewind re-decodes from the beginning of the stream, which is why we always require opt-in; not just because it may be even more expensive in other formats

dev-docs/formats/zip.md
```diff
@@ -0,0 +1,408 @@
...
+- **Overlapping entries** are a distinct crafted shape, caught by stdlib's open-time
+  overlap guard and translated to `CorruptionError`.
+- **ZipCrypto password confirmation is a measurable side channel by design**: the STORED
+  path reads the member to disambiguate candidates. Documented cost, not a leak.
+- The real-world case behind the cross-platform name policies arrived as a ZIP: a macOS
+  archive containing a `stuff_etc.` folder, a trailing dot Win32 silently trims. Because
+  the offending segment is a directory, rejecting it takes every member beneath it. See
+  [ADR 0013](../decisions/0013-cross-platform-name-safety-policies.md) and O3.
+
+## 5. Pitfalls
```
davitf: is "Pitfalls" a good name here? it seems a mix of pitfalls (things a user should be aware of or can't do) and open issues. Rename, or split into two lists?

dev-docs/formats/zip.md
davitf: **Real, and your instinct that it needed checking was right twice over** — the shape exists, and the page was crediting the wrong check for it. Measured in `d86f9d5`.

`PK\x03\x04` occurs **148 times** across 3 332 ELF and PE files under `/usr/bin`, `/usr/lib`, `/usr/local` and `/opt`. **23 of the 148** are the magic followed by a run of zeros, so it is not a story about padding — it happens about one time in six. The other 125 are your other guess: the magic landing in the middle of unrelated bytes.

But the validator rejects all 148 on `version_needed` alone, which has to fall in 10–99 and instead reads as whatever follows the magic — zeros included, since 0 is below the floor. The name-length and name+extra checks **never fire on this corpus**. So "that last pair is what rejects `PK\x03\x04` followed by the zero-fill" named a check the input never reaches; the page now says the version floor does nearly all the work, and that the later checks exist for shapes this corpus does not contain, such as a header truncated by end of file.

---

dev-docs/formats/zip.md
davitf: Agreed on both counts — "rarer would be a reason *to* use them" is the flaw in the argument, and "scan needle" was project jargon. Rewritten along your lines in `8bc6143`, with your spanned-SFX claim measured in `d86f9d5`.

**Your recollection is right.** `7z a -tzip -sfx -v40k` produced a 416 KB `out.zip` — the stub, standalone, containing **zero** ZIP signatures of any kind — beside `out.zip.001…003`, which join into an ordinary byte-split ZIP starting with `PK\x03\x04` and carrying no spanning marker anywhere. So a concatenated stub-plus-spanned-ZIP is not a shape that tool produces at all, which is the same thing `open-issues.md` P17 describes for 7z and RAR.

For the record on the other half: `zip -s` really does put `PK\x07\x08` at offset 0 of `.z01` and nowhere else — one occurrence in the whole set, none in `.z02` or the final `.zip`.

---

dev-docs/formats/zip.md
davitf: Reworded in `8bc6143` to roughly your sentence, minus the "by default" — because the answer to your parenthetical is **no**.

`THOROUGH_BUDGET` raises `max_probe_links` (8 → 32), `max_index_bytes`, `completion_window_bytes` and `collect_nonmaximal_candidates`, several of which are marked "not honoured yet". It keeps `max_scan_bytes` at the same 2 MiB as BALANCED, and it does not touch the gate at all: `if cue is not ExecutableCue.NONE` at `detection.py:609` runs before the budget is consulted, and the budget only sizes the window inside it. So no mode a caller can select makes the scan unconditional — that is the exhaustive tier, designed and not shipped. `d86f9d5` puts that detail on the page so the claim is checkable rather than asserted.

---

dev-docs/formats/zip.md
davitf: Yes, both halves — and it is pinned rather than assumed.

`zip_reader.py:668` surfaces the CDH CRC as `hashes["crc32"]`, and the AE-2 branch withholds it because the format zeroes the field. Reading the eight `pyzipper` archives I built for the AE-1 work, `crc_exposed` came back `True` for every AE-1 member (including AE-1 over LZMA) and `False` for every AE-2 one. `test_external_ae1_archive_from_pyzipper` asserts `HashAlgorithm.CRC32 in member.hashes` on the third-party AE-1 fixture, `test_7z_aes_zip_roundtrip` asserts `not in` on 7z's AE-2 output, and `test_handbuilt_aes_roundtrip` checks both directions across 128/192/256 × STORED/DEFLATE.

It is also *verified*, not merely exposed: corrupting the stored CRC on an AE-1 member raises `CorruptionError: Digest mismatch for 'crc32'`. `8bc6143` says both on the `hashes["crc32"]` row.

---

dev-docs/formats/zip.md
davitf: Right on the second point, and it was the more important one — the row had the reasons in the wrong order. The rewind re-decoding from the member's start is why the opt-in exists *here*, not a footnote to protecting TAR/7z users; [ADR 0003](https://github.com/davitf/archivey/blob/main/dev-docs/decisions/0003-member-streams-opt-in.md)'s own context line names "O(n) rewinds" first and the row had dropped that. Fixed in `8bc6143`.

On concurrency, one correction to the mechanism: archivey multiplexes over a **single** handle in both cases, not only for streams. `zip_reader.py:424` keeps one `fp` and leaves reads to stdlib's `_SharedFile` lock (with an extra lock around `open`/`close` for free-threaded builds), so a path source gets seek-multiplexing exactly like a caller-supplied stream — it never opens a second handle. What makes it cheap for ZIP is not the source kind but that a ZIP source is always seekable, since we refuse the other kind at open. So "free only via a file" is not quite the split; the split is seekable-vs-not, and ZIP has no non-seekable case to worry about.

---

dev-docs/formats/zip.md
davitf: Renamed to **Sharp edges** in `5d26b8a`, kept as one table.

The name was the whole problem: "Pitfalls" describes only the live-with-it half, so the open issues sitting next to them read as misfiled. "Sharp edges" covers both, and it is the phrase `pair-workflow.md` already used to describe what the section holds.

Not split, on the reasoning that a reader arrives with a *symptom* and does not yet know which kind it is — which is exactly what the where-it-lives tag tells them, per row, once they have found it. Two lists would make them pick before they know the answer. The **library** rows also sit on the boundary rather than one side of it: unfixable while we wrap `zipfile`, fixable the day we replace it, and after the two retags in this round that is four of the nine rows. The template now records that reasoning so the next format page does not re-argue it.

---

dev-docs/formats/zip.md
davitf: Retagged **format / library** with your expansion, in `8bc6143`. The index living at the end is the format; refusing *everything* rather than walking local headers forward is stdlib — a native reader could stream members in archive order, and would only have to admit that what lives solely in the central directory (the authoritative CDH run, external attributes, comments) is unavailable until it reaches the end.

---

dev-docs/formats/zip.md
davitf: Yes — **library**, retagged in `8bc6143`. The format defines `(disk, offset-within-disk)` addressing perfectly well and a native reader that understood disk numbers could follow it across the parts; `zipfile` cannot, and a linear join is not a substitute for it. "**format** for now" was a hedge that put the blame in the wrong place.

---

dev-docs/formats/zip.md
davitf: Fair — it was jargon standing in for the actual mechanism. Rewritten in `8bc6143`.

What it meant: with several password candidates, the one-byte ZipCrypto verifier is too weak to choose between them, so a **STORED** member has to be read through and its CRC compared. Anyone who can time the call, or watch how much of the archive is read, learns that some candidate got past the byte check and that the member is stored. They learn nothing about the password or the plaintext. It is a cost we accept and document rather than a leak we missed — the alternative is accepting a wrong password one time in 256 — and the page now says that instead of the phrase.

---

dev-docs/formats/zip.md
davitf: Confirmed and stated in both places, in `8bc6143`. `zip_reader.py:618-620` converts `\` to `/` only when `create_system` is in `_BACKSLASH_SEPARATOR_SYSTEMS` (DOS/FAT-family origins); a Unix-origin entry keeps it as a literal filename character. `raw_name` is never rewritten either way, so the stored bytes survive the choice — the prose above the table says it, and the `name` and `raw_name` rows now say it too.

---

dev-docs/formats/zip.md
davitf: Stated in `8bc6143`, on the `type` row and in the lead-in to the table. `zipfile` has `is_dir()` and nothing for symlinks — no `is_symlink`, no mode interpretation — so archivey reads `stat.S_ISLNK` off `external_attr >> 16` for Unix-origin entries (`zip_reader.py:610`). Stdlib hands over the central-directory fields and the raw `extra` blob; the classification is ours.

---

dev-docs/formats/zip.md
davitf: Stated in `8bc6143`, and stronger than "doesn't read these additional formats" — `zipfile` parses **no** extra field except ZIP64. Its `_decodeExtra` handles `0x0001` and ignores everything else, so `0x5455` (Extended Timestamp) and `0x000A` (NTFS) never appear anywhere in its source. All `ZipInfo` offers is `date_time` from the DOS field and the raw `extra` bytes; archivey does the parsing, the UTC conversion and the precedence ordering.

---

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,185 @@
...
+`SFX_MAX` is shared by three call sites on purpose: the detector, `rar_parser`'s own scan
+and `sevenzip_parser`'s. Three separate bounds drift, and a stub the RAR parser accepts but
+the detector does not is a file that opens under `format=RAR` and fails under auto-detect.
+
+**The cue is two-tiered.** `WEAK` is the bare prefix — `MZ`, `\x7fELF`, or `#!`. `STRONG`
+means the header structurally parses: a DOS header whose `e_lfanew` actually points at
+`PE\0\0`, an ELF identification block whose class, encoding and version are all valid, or a
+Mach-O header whose `cputype`/`filetype` or fat arch table parse. A strong cue with no
+needle hit suppresses the content probes; a weak one does not, because `MZ` is two bytes and
+a genuine Brotli stream starting with them must stay detectable.
```
davitf: I think we're missing the reasoning for the two tiers here, do we have it stored in our ADRs or specs? I think it's something like: the content probes have many false positives (particularly Brotli, which matches some executable headers), so we skip probing executables to avoid identifying them as corrupted archives. Please research

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,185 @@
...
+
+Mach-O raises no weak cue at all — `ca fe ba be` is also the Java class-file magic, and a
+weak cue there would put every `.class` file through the scan.
```
davitf: "raises no weak cue" leaves it unclear whether it's always strong or always ignored

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,185 @@
...
+The numbers behind the current gate, measured on a `/usr` tree:
+
+- Widening the cue from "executable" to "prefix-shaped" (adding Mach-O and `#!`) enrolled
+  **742 more files against 2 868 already scanned**, about 26% more. The newly enrolled ones
+  are mostly small scripts — median 2 959 bytes, one file in 734 large enough to reach the
+  window at all — so the whole tree cost **10.3 MiB** more, not the 2 MiB per script the
+  bound alone suggests.
+- The scan's I/O is now **1× the window** on a miss, because every front-of-source read goes
+  through one monotone prefix workspace that only fetches the delta. Before that workspace,
+  the geometric peek loop re-requested from byte 0 each step and a miss billed 1.66×.
```
davitf: no need to talk about previous limitations and implementation details here. it's expected that if we scan up to 2MiB, that's the amount we'll have to read

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,185 @@
...
+
+The tail probe is gated for the opposite reason. It cannot be cued — a ZIP with `PK\x03\x04`
+at offset 0 is already found by near magic, so cueing it on a front hit would only find ZIPs
+that were already found. Uncued means one tail read on *every* source. On a hit that read is
+nearly free, because whatever opens the archive reads the EOCD anyway; on a miss it is pure
+waste, and most sources are not ZIPs. Which entry point the caller used does not change that
+arithmetic: `open_archive` runs the same detection as a bare `detect_format()` and wastes the
+same read on the same files. The open cost question is what that wasted seek is worth on a
+cold cache or a remote source, which nobody has measured.
```
davitf: hard to understand, please try to simplify as well

dev-docs/topics/prefixed-archives.md
```diff
@@ -0,0 +1,185 @@
...
+| ZIP | `PK\x03\x04` only | Cheap LFH sanity: version-needed in range, no reserved general-purpose bits, known method id, non-empty name, name+extra inside the source. Rejects `PK\x03\x04` plus the zero-fill an ELF or PE stub pads with. EOCD confirmation is the tail probe's job, not this tier's |
+| 7z | `37 7A BC AF 27 1C` | `StartHeaderCRC` over the 20-byte StartHeader, plus `offset + 32 + NextHeaderOffset + NextHeaderSize` landing at EOF |
+| RAR | RAR3 and RAR5 markers | The CRC-checked main header that follows the marker |
+
+ZIP's other two magics are deliberately **not** needles — and not because they turn up in
+executables more often than `PK\x03\x04` does. They turn up about as often; the difference is
+that a local-header hit can be *confirmed* and these two cannot. There is no cheap structure
+to check them against, and what a hit would buy is an empty archive or a spanning marker, so
+every match inside the window would be taken on faith. The survey below is what that faith
+would have cost on one machine.
```
davitf: already expanded in the zip document, not needed here I think (or just point to the zip doc)

dev-docs/topics/prefixed-archives.md
davitf: Researched — it is written down, in three places, and your reconstruction is close but wrong in two ways that matter. Added to §2 in `d51a9a8`.

**Where it lives:** [`format-detection`](https://github.com/davitf/archivey/blob/main/openspec/specs/format-detection/spec.md) §"Executable-looking prefixes must not silently become a wrong stream format" (the graded rule, settled by measurement in the archived `sfx-format-detection` design), the `ExecutableCue` docstring in `internal/sfx.py:112`, and the measurements in `investigations/brotli-content-probe-results.md`.

**First correction: the failure is a wrong *success*, not "identifying them as corrupted archives."** An error would be acceptable. What the rule prevents is a probe claiming a stream codec on a stub so that `open_archive` *succeeds*, handing back a fabricated single-file member — `installer.uncompressed` — with no complaint. That is why a structurally confirmed executable containing no archive suppresses the probes outright.

**Second: it cannot be a flat "skip probing executables",** which is what the two tiers are for. Two or four bytes are not proof, so refusing a probe on a bare `MZ` would reject a genuine Brotli stream whose first bytes happen to look executable. The spec also forbids fixing it by tightening the probe instead: measured, a 16× larger prefix moves the false-positive rate 8.27% → 8.13%, and demanding decoded output loses real streams roughly one for one.

**And the premise itself does not hold, which I think is worth having on the page.** Real PE binaries do *not* trip the Brotli probe: **0 of 100** across MSVC, Go's own linker, MinGW-w64 and every `MZ` file on the survey machine, because the canonical `MZ\x90\x00` DOS-stub prologue fails Brotli's nibble check (97/100 carry `90 00`). The one measured collision between a probe and an executable is a different probe — a zero-filled Mach-O stub came back `LZMA_ALONE`, since `cf fa ed fe` passes the LZMA-Alone properties gate. So strong-cue suppression is an invariant held on principle, not a patch for a measured collision rate; the probes' real false-positive problem is arbitrary data, which is P12 / O10 and stays scoped there.

That also caught a section heading contradicting itself: "Cost is what tiers this, not false positives" is true of *whether to scan*, and not of probe suppression, which is a correctness call. §3 now scopes its claim.

---

dev-docs/topics/prefixed-archives.md
davitf: It is **always strong or ignored, never weak** — clarified in `d51a9a8`. `executable_cue` (`internal/sfx.py:180`) returns `STRONG` when `_is_macho(prefix)` parses and otherwise falls through to `NONE`, so a Mach-O-looking file that does not parse is not scanned at all. The page now says that instead of only saying which tier it skips.

---

dev-docs/topics/prefixed-archives.md
davitf: Dropped in `d51a9a8`. Agreed — a reader who has just been told the bound is 2 MiB expects a miss to cost 2 MiB, so the bullet was explaining why the number is now what it should have been all along, which is changelog rather than reference.

---

dev-docs/topics/prefixed-archives.md
davitf: Replaced with a pointer in `d51a9a8`: "ZIP's other two magics are deliberately **not** needles; `formats/zip.md` §2.1 has the reasoning, which is ZIP's own. The survey below is the evidence behind it."

That is also the page-split rule working as intended — the reasoning turns on ZIP's own records, so it belongs on the format page, while the survey is shared evidence about scanning binaries and stays here.

---

dev-docs/topics/prefixed-archives.md
davitf: Rewritten in `d51a9a8`. It was eight lines to say four things, and the "cannot be cued" argument was stated before the reader knew why anyone would want to cue it:

> **The tail probe is gated the other way round.** Nothing can cue it: the front of a prefixed ZIP looks like whatever the stub is, and a ZIP whose front *does* say `PK\x03\x04` was already found by near magic. So it would cost one tail read on every source — free when it finds a ZIP, since the reader goes on to read the EOCD anyway, and wasted otherwise, which is most files. What that wasted seek costs on a cold cache or over a network is unmeasured, and that is the open question. Which entry point the caller used is not: `open_archive` and `detect_format()` run the same detection and waste the same read.

The `open_archive` sentence stays because it was the thing you had to correct me on two rounds ago, and losing it would invite the same wrong reading back.

---

PR #285, opened 2026-09-02
fix(zip): refuse split volumes via EOCD disks and .zip.NNN names

davitf:
## Summary

Fixes two “valid input, wrong error” ZIP split-detection bugs and thins the matching `format-zip` requirement.

1. **Info-ZIP final `.zip` part** — EOCD `this_disk` / `cd_start_disk` non-zero → `UnsupportedFeatureError` at open. `0xFFFF` is the ZIP64 sentinel.
2. **Split segment names** — Info-ZIP `.zNN` and 7-Zip `.zip.NNN` refused in `open_archive` before detection (magic-less middle/last parts are not `FormatDetectionError`).

Spec: MUST-refuse prose kept; scenario matrix collapsed; handbook §5/§7 updated; OpenSpec change archived in this PR. Test pointers handbook-only.

## Review dispositions

| ID | Disposition |
| --- | --- |
| F1 | Fixed (D1=a) — `is_zip_split_segment_name` early-refuse; `.z02` pinned |
| F2 | Fixed — genuine ZIP64 EOCD+locator sentinel test |
| F3 | Fixed — empty / bare EOCD pinned |
| F4 | Fixed — shared `ZIP_MULTI_VOLUME_MSG` |

## Test plan

- Synthesised tests (no `zip` binary gate); real `zip -s` / `7z -v` exercised locally
- `./scripts/check.sh --fix` — passed
- `./scripts/test.sh --all-configs` — passed (2167)

Base: `claude/zip-knowledge-inventory-yixwjs`.

<div><a href="https://cursor.com/agents/bc-9d1d7377-24e6-44fe-9f47-3692606bd6d2?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9d1d7377-24e6-44fe-9f47-3692606bd6d2&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


davitf:
Two findings against `2e2ee76`, measured with real `zip` / `7z` output rather than synthesised parts. The EOCD mechanism itself checks out (see the note at the end).

### G1 🟡 `[important]` · CONFIRMED — `\.z\d{2}$` misses splits with more than 99 parts

`src/archivey/internal/zip_detect.py` — `_ZIP_SPLIT_SEGMENT_RE` requires **exactly** two digits. Info-ZIP does not stop at `.z99`:

```
$ head -c 7500000 /dev/urandom > many_payload.bin
$ zip -q -s 64k many.zip many_payload.bin
$ ls many.z* | wc -l
115
$ ls many.z* | grep -vE '\.z[0-9]{2}$|\.zip$' | head -3
many.z100
many.z101
many.z102
```

Observable result on this branch:

```
many.z01   -> UnsupportedFeatureError: Multi-volume (split/spanned) ZIP...   correct
many.z100  -> FormatDetectionError: Could not detect archive format...      the bug this PR fixes
many.zip   -> UnsupportedFeatureError: Multi-volume (split/spanned) ZIP...   correct, via EOCD disks
```

So the "valid input, wrong error" case survives for any set over 99 parts — which is a 6.4 MB archive at the default 64 KB minimum split, not an exotic size.

**Fix:** `\.(?:z\d{2,}|zip\.\d{3,})$`. Worth a test case at `.z100` next to the existing `.z02` pin, since the boundary is exactly where the current regex stops.

### G2 🟡 `[important]` · CONFIRMED — `.zip.NNN` is refused, but the identical 7-Zip split is supported for `.7z`

7-Zip's `-v` is a **raw byte split**, not a spanned set. Concatenating the parts yields a valid archive:

```
$ 7z a -tzip -v40k sp.zip big.bin s.txt      # -> sp.zip.001 .002 .003
$ python -c "...b''.join(parts) -> rejoined.zip"
zipfile.ZipFile('rejoined.zip').namelist() -> ['big.bin', 's.txt']
```

And the same tool, same split style, `.7z` container already auto-joins today through `volumes.py`:

```
sv.7z.001    opens: ['big.bin', 's.txt']      # _7Z_VOLUME_RE + ConcatenatedFile
sp.zip.001   -> UnsupportedFeatureError: ...rejoin first
```

So a user who ran `7z a -v` gets a working archive for `.7z` and a refusal for `.zip`, and the refusal says "rejoin first" for something the library rejoins for the sibling format. That is a format-parity surprise rather than a bug — before this PR `.zip.001` raised `FormatDetectionError`, so nothing regressed.

The distinction this PR collapses is worth keeping explicit, because the two families need **opposite** answers:

| Family | Produced by | Structure | Right answer |
|---|---|---|---|
| `.z01`…`.z99`/`.z100`… + final `.zip` | `zip -s` (Info-ZIP) | true multi-disk spanned set | **refuse** — not concatenable, as this PR does |
| `name.zip.001`… | `7z -v` | raw byte slices of one file | joinable, exactly like `.7z.NNN` |

**Decision needed, not just a fix:** either extend the `volumes.py` sibling-joining to `.zip.NNN`, or keep the refusal and record why (a line in `formats/zip.md` §3, where split naming already lives). What should not stand is the two 7-Zip splits behaving differently with nothing saying so.

### Not a finding — the EOCD check holds up

I went after `_classic_eocd_declares_split` expecting a false positive from a `PK\x05\x06` planted in the archive comment, and it does not break: stdlib's `_EndRecData` takes the same last-occurrence `rfind`, so this check always inspects the record stdlib actually parsed, and the two cannot disagree about which EOCD is real.

That invariant is what makes the naive `rfind` safe here, and it is not obvious from the code — the docstring explains the `0xFFFF` sentinel but not this. One sentence ("same last-occurrence rule stdlib uses, so we read the record it parsed") would save the next reader the trace.

The `0xFFFF` sentinel handling itself is right, and the comment explaining why a naive `!= 0` would refuse legitimate ZIP64 archives is the useful kind.

### What I ran

`./scripts/check.sh` — all checks passed. `./scripts/test.sh tests/test_zip.py` — 62 passed. Repros above executed against `2e2ee76`.

Note on scope: the explicit-`format=` precedence issue raised earlier in this review is **not** repeated here — it is parked in `dev-docs/IDEAS.md` under "Volume-shaped names: detect first, then upgrade", and the parked shape resolves it as a side effect (detection is skipped when the caller supplies a format, so nothing upgrades).

---

davitf:
Second review pass against `27857ec`, focused on the doc updates. One live code finding, four doc findings, and the mechanism checks out — details below.

### H1 🟡 `[important]` · CONFIRMED — the filename refuse overrules an explicit `format=`

D1=a moved the name check into `open_archive` **before** format resolution, and it never consults `format`. So a file that merely *ends* in a volume-shaped name is refused as a split ZIP even when the caller said it is something else.

Measured on `27857ec` against a genuine `.tar.gz` renamed:

```
mytar.z01                                  -> UnsupportedFeatureError: Multi-volume (split/spanned) ZIP...
mytar.z01, format=ArchiveFormat.TAR_GZ     -> UnsupportedFeatureError: Multi-volume (split/spanned) ZIP...
detect_format("mytar.z01")                 -> ArchiveFormat.TAR_GZ
```

Same three on base `14b48f7`: `['hello.txt']`, `['hello.txt']`, `TAR_GZ`. Identical for `mytar.zip.001`. So this is a regression, and `detect_format` and `open_archive` now disagree about the same bytes.

**Why it matters beyond the contrived filename.** The check sits at `core.py:250`, immediately above the block #225 added for the directory case, whose comment reads:

> Silently overruling `format=` here would hand back a reader over the directory tree for a caller who asserted something else -- the wrong data, succeeding. **Every other explicit-format conflict is refused loudly**; so is this one.

The new check is the counter-example, three lines above the comment asserting it does not exist. `open-issues.md` P8 settled this class: an explicit `format=` is either honoured or refused loudly *as a conflict*. Raising a ZIP error at a caller who asked for TAR_GZ is neither.

**This was not resolved by parking it.** The earlier pass declined to repeat the precedence issue on the grounds that the parked IDEAS entry "resolves it as a side effect (detection is skipped when the caller supplies a format, so nothing upgrades)". That is true of the *parked* shape and not of the shipped one — the shipped check runs before format resolution and reads only the name. The IDEAS entry as written does not mention `format=` either, so the parking does not record the issue.

**Fix direction.** Skip the name refuse when `format` is explicitly supplied and is not ZIP (two lines, at the same site). D1=a exists to fix the error class on the *auto-detect* path; an explicit `format=` never had that problem. Mirroring #225 exactly — `ArchiveyUsageError` naming the conflict — is the stricter alternative.

Practical risk is low: I cannot name a real non-ZIP file ending `.z01` or `.zip.NNN`. The argument is contract consistency, and the fix is small.

### H2 🟢 `[nit]` — §2.2's lead sentence is now false

`dev-docs/formats/zip.md` §2.2 opens **"Split sets are rejected before anything else."** and then describes a second check that runs *after* stdlib has opened the archive and parsed the entire central directory. Two mechanisms at two different times; the lead claims one. Suggest **"Split sets are refused by two different checks."**

### H3 🟢 `[nit]` — §7 files false-positive guards under a "refused" claim

The Verify row `Split segment refused` now lists five tests, two of which pin the opposite claim: `test_eocd_zip64_disk_sentinel_still_opens` and `test_plain_prefixed_and_empty_zip_still_open` assert that archives **still open**. Worth two rows — "Split sets refused" and "Split checks do not fire on single-volume archives" — since the second is the one that catches a regression in the disk-field logic.

### H4 🟢 `[nit]` — a PR number inside a handbook page

§3 ends `(see PR #285 G2)`. The handbook is read long after the PR conversation is cold, and a `G2` id inside a review thread is not resolvable from the page. `zip.md`'s convention elsewhere is to link a register row. Either register the join-vs-refuse question and link that, or state it in one clause without the reference. `IDEAS.md`'s `#285 D1 discussion` is the same shape but matters less — that file is a backlog.

### H5 🟢 `[nit]` — `#### Scenario: multi-volume ZIP matrix` no longer contains a matrix

Thinning it to prose was right; the heading still says matrix. `#### Scenario: multi-volume ZIP refusal` validates the same.

### Checked and clear

- **Both bugs fixed, verified on real tool output**, not synthesised parts. `zip -s 64k` set: `set.z01`, `set.z04`, `set.zip` all refuse. `7z -v40k` set: `.001` and `.005` both refuse. Base gives `CorruptionError`, `FormatDetectionError` and a convincing-listing-then-`CorruptionError` respectively.
- **No false positives**: plain, empty (bare 22-byte EOCD), prefixed (shebang + ZIP), and a genuine ZIP64 archive with locator and 70 000 entries all still open and list.
- **The `rfind` invariant is real.** I read `zipfile._EndRecData`: `maxCommentStart = max(filesize - (1 << 16) - sizeEndCentDir, 0)` then `data.rfind(stringEndArchive)` — byte-identical window and rule to `_classic_eocd_declares_split`. The docstring's claim that the two cannot disagree about which EOCD is real holds.
- **Full suite on this branch: 2680 passed, 23 skipped.** All five test names cited in §7 exist.
- **§5's collapse is correct.** Both `archivey`-tagged rows describe behaviour that is genuinely fixed, and the surviving row's `**format** for now` is the right tag for a refusal stdlib cannot lift.
- **CI red is not this PR.** `ubuntu-latest / py3.12 / all` failed in 16s on `apt-get update` — `403 Forbidden` from `packages.microsoft.com`, before any test ran. Re-run it.

### What I ran

Real `zip -s 64k` and `7z -tzip -v40k` sets; base-vs-HEAD comparison for every claim above; `pytest` full suite on the branch worktree; `zipfile._EndRecData` read directly.

---

PR #286, opened 2026-09-03
feat(zip): join 7-Zip .zip.NNN split sets

davitf:
**Stacked on #285** (`claude/zip-split-volume-detection`) — this modifies the refuse
paths that PR adds. Review it after #285. Rebased onto #285's H1 fix (`e1a1f24`), so the
diff here is only the joining work.

## The asymmetry

7-Zip's `-v` is a raw byte split, not a spanned archive. Measured:

```
7z a -tzip -mx0 -v40k set.zip span.bin         -> set.zip.001 … .004
b"".join(part.read_bytes() for part in parts)  -> valid ZIP; zipfile lists both members
rejoined EOCD disk fields                      -> (0, 0), an ordinary single-volume ZIP
```

archivey already rejoined the identical split for 7z — `sv.7z.001` opens and lists —
while `set.zip.001` raised `UnsupportedFeatureError`. Same tool, same flag, same byte
slicing, opposite answers, and the message told the user to rejoin something the library
rejoins for the sibling format.

| Family | Made by | Structure | Answer |
| --- | --- | --- | --- |
| `name.z01 … name.zip` | `zip -s` | true spanned set, entries at (disk, offset) | refuse (unchanged) |
| `name.zip.001 …` | `7z -v` | raw byte slices of one ZIP | **join** (this PR) |

## What changed

**One pattern, not a third copy.** `_7Z_VOLUME_RE` became `_NUMBERED_VOLUME_RE` covering
`.7z|.zip`, and `_validate_7z_volume_sequence` became `_validate_numbered_volume_sequence`
— its name and error message were 7z-specific and would have started lying the moment ZIP
used it. `review/backlog.md` already carries a registered duplication item about ZIP
local-header layout in four places; a third near-identical discovery branch here would
have been the next one. `.zNN` deliberately does not match the widened pattern.

**The trap.** `resolve_source` runs *before* the refuses added by #285, and a joined
source keeps part one's name — so `is_zip_split_segment_name(archive_name)` fires on the
very set that was just rejoined. Both sites (`core.py`, `ZipReader.__init__`) now key off
whether the set was joined rather than the name alone. Verified by running, not reading.

That check sits alongside #285's H1 condition rather than replacing it, so the refuse now
requires all three: not joined, volume-shaped name, and no conflicting explicit
`format=`. H1's `test_volume_shaped_name_honours_explicit_non_zip_format` still passes
unchanged.

**Still refuses**, same message: Info-ZIP `.zNN`; its final `.zip` (EOCD disk fields); a
`.zip.NNN` part whose siblings are not on disk. Checked against a real `zip -s 64k` set —
`.z01`, `.z02` and `.zip` all refuse.

**Unrelated latent bug, fixed because this widens it.** The ordering key scanned for the
first `.partN` *anywhere* in the name, so every part of `my.part1.zip.001 … .003` parsed
as `1` and `sorted()` fell back to `iterdir` order — arbitrary, and silently wrong bytes
out. Pre-existing for `.7z.NNN`; more reachable with ZIP names. Each key now reads its
number from the pattern that classified the name (`_numbered_part_number`,
`_rar_part_number`, `_rnn_part_number` — one idiom, no dead code left behind).

## Two calls made from precedent rather than escalated

Both were flagged as likely escalations; both are answered exactly by the `.7z.NNN`
sibling, and matching it is this PR's whole thesis:

- **A `.zip.NNN` set that does not concatenate into a valid ZIP.** A numbering gap is
  `TruncatedError` (the volumes layer, as for 7z); bad *content* is whatever the backend
  says, i.e. `CorruptionError`. No new rule.
- **`is_multivolume` for a joined set.** `True`, plus `extra["zip.volume_count"]` — what
  `sevenzip_reader` reports for `.7z.001`, which is equally "one archive sliced in
  pieces". It describes how the archive arrived, not its ZIP structure. Say the word if
  you would rather it read `False` and let `volume_count` carry it alone.

## One claim I had to walk back

I first wrote that Info-ZIP's parts "cannot be concatenated into anything stdlib can
address". That is false, and a reviewer would have disproved it in one command: a real
three-disk `zip -s 64k` set *does* rejoin into something `zipfile` lists correctly. The
difference is guarantee, not outcome — entries carry a starting disk, so a linear join
only coincides with the archive's own addressing. The docs and the spec now say
"reconstructs only by coincidence", and the refusal is justified by the non-zero EOCD
disk fields (`(2, 2)` on that fixture) rather than by an unreadable rejoin.

## Tests

Red–green. The joined set is built with real `7z a -tzip -mx0 -v40k`. The Info-ZIP shapes
are **synthesised**: `zip` is not installed on CI, macOS or Windows, so a `zip -s` fixture
would skip on every machine that matters — though I did use a real one locally for the
verification below.

Covered: opens from `.001` and from a middle part; a member spanning part boundaries reads
byte-identical; a missing part is `TruncatedError`; Info-ZIP `.z01`/`.z02`/`.zip` still
refuse; a lone `.zip.001` with no siblings still refuses; plain single-volume ZIP
unaffected; `cost` and `is_multivolume` assertions; the `my.part1.zip.NNN` mis-sort
regression, also as a Hypothesis case.

Manual verification against real tool output, since it is the one a naive fix gets wrong:

```
part sizes: [40960, 40960, 40960, 27403] -> span.bin (150000 B stored) crosses 3 boundaries
  set.zip.001: members=['span.bin', 'tail.txt'] byte-identical=True is_multivolume=True {'zip.volume_count': 4}
  set.zip.002: members=['span.bin', 'tail.txt'] byte-identical=True is_multivolume=True {'zip.volume_count': 4}
  set.zip.004: members=['span.bin', 'tail.txt'] byte-identical=True is_multivolume=True {'zip.volume_count': 4}
Info-ZIP spanned set (real `zip -s 64k`):
  iz.z01: UnsupportedFeatureError   iz.z02: UnsupportedFeatureError   iz.zip: UnsupportedFeatureError
```

## Docs and spec

`openspec/specs/format-zip/spec.md` — requirement renamed and rewritten, proposed through
`openspec new change --schema minimalist` and archived as the last task, same PR. One
scenario in, one scenario out (heading merged with #285's H4 rename: "multi-volume ZIP
join and refusal", since it now covers both). Also `dev-docs/formats/zip.md`
(§at-a-glance, §2.2, §3, §5, §6, §7), `dev-docs/open-issues.md` P2 (half of "detected and
rejected" stopped being true) and P17 (whose quoted regex this change edits),
`dev-docs/IDEAS.md`, `dev-docs/topics/prefixed-archives.md` (ZIP now shares P17's
`vol.exe.001` blind spot — noted, not fixed, and not made worse: it joined the existing
pattern rather than adding another), and `docs/formats.md`.

## Gates

`./scripts/check.sh --fix` clean (both type checkers). `./scripts/test.sh --all-configs`
green in all three: `[all]` 2689 passed / 23 skipped, `[all-lowest]` 2689 / 23,
`[core-only]` 2176 / 479.

Cross-platform review per AGENTS.md: no `read_text`/`open` without `encoding`, all new
fixture names Windows-legal, no path strings compared, and no test relying on two names
differing only by case. `7z`-dependent tests are gated with `requires_binary`.

https://claude.ai/code/session_014o1iwDmwTBR52DWFtGRmVE


src/archivey/internal/volumes.py
```diff
@@ -25,16 +25,31 @@
...
 SourceSequence = Sequence[SourceItem]
 
-_7Z_VOLUME_RE = re.compile(r"^(?P<base>.+\.7z)\.(?P<part>\d+)$", re.IGNORECASE)
+# 7-Zip's ``-v`` writes ``name.7z.001``/``name.zip.001`` — in both cases a *raw byte
+# split* of one finished archive, so the parts concatenate back into the original and
+# one pattern serves both. Info-ZIP's ``name.z01 … name.zip`` deliberately does not
+# match: that is a true spanned set addressed by (disk, offset), which concatenation
+# cannot reconstruct, and it is refused in the ZIP backend instead.
+_NUMBERED_VOLUME_RE = re.compile(
+    r"^(?P<base>.+\.(?:7z|zip))\.(?P<part>\d+)$", re.IGNORECASE
```
davitf: **F1** 🔴 `[blocking]` · `CONFIRMED` · wrong data, no error

`(?P<part>\d+)` accepts one- and two-digit suffixes, so `name.zip.1` / `name.zip.2` are now discovered as a volume set and concatenated. Those are the names `wget` and naive rotation produce for duplicate downloads of the same file — two *independent complete* ZIPs, not slices.

Measured, two complete single-member ZIPs side by side:

```
$ # backup.zip.1 contains first.txt, backup.zip.2 contains second.txt
=== PARENT (e1a1f24) ===
open backup.zip.1: members=['first.txt']  multivol=False
open backup.zip.2: members=['second.txt'] multivol=False
=== PR (22fe279) ===
open backup.zip.1: members=['second.txt'] multivol=True
open backup.zip.2: members=['second.txt'] multivol=True
```

Asking for `backup.zip.1` returns the *other file's* contents, silently. This is a regression against the parent commit, and it is the same "wrong bytes, no error" class as the `.partN` ordering bug this PR fixes.

The set-completeness check does not catch it: `_validate_numbered_volume_sequence` sees `[1, 2]`, which is exactly `1..N`.

Two internal regexes also disagree about the same name, which is what let this through:

```
name               join-regex   is_zip_split_segment_name
backup.zip.1       True         False
backup.zip.01      True         False
backup.zip.001     True         True
```

**Fix direction:** require three or more digits for the `zip` alternative, matching both `is_zip_split_segment_name`'s `zip\.\d{3,}` and what 7-Zip actually writes (`.001`, widening to 4 digits past 999) — and matching this PR's own spec text, which says `name.zip.001 … name.zip.00N`:

```python
r"^(?P<base>.+\.(?:7z|zip))\.(?P<part>\d{3,})$"
```

The same root cause pre-exists for `.7z.1` (verified: `discover_volume_siblings` joins `backup.7z.1` + `backup.7z.2` on the parent too), so applying `\d{3,}` to both alternatives fixes both and keeps the "one pattern serves both" shape. Nothing that 7-Zip emits stops matching.

Red–green candidate: two complete ZIPs named `x.zip.1` / `x.zip.2`; assert `open_archive("x.zip.1")` lists that file's own member.

Mixed-digit and mixed-case duplicates in one directory (`y.zip.1`, `y.zip.01`, `y.ZIP.001`) currently collapse into one base and raise `TruncatedError` — loud, so not part of this finding.

---

dev-docs/formats/zip.md
```diff
@@ -247,13 +263,31 @@ has no local header to point at, so there `payload_offset` is the EOCD-derived b
...
+to the original and why the rejoined EOCD reads disk `(0, 0)`. That is the same thing it
+does to `.7z`, so ZIP and 7z get the same answer for the same input.
+
+Info-ZIP's do not, and the difference is one of guarantee rather than of outcome on any
+one file. Its entries carry a *starting disk* alongside the offset, so concatenation is
+not the archive's own addressing — it merely coincides with it when every entry happens
+to sit on the disk a linear join would put it on. A three-disk `zip -s 64k` set does
+rejoin into something stdlib lists correctly; nothing about the format promises the next
+one will, and the rejoined EOCD still names disk `(2, 2)`. Refusing on those disk fields
+(§5) is refusing the thing we cannot resolve, not the thing that happened to work.
```
davitf: **F2** 🟡 `[important]` · `CONFIRMED` · doc claim is measurably wrong

"the difference is one of guarantee rather than of outcome on any one file" does not survive testing, and neither does the sentence that supports it.

I built the exact fixture this paragraph names — a three-disk `zip -s 64k` set — and joined it:

```
=== three-disk `zip -s 64k` ===
joined bytes: 150359 | head: b'PK\x07\x08'
LISTS: ['note.txt', 'payload.bin'] -> lists correctly: True
  read note.txt:    BadZipFile: Bad magic number for file header
  read payload.bin: BadZipFile: Bad magic number for file header
rejoined EOCD disk fields: (2, 2)
```

So: it lists correctly (the page is right), the EOCD is `(2, 2)` (the page is right), and **zero of its members read**. The outcome differs from `7z -v` too, not just the guarantee — on this very file.

This is not a large-set effect. Same result at every size I could build:

| set | disks | lists | members read OK |
| --- | --- | --- | --- |
| `zip -s 150k` | 3 | 16/16 | **0/16** |
| `zip -s 64k` | 3 | 2/2 | **0/2** |
| `zip -s 100k` | 5 | 16/16 | **0/16** |
| `zip -s 64k` | 12 | 12/12 | **0/12** |

The mechanism is worth having on the page because it explains why it is never a near miss: the CD's `cd_offset` is relative to the *last* disk, so stdlib's stub-prefix heuristic computes a global `concat` of "size of all preceding disks" (307200 on the 150k set) and applies it to every entry. Nothing lands anywhere near a local header. "Merely coincides with it when every entry happens to sit on the disk a linear join would put it on" implies a coincidence that can occur; for a genuinely multi-disk set it does not.

**Fix direction:** keep the structural argument, drop the outcome concession. Something like: a joined `zip -s` set *lists* correctly, because the central directory is intact on the final disk — and then reads nothing, because every entry's offset is relative to its own disk. Refusing on the non-zero EOCD disk fields (§5) refuses a set that cannot be read, not one that happened to work.

Worth noting because this wording replaced a stronger claim that was already walked back once in the PR body; it is closer, but still concedes more than the format does. The refusal itself is correct and unaffected — archivey rejects all of these before stdlib ever gets the chance (verified on real `zip -s` output for `.z01`, `.z02` and the final `.zip`, and for an explicitly-passed sequence of all parts).

---

dev-docs/formats/zip.md
```diff
@@ -290,7 +324,9 @@ ZIP-specific only. General extraction and name hazards are §2.4.
...
+| A `.z01`…`.zip` split set is refused with "rejoin first", while a `.zip.001`…`.00N` set beside it opens | **format** for now | Not an inconsistency: the first is a true spanned set addressed by (disk, offset), which stdlib cannot resolve and which concatenation reconstructs only by coincidence; the second is `7z -v` byte slices that rejoin into an ordinary ZIP (§3). Filename rules catch `.zNN`; EOCD disk fields catch Info-ZIP's final `.zip` part (`0xFFFF` is the ZIP64 sentinel, not a disk number). [`open-issues.md`](../open-issues.md) P2 |
+| A single `.zip.001` handed over without its siblings is refused rather than read as a ZIP | **archivey** | Joining needs parts `1..N` beside it. A first part often *does* carry a valid local header, so reading it alone would list members and then fail mid-data; "rejoin first" is the honest answer. A numbering gap is `TruncatedError` instead |
```
davitf: **F3** 🟡 `[important]` · `CONFIRMED` · stated reason is wrong

"reading it alone would list members and then fail mid-data" — it would not list anything. A first part carries no end-of-central-directory record (that lives in the last part), so stdlib refuses at open:

```
part 1 alone: stdlib raised at OPEN -> BadZipFile: File is not a zip file
part 1 starts with a local file header: True
part 1 contains an EOCD signature: False
```

The first half of the sentence is right (`PK\x03\x04` at offset 0) and the refusal is right; only the consequence is wrong. It matters because this row is the page's answer to "why not just read the part I was handed", and the honest answer is stronger than the one written: you cannot read it at all, not even a listing, because the index is in another file.

**Fix direction:** "…so it looks like a ZIP to a detector, but the central directory is in the last part — stdlib refuses at open with `File is not a zip file`. 'Rejoin first' names the actual problem."

---

openspec/specs/format-zip/spec.md
```diff
@@ -167,24 +167,31 @@ target bytes. Under `RAISE`, listing halts with `DiagnosticRaisedError`.
...
-with non-zero EOCD disk fields, 7-Zip `.zip.NNN`, or ZIP64 locator `disks > 1` —
-raises `UnsupportedFeatureError` (not `CorruptionError`, `FormatDetectionError`,
-or a raw stdlib `BadZipFile`).
+### Requirement: Join 7-Zip .zip.NNN sets; reject spanned ZIP cleanly
+
+7-Zip's `-v` byte-splits one finished single-disk ZIP into
+`name.zip.001 … name.zip.00N`, exactly as it splits `name.7z.NNN`. With every
+part `1..N` present beside the one named, `open_archive` SHALL concatenate them
+and read the result as the ordinary ZIP it is, from any part, reporting
+`ArchiveInfo.is_multivolume = True` and `extra["zip.volume_count"] = N`.
```
davitf: **F4** 🟢 `[nit]` · `CONFIRMED` · says which object carries `is_multivolume`, not which carries `extra`

`ArchiveInfo.is_multivolume` is qualified; the bare `extra["zip.volume_count"]` next to it is not. Both live on `ArchiveInfo` — confirmed at `zip_reader.py:1460`, and it does read back:

```
open_archive("name.zip.001").info.extra  ->  {'zip.volume_count': 4}
reader.members()[0].extra                ->  {}
```

`ArchiveMember.extra` exists too and is the more reachable of the two, so a reader who does not already know goes looking there and finds nothing. One word fixes it: `ArchiveInfo.extra["zip.volume_count"] = N`.

Same fix applies to `dev-docs/formats/zip.md:144`, which has the identical bare form.

For what it is worth, the value itself is right and matches 7z exactly — `sevenzip_reader.py:782,784` reports `is_multivolume=self._volume_count > 1` and `extra={"7z.volume_count": ...}`, and a joined `.7z.001` set returns `{'7z.volume_count': 4}` beside this one's `{'zip.volume_count': 4}`. The precedent claim in the PR body holds.

---

src/archivey/internal/backends/zip_reader.py
```diff
@@ -437,6 +437,10 @@ def __init__(
         # When measurement is on we open a Path ourselves to install a seek counter;
         # ZipFile does not close a caller-supplied file object, so we own it.
         self._owned_fp: BinaryIO | None = None
+        # >1 when the source is a joined volume set (ConcatenatedFile). 7-Zip's
+        # ``-v`` on a ZIP is a raw byte split, so the join is an ordinary ZIP and the
+        # count is reported rather than acted on.
+        self._volume_count: int = getattr(source, "volume_count", 1)
```
davitf: **F6** 💡 `[suggestion]` · `DISPROVEN` as a bug, re-filed as clarity

I went looking for a way to lose `volume_count` between `resolve_source` and here, because `getattr` on a `Path | BinaryIO` silently answers `1` for anything that does not carry the attribute — and `core.py` rewraps `reader_source` in three places after resolution (`PeekableStream`, the RAR volume reopen, `fix_stream_start_position`). A `SlicingStream` wrapper would not forward it, and the failure mode is the bad one: the split-name refuse fires on a set that was just successfully joined.

It does not happen. `ConcatenatedFile` is seekable, so the `PeekableStream` branch is skipped; the RAR branch is container-gated; and detection leaves the stream at 0, so `fix_stream_start_position` returns it unchanged. I checked the sharpest case I could construct — a shebang-prefixed ZIP byte-split into `pref.zip.001…003`, which has a non-zero `payload_offset`:

```
pref.zip.001 (stub + zip, split)   OPEN ok members=16 multivol=True extra={'zip.volume_count': 3}
pref.zip (unsplit control)         OPEN ok members=16 multivol=False extra={'zip.volume_count': 1}
```

So this is correct today and consistent with `sevenzip_reader.py:223`, which does the same thing — not a new pattern, and not worth changing on its own.

What is missing is the invariant. "Nothing between `resolve_source` and the backend rewraps the joined stream" is load-bearing, holds only by the three coincidences above, and is written down nowhere. A future wrapper added to `core.py` breaks a joined ZIP set and the symptom is a confusing refusal, not a crash. A line on this attribute saying the source must reach the backend unwrapped for the count to survive would cost nothing and is the kind of thing the tree cannot currently tell a future editor.

---

tests/test_volumes.py
```diff
@@ -34,6 +34,41 @@ def test_discover_7z_volume_siblings_natural_order(tmp_path: Path) -> None:
...
+    # 7-Zip's -v splits a .zip the same way it splits a .7z, so the same numbered-part
+    # discovery serves both.
+    for name in ("set.zip.010", "set.zip.002", "set.zip.001"):
+        (tmp_path / name).write_bytes(b"")
+    siblings = discover_volume_siblings(tmp_path / "set.zip.002")
+    assert siblings is not None
+    assert [p.name for p in siblings] == ["set.zip.001", "set.zip.002", "set.zip.010"]
+
+
+def test_discover_orders_parts_when_base_contains_partN(tmp_path: Path) -> None:
```
davitf: **F7** 🟢 `[nit]` · `CONFIRMED` · this test's redness depends on the filesystem

The bug it pins is real — I reproduced it on the parent commit, and it is the worst kind (silently wrong bytes, no error):

```
=== parent e1a1f24 ===
raw iterdir order:          ['my.part1.7z.003', 'my.part1.7z.001', 'my.part1.7z.002']
discover_volume_siblings -> ['my.part1.7z.003', 'my.part1.7z.001', 'my.part1.7z.002']
joined == original bytes:   False
```

But this test only goes red because `iterdir` returns creation order on ext4/tmpfs. On a filesystem that happens to return names sorted, the buggy sort key produces the right answer by luck and the test passes against unfixed code. Since CI includes macOS and Windows, the guard is weaker there than it looks here.

Not worth restructuring, because the deterministic coverage already exists: `test_volume_part_numbers_sort_stable` in `tests/test_property_safety.py` asserts `_numbered_part_number` directly over `.partN`-containing bases, which fails on the parent for any set of unique part numbers regardless of directory order. Worth one line here pointing at that, so a future reader does not mistake this for the whole guard.

Marking it because it is exactly the "no test in the suite catches this on platform X" shape rather than a missing example.

---

davitf (commented):
## 1. Maintainer briefing

**What this change is.** 7-Zip's `-v` flag byte-splits a finished `.zip` into `name.zip.001 … .00N`; those parts concatenate back into the original archive. Archivey already rejoined the identical split for `.7z.NNN` but refused it for `.zip`. This PR joins them by widening one regex in `internal/volumes.py` to cover both extensions, and teaches the two "rejoin first" refusals (in `core.py` and `ZipReader.__init__`) to fire on the *name* only when the set was **not** joined. Info-ZIP's `name.z01 … name.zip` — a genuinely spanned set addressed by (disk, offset) — keeps refusing. It also fixes a latent ordering bug found on the way.

**Snapshot.** Medium (~437 added / ~114 removed, 16 files). Gates re-run by me on `22fe279`, not copied from the body: `pytest` 2689 passed / 23 skipped, `./scripts/check.sh` all green (ruff, pyrefly, ty, `openspec validate --all` 29 passed, docs nav, strict docs build). All 23 CI check runs are green — the `mergeable_state: unstable` was a run in progress, not a failure; nothing here is red.

**Verdict: 🔄 Request Changes**, for one finding. (Recorded as a review comment because GitHub will not accept "request changes" from the account that opened the PR.)

**Main points**

- **F1 🔴** The widened regex takes `\d+`, not `\d{3,}`, so `name.zip.1` / `name.zip.2` — what `wget` and naive rotation produce for duplicate downloads — are joined as a volume set. `open_archive("backup.zip.1")` returns the *other file's* members, silently. Regression against the parent; one-token fix, and it also aligns the regex with this PR's own spec text.
- **F2 🟡** `dev-docs/formats/zip.md` §3's "the difference is guarantee, not outcome" is measurably wrong on the exact fixture it cites. A three-disk `zip -s 64k` set lists correctly and then reads **zero of its members**. Same at 3, 5 and 12 disks.
- **F3 🟡** The new §5 pitfall row says reading a lone first part "would list members and then fail mid-data". It cannot list at all — the central directory is in the last part, so stdlib refuses at open.
- **F4 🟢** `extra["zip.volume_count"]` is on `ArchiveInfo.extra`; neither the spec nor the handbook says which `extra`, and `ArchiveMember.extra` is the more reachable one.
- **F5 🟢** `review/docs-content/worker-inputs/verdicts-A.md:30` still quotes the removed error string `Incomplete 7z multi-volume set`. No location in this diff, so it is here rather than inline.
- **F6 💡 / F7 🟢** A disproven concern re-filed as a missing invariant, and a note that the ordering test's redness is filesystem-dependent. Both inline.

**What's fine** — and I checked these by running rather than reading:

- **The premise holds.** `7z a -tzip -v100k` then `cat` gives a byte-for-byte valid single-disk ZIP: same 322070 bytes, same 16 members, `testzip()` clean, content digest identical to the unsplit original.
- **The trap the PR claims to have handled is real, and all three edits are load-bearing.** Applying the diff file-by-file onto the parent: parent → refuse; `+volumes.py` → refuse; `+volumes.py +core.py` → refuse; `+volumes.py +zip_reader.py` → refuse; all three → opens, correct digest. A fix touching only `volumes.py` would indeed have looked right and still refused.
- **The ordering bug genuinely pre-existed.** On `e1a1f24`, `discover_volume_siblings` on `my.part1.7z.001…003` returns raw `iterdir` order and the join is not the original bytes. It is fixed here, and it affects `.7z.NNN` on `main` today.
- **Info-ZIP still refuses everywhere I pushed it** — real `zip -s` sets at 3, 5 and 12 disks, from `.z01`, from a middle part, from the final `.zip`, and as an explicitly-passed sequence of all parts (that last one via the EOCD disk-field check, exactly as the comment in `core.py` claims).
- **No false positives.** Plain, bare-22-byte-EOCD empty, shebang-prefixed, ZIP64, WinZip AES and ordinary single-volume `.zip` all still open; a lone `.zip.001` and a numbering gap still refuse; `.7z.NNN` is unchanged and `.rar` / `.partN.rar` / `.rNN` discovery is intact.
- **Tests are honest about their gating.** The joined-set fixture shells out to real `7z` behind `@requires_binary("7z")`; the Info-ZIP shapes are synthesised, with the reason written down. Nothing is gated on `zip`, which `scripts/setup-dev-env.sh` does not install — so nothing skips quietly on CI, macOS or Windows.
- **The spec did not get thicker.** One requirement renamed and rewritten, one scenario out and one in, archived in-PR. The `RENAMED` + `MODIFIED` pair rewrote the requirement **in place** rather than appending a stale duplicate — verified from the resulting `openspec/specs/format-zip/spec.md` diff, which `openspec validate --strict` would not have caught on its own.
- Cross-platform: no `read_text`/`open` without `encoding` in the new tests, no Windows-illegal fixture names, no fixture pair differing only by case.

---

## 3. Maintainer decisions

**1. Should the digit-count tightening that fixes F1 also apply to `.7z`, or only to `.zip`?**

**Why it matters.** F1 is the one blocking finding: today `open_archive("backup.zip.1")` hands back a different file's contents with no error. The fix is to require three or more digits after the extension. The question is only how wide to make it, because the identical hole already exists for `.7z.1` on `main` — the PR merged the two patterns into one regex, so the narrow fix means deliberately re-splitting them.

**Options.**

- **A — one pattern, `\d{3,}` for both.** Fixes the ZIP regression and the pre-existing `.7z` one together, keeps the single `_NUMBERED_VOLUME_RE` the PR argued for. Changes `.7z` behaviour in a PR titled "join `.zip` sets"; nothing 7-Zip emits stops matching (it always pads to at least 3), but a hand-renamed `vol.7z.01` set that joins today would stop.
- **B — `\d{3,}` for the `zip` alternative only.** Strictly scoped to the regression this PR introduces. Costs the shared pattern — the regex grows two branches — and leaves `.7z.1` knowingly broken, which under CONTRIBUTING's clean-as-you-go needs a written home rather than silence.
- **C — ship as is, fix separately.** Not recommended: it ships a silent wrong-data regression.

**Evidence.** Two complete single-member ZIPs named `backup.zip.1` and `backup.zip.2`; parent `e1a1f24` returns `['first.txt']` and `['second.txt']` respectively, PR `22fe279` returns `['second.txt']` for both. The same `discover_volume_siblings` call joins `backup.7z.1` + `backup.7z.2` on the parent, which is how I know the `.7z` half is not new. Full detail and the regex-disagreement table are in the F1 inline comment.

**Recommendation.** **A.** The PR's own thesis is that one splitter deserves one pattern and one answer; splitting the regex to protect a `.7z.01` set nobody produces trades the argument away for nothing. It also makes `_NUMBERED_VOLUME_RE` agree with `is_zip_split_segment_name`, whose `zip\.\d{3,}` is already the stricter of the two — that disagreement is what let this through.

**Default if you ignore this.** Option C ships: `.zip.1` / `.zip.2` pairs are silently joined and read as one archive.

---

---

src/archivey/internal/volumes.py
davitf: Follow-up on F1: the regex named groups in the comment above may render with the angle-bracket groups stripped, which makes the fix snippet unreadable. Stated without them, in case:

The change is one token in `_NUMBERED_VOLUME_RE` at `src/archivey/internal/volumes.py:34` — the part-number quantifier goes from `\d+` to `\d{3,}`. Everything else on that line stays as it is: same two named groups, same `(?:7z|zip)` alternation, same `re.IGNORECASE`.

That is the whole fix. `\d{3,}` is what 7-Zip actually emits (`.001`, growing to four digits past part 999), it is what `is_zip_split_segment_name` already requires for the `zip` branch, and it is what this PR's own spec text describes (`name.zip.001 … name.zip.00N`).

---

davitf: **Fixed** in `eb8cba6`, with **option A** — `\d{3,}` on both `.7z` and `.zip`, maintainer's call.

Reproduced your evidence first: `backup.zip.1` → `['second.txt']` on `22fe279`, `['first.txt']` on the parent. Also confirmed the `.7z` half predates this PR — `backup.7z.1` and `backup.7z.2` both return `['a.txt']` on `e1a1f24` too — which is what made the scope a real fork rather than a bug report.

B would have scoped the fix to `.zip` only, but that splits apart the single pattern this PR introduced and leaves the 7z half knowingly broken, needing a register row for it. Your point about the two regexes disagreeing decided it: `is_zip_split_segment_name` already required `zip\.\d{3,}`, and them disagreeing about the same name is the root cause. They now agree.

Red–green: `test_short_numeric_suffix_archives_read_their_own_contents` (the public-API half) and `test_discover_short_numeric_suffixes_are_not_a_volume_set` parametrised over both extensions, covering `.1` and `.01`. Both fail on the parent of this commit.

---

dev-docs/formats/zip.md
davitf: **Fixed** in `eb8cba6`. You were right that the wording still conceded too much, and I could reproduce it — the paragraph now leads with *lists correctly and then reads nothing*.

I traced the mechanism so the page states something checked rather than argued, and it is sharper than "coincidence":

```
disk sizes [65536, 65536, 19117]   EOCD at 150167, size_cd=75, offset_cd=19020
stdlib adjustment = eocd - size_cd - offset_cd = 131072   (= the two preceding disks)
first entry header_offset after it = 131076 -> bytes there b'\xaa\xff\x82\x12'
```

`offset_cd` is relative to the *last* disk, so the adjustment comes out as the total size of everything before it and is then applied to every entry. Nothing can land on a local header, which is why it is never a near miss at any size.

That replaced both the outcome concession and the "merely coincides" sentence; the §5 row that had inherited the same phrase is fixed too.

---

dev-docs/formats/zip.md
davitf: **Fixed** in `eb8cba6`, using your wording. The row now says the part opens with `PK\x03\x04` so it looks like a ZIP to a detector, but the central directory is in the last part — stdlib refuses at open with `File is not a zip file`, and not even a listing is available.

Agreed the honest answer is the stronger one: you cannot read it at all, and that is a better argument for "rejoin first" than a partial read would be.

---

openspec/specs/format-zip/spec.md
davitf: **Fixed** in `eb8cba6` — `ArchiveInfo.extra["zip.volume_count"]` in the spec, in the archived delta, and at `zip.md:144`, each noting that `ArchiveMember.extra` stays empty.

This one caught me out the same way while I was checking the PR body: I looked on `reader.members()[0].extra`, found `{}`, and flagged it as a possible discrepancy. Being the more reachable of the two objects is exactly the problem, so the page now says which is which rather than leaving it to be discovered.

---

src/archivey/internal/backends/zip_reader.py
davitf: **Fixed** as clarity, in `eb8cba6` — agreed it is not a bug today, and thank you for chasing it to the three reasons rather than stopping at "it works".

The attribute now carries the invariant: the count survives only while `core.open_archive` hands the joined stream through unwrapped, which holds because `ConcatenatedFile` is seekable (so the `PeekableStream` branch is skipped), the RAR reopen is container-gated, and detection leaves the position at 0 (so `fix_stream_start_position` returns the same object). The note says what breaks if a future wrapper lands between the two, and that the symptom is a confusing refusal rather than a crash — which is the part a reader could not recover from the code.

---

tests/test_volumes.py
davitf: **Fixed** in `eb8cba6` — one comment, as you suggested, rather than restructuring.

The test now records that it only goes red where `iterdir` returns creation order, and points at `test_volume_part_numbers_sort_stable` in `tests/test_property_safety.py` as the order-independent guard. Worth keeping both: this one is the readable reproduction, that one is what actually holds on macOS and Windows.

---

dev-docs/formats/zip.md
```diff
@@ -290,7 +336,9 @@ ZIP-specific only. General extraction and name hazards are §2.4.
...
-| A `.z01`…`.zip` or `.zip.001`…`.zip.00N` split set is refused with "rejoin first" | **format** for now | ZIP addresses entries by (disk, offset), which stdlib cannot resolve and naive concatenation cannot fake. Filename rules catch `.zNN` and `.zip.NNN`; EOCD disk fields catch Info-ZIP's final `.zip` part (`0xFFFF` is the ZIP64 sentinel, not a disk number). [`open-issues.md`](../open-issues.md) P2 |
+| A `.z01`…`.zip` split set is refused with "rejoin first", while a `.zip.001`…`.00N` set beside it opens | **format** for now | Not an inconsistency: the first is a true spanned set addressed by (disk, offset), which stdlib cannot resolve and which concatenation does not reconstruct at all (§3); the second is `7z -v` byte slices that rejoin into an ordinary ZIP (§3). Filename rules catch `.zNN`; EOCD disk fields catch Info-ZIP's final `.zip` part (`0xFFFF` is the ZIP64 sentinel, not a disk number). [`open-issues.md`](../open-issues.md) P2 |
```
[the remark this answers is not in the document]
davitf: **Fixed** in `f5261fe`, across all four sites — and testing the mechanism showed §3 was wrong too, in the other direction, so the fix is not the one suggested.

stdlib's adjustment is `eocd - size_cd - offset_cd`, which comes out as the total size of the disks *preceding the last one*. That is not a number correct for nothing; it is the number correct for the **final disk**. So a member whose local header sits on the last disk reads perfectly after a naive join, and only the members before it land in someone else's data.

Built with `zip -0 -s 64k` over a large member and a small one written last, then joined by hand:

```
big.bin    header_offset=131076   BadZipFile: Bad magic number for file header
last.txt   header_offset=150069   read OK, 17 bytes
```

So neither wording was right: not "cannot reconstruct at all" (this row), and not "only by coincidence" (the spec) — which member survives is not a coincidence, it is determined by where `zip -s` happened to cut. The earlier measurement of 3, 5 and 12-disk sets reading *no* members holds because none of them had a small tail member; that is why it looked absolute, and I overstated it in the original F2 comment as "nothing lands anywhere near a local header".

All five sites now say the same measured thing — a linear join lists correctly and then reads only whichever members happen to sit on the last disk — and §3 carries the evidence above. It is also a better argument for refusing than either previous version: partial readability that depends on the split point is worse than a clean failure. The failures at least stay loud, since a wrong offset misses `PK\x03\x04` rather than returning plausible bytes.

---

src/archivey/internal/volumes.py
```diff
@@ -25,16 +25,40 @@
 SourceItem = str | Path | BinaryIO
 SourceSequence = Sequence[SourceItem]
 
-_7Z_VOLUME_RE = re.compile(r"^(?P<base>.+\.7z)\.(?P<part>\d+)$", re.IGNORECASE)
+# 7-Zip's ``-v`` writes ``name.7z.001``/``name.zip.001`` — in both cases a *raw byte
+# split* of one finished archive, so the parts concatenate back into the original and
+# one pattern serves both. Info-ZIP's ``name.z01 … name.zip`` deliberately does not
+# match: that is a true spanned set addressed by (disk, offset), which concatenation
+# cannot reconstruct, and it is refused in the ZIP backend instead.
```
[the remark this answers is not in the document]
davitf: **Fixed** in `f5261fe` — the comment now says a linear join lists correctly and then reads only whichever members happen to sit on the last disk. See the §5 thread for the measurement; the short version is that stdlib's one adjustment is exactly right for the final disk, so this is partial readability rather than none.

---

tests/test_zip.py
```diff
@@ -487,6 +492,84 @@ def test_volume_shaped_name_honours_explicit_non_zip_format(tmp_path: Path) -> N
...
+    (tmp_path / "set.zip.002").unlink()
+    with pytest.raises(TruncatedError, match="Incomplete multi-volume set"):
+        open_archive(first)
+
+
+def test_infozip_spanned_set_still_refused(tmp_path: Path) -> None:
+    """Info-ZIP ``.z01 … .zip`` keeps refusing, siblings present or not.
+
+    ``zip -s`` writes a true spanned set: entries are addressed by (disk, offset) and
+    the parts do not concatenate into a readable ZIP, so it stays a different thing
```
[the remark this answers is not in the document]
davitf: **Fixed** in `f5261fe` — the docstring drops "do not concatenate into a readable ZIP" for the measured behaviour: a linear join lists correctly and then reads only whichever members happen to sit on the last disk, which is what makes it a different thing from 7-Zip's byte slices. Evidence in the §5 thread.

---

dev-docs/formats/zip.md
```diff
@@ -305,23 +353,26 @@ ZIP-specific only. General extraction and name hazards are §2.4.
...
 | Sniff unflagged names for UTF-8 validity; do not guess legacy codepages | Validation is near-conclusive; guessing has no oracle and a plausible wrong name is worse than a visible garble | An off-the-shelf charset detector, which can override a *valid* UTF-8 string with a legacy guess |
-| Reject split sets rather than approximate them | Naive segment concatenation is unreliable and would mis-read data rather than fail | Concatenating segments and hoping |
+| Join `7z -v` byte slices; reject Info-ZIP spanned sets | The first are slices of one finished archive and rejoin exactly — archivey already rejoins the identical split for `.7z.NNN`, so refusing here answered the same input two ways. The second cannot be reconstructed by concatenation at all | Refusing both (the shape of the rule was the filename, not the structure); concatenating spanned segments and hoping |
```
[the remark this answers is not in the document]
davitf: **Fixed** in `f5261fe` — the decision row now says a linear join of an Info-ZIP set lists correctly and then reads only the members that happen to sit on the last disk. Measurement in the §5 thread; the refuse argument is stronger for it, since partial readability that depends on the split point is worse than a clean failure.

---

PR #290, opened 2026-09-03
docs: add the RAR format handbook page

davitf:
## Goal

`dev-docs/formats/rar.md`, in the eight-section shape `zip.md` settled.

**Non-goals:** no `src/` changes. Registers keep their status (P6/P9/P11/P17, O1/C1, the `unar` known-issues entry) and are linked, not restated.

## What is structurally new

RAR is the first format whose read path crosses a **process boundary**: native parser for metadata, an `unrar` subprocess for member data. §2's per-stage prompt — *who does the work / what is format-specific / what is refused* — does not reach that, so a delegating stage answers a fourth question: **what crosses the boundary**. §2.3 is argv construction one way, an exit code plus a byte count the other.

The boundary is a §1 shape property rather than an implementation note, so its consequences stay attached to it: listing without the binary, the path requirement and the temp copy, the member name as an argv token, two numbers of error, pipes that cannot seek, and one banner probe per process.

`pair-workflow.md` §Format page structure gains one clause for the fourth question; the shape otherwise held.

## Carried from the unmerged PR #101

That branch's piping investigation was never merged and is **not** superseded by `alternative-rar-decompressors.md` — that one asks which binary can decode RAR data, #101 asks how the archive can be handed to it. Rather than merge a 260-line notebook into a directory layout that changed under it, the conclusions are on the page and the measurements are runnable: **`scripts/exploration/rar_unrar_input_matrix.py`**, alongside the existing `rar_decompressor_matrix.py`. Every row re-measured against unrar 7.00 here.

The load-bearing fact: `unrar` **seeks** the archive and refuses every non-seekable input, so the temp copy for a stream source is not laziness and the pipe family of ideas is closed. Worth recording because unrar's own switch list advertises `-si` "Read data from standard input" — a `rar` compressor switch that is rc 7 on unrar. A `memfd` is the one door left open and does not survive multi-volume: `unrar` discovers later volumes by filename, so part 1 as an anonymous fd stops at 839 of 1600 bytes, exactly as a byte-concatenation does.

## Self-review

A sub-agent reviewed the PR with `code-review-skill` + `unslop` and returned 28 findings. All fixed. The two that mattered:

- **The §4 safety claim was wrong.** It said RAR3 names have "no amplification to cap separately" — inherited verbatim from the archived RAR review's *"not a finding"* list, which is the promote-an-archived-conclusion failure `pair-workflow.md` warns about. RAR3 names are compressed, and the RLE branch emits ~103 characters per input byte while `_std_byte()` returns `?` without stopping the loop. Measured: 240 KB of input → **10.9 s** of CPU in `parse_rar_archive`, linear, ~13 MB transient per 64 KiB header. Same class as the vint bomb the page celebrates two bullets below.
- **"Not redistributable" was the wrong licence claim**, twice. UnRAR licence clause 3: *"The UnRAR utility may be freely distributed."* It is **non-free**, which is what the rest of the repo says and what the "no pip extra can ship it" consequence actually rests on.

Also fixed: `cryptography` is needed for RAR3/RAR4 header decryption too; the exit-10 row omitted the hash suppression the code applies; `comment` and `accessed`/`created` are never set and the table implied otherwise; §7 presented a question P9 has already decided; the 7z row overstated what the plugin buys (no stdin password, missing member → rc=0, no `path;n`); RAR5 does record and validate a volume number; and a dozen smaller precision fixes.

## Tests added

Two §8 rows cited tests that did not pin them. Rather than weaken the claims:

- `test_listing_and_stored_reads_need_no_unrar` — the native-metadata split end to end, with **no binary on `PATH` at all**.
- `test_stored_nonsolid_archive_spawns_no_unrar_process` — the §2.3 zero-subprocess measurement.
- `AccessCost.SOLID` / `solid_block_count is None` asserted in `test_basic_solid_stream_and_random` (the existing `[rar]` cost-receipt param is a *nonsolid* archive, so the solid half had no test).

## Verify

```bash
./scripts/check.sh && ./scripts/test.sh --all-configs
python3 scripts/exploration/rar_unrar_input_matrix.py
```

Green: `[all]` 2719 passed / 23 skipped, `[all-lowest]` 2719 / 23, `[core-only]` 2198 / 487. All 53 links resolve (file and anchor); all 25 §8 pointers name test functions that exist — both checked mechanically.

## Open, and named on the page rather than filed

`seekable_members=True` is quietly not honoured on the `unrar` route; `accessed`/`created` are parsed and discarded; the spec SHALL-requires a `PackageNotInstalledError` for an "incompatible" `unrar` and no version floor exists; and `docs/formats.md` still tells users that solid random opens "may use explicit temp materialization" — they do not.

https://claude.ai/code/session_01UwQFvMKi6oXGE8hbWzkj8x


dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+| Write | **Not shipped**, for any format — no `archivey.create`, no writer module (`PLAN.md` phase 9) |
+| Source | Seekable only, in both access modes |
+| Listing cost | `INDEXED` |
+| Access cost | `SOLID` for a solid archive, `DIRECT` otherwise; `solid_block_count` is always `None` (§1) |
+| Stream capability | `SEEKABLE` — about the *source*. A member stream backed by `unrar` is a pipe and is never seekable (§2.3) |
+| Core dependencies | None **for listing**. Member data needs the RARLAB `unrar` binary on `PATH`, which no pip extra can supply |
+| Optional | `[recommended]` (`cryptography`): RAR5 header decryption. BLAKE2sp needs nothing — it is implemented on stdlib `hashlib` |
+| Refuses | Non-seekable sources · a non-RARLAB `unrar` (no fallback to `unar` / `7z` / `bsdtar` / `unrar-free`) · a member whose name contains `*` or `?`, on the `unrar` path only · a later volume opened without its first · writing |
+
+**Claimed but not shipped.** [`format-rar`](../../openspec/specs/format-rar/spec.md)
```
davitf: We're phasing out the specs, these documents will be the source of truth. What's written there is too convoluted and hard to consume. Specific scenarios will be exercised via tests that will be mentioned from the docs. So you can remove references to the spec, at most link it as a reference; please see what the zip page did and tell me if it's different from what I said

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+overstates what is checked: `find_rarlab_unrar` reads the binary's banner for `UNRAR` plus
+`Alexander Roshal`/`RARLAB` and applies **no version floor**, so an ancient RARLAB build
+passes identification and fails later, per member.
+
+## 1. Shape
+
+Four properties generate most of this page.
+
+```
+  [ SFX stub? ]  [ magic ]  [ MAIN ]  [ FILE hdr | packed data ] × n  [ ENDARC ]
```
davitf: what's MAIN, and what's in it? it would be nice to have a table of the basic descriptions of these, like the zip page

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+  ─────────────────────────────────────────────────────────── │ ───────────────
+   archivey's parser reads everything above                    │
+                                                    ╔══════════▼═══════════╗
+                                                    ║  unrar  (subprocess) ║
+                                                    ╚══════════════════════╝
+```
+
+**The metadata is ours; the bytes are another program's.** RAR compression is proprietary
+and specified only by its reference implementation, whose licence permits decompression and
+forbids reverse-engineering a compressor. So the header layout was reverse-engineered and
```
davitf: not reverse engineered (at least not totally), they do publish a reference to the metadata, right?

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+  the copy to one member and make the disk-vs-RAM question much less interesting. Nobody has
+  measured either.
+- **What should `seekable_members=True` do on a pipe-backed member?** Three defensible
+  answers — refuse the flag for RAR the way a non-seekable source is refused, honour it by
+  respawning `unrar` on a backward seek (which is a whole-archive decode for a solid member
+  and an O(member) one otherwise), or report through `CostReceipt` that it was not applied.
+  Today it is accepted, quietly not honoured on the `unrar` route, and honoured on the direct
+  one, so the same archive answers differently per member. No measurement exists for what a
+  respawn would cost, and no corpus evidence for how often a caller seeks inside a RAR member
+  at all.
```
davitf: respawn on a backwards seek. it's the same reopen approach we take for other formats. we should never not honour it; consistency between formats is a core promise of the library

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+settled by reading more code. Distinct from §5, which is behaviour a caller already sees.
+
+- **Should the pipe routes report their cost, and in which channel?** Two costs are invisible
+  today and they are not the same shape. The whole-archive temp copy for a stream source is
+  P11's open question — a `CostReceipt.notes` entry, a diagnostic, or both. The other has no
+  home: an out-of-order solid `open()` is a full decode, and a second one is a second full
+  decode, with `AccessCost.SOLID` the only signal and no way to distinguish the first from the
+  fiftieth. Whether that is one answer or two depends on whether the boundary itself deserves
+  a note, which nobody has decided.
+- **Should the stream-source copy go to RAM instead of disk?** An anonymous `memfd` is
```
davitf: This might be interesting, but being linux-only makes it a bit uncomfortable. This reminds me of the "hack" rarfile uses: for files < 20MB, it writes a synthetic rar temp file with only the member to be decompressed. we could do that for memfd files: the limit would then be the compressed member size (for non-solid), not the whole archive size.

Now that I talk about it, a mention of this rarfile hack might be worth; I think there's an idea about it already, read IDEAS.md or search

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+| Reading one member of a solid archive out of order decodes the whole archive, and doing it twice decodes it twice | **format** / **archivey** | No per-block boundaries to resume from (§1); the spec's temp-directory strategy that would amortize it is unimplemented (§2.4). `AccessCost.SOLID` is the signal |
+| Handing over a `BytesIO` writes a full-size copy of the archive to `/tmp`, with nothing in `diagnostics` or `cost.notes` | **archivey** | `unrar` needs a path. The trigger is per-member, so the first stored member is free and the next compressed one is not. [`open-issues.md`](../open-issues.md) P11 |
+| A member whose name contains `*` or `?` cannot be read, though it lists fine | **library** | `unrar` include masks treat both as wildcards and offer no escape, so the member cannot be addressed unambiguously; `UnsupportedFeatureError` is preferred over possibly returning another member's bytes (§2.3). A stored member of the same name reads fine, since it never reaches `unrar` |
```
davitf: Didn't we fix this? Or maybe we considered a fix but haven't implemented it yet. The idea was to pass an actual wildcard to unrar (e.g. for a file named a*.txt, passing `a*.txt` or  `a?.txt` would match it; we'd have to figure out what other files would also be matched and output by unrar, and skip them

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+| `seekable_members=True` is accepted, and the member stream is still not seekable | **archivey** | Honoured only on the direct-slice route. An `unrar`-backed member is a pipe: `seekable()` is `False` and `seek()` raises `io.UnsupportedOperation`, while `reader.member_streams` still reports `SEEKABLE` and no diagnostic is emitted. [`docs/access-and-cost.md`](../../docs/access-and-cost.md) describes the flag's "otherwise" case as re-decompressing from the start, which is not what happens here. No register row covers it yet |
+| Reading one member of a solid archive out of order decodes the whole archive, and doing it twice decodes it twice | **format** / **archivey** | No per-block boundaries to resume from (§1); the spec's temp-directory strategy that would amortize it is unimplemented (§2.4). `AccessCost.SOLID` is the signal |
+| Handing over a `BytesIO` writes a full-size copy of the archive to `/tmp`, with nothing in `diagnostics` or `cost.notes` | **archivey** | `unrar` needs a path. The trigger is per-member, so the first stored member is free and the next compressed one is not. [`open-issues.md`](../open-issues.md) P11 |
```
davitf: BytesIO or any other stream. see on a comment below a possible way to reduce it

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+
+## 5. Sharp edges
+
+*Where it lives*: **format** — inherent, no implementation fixes it · **library** — the
+`unrar` binary's behaviour, fixable only upstream or by replacing it · **archivey** — ours.
+
+| What you see | Where it lives | More |
+| --- | --- | --- |
+| Listing an archive works on a machine where reading it fails | **format** | The compressor is proprietary and its reference tool is not redistributable, so member data needs a binary no pip extra can supply (§1, §3). `PackageNotInstalledError` names it and names the lookalikes that will not be accepted |
+| `seekable_members=True` is accepted, and the member stream is still not seekable | **archivey** | Honoured only on the direct-slice route. An `unrar`-backed member is a pipe: `seekable()` is `False` and `seek()` raises `io.UnsupportedOperation`, while `reader.member_streams` still reports `SEEKABLE` and no diagnostic is emitted. [`docs/access-and-cost.md`](../../docs/access-and-cost.md) describes the flag's "otherwise" case as re-decompressing from the start, which is not what happens here. No register row covers it yet |
```
davitf: this must be fixed, see comment below

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+| A corrupt encrypted member can be reported as a wrong password | **library** | `unrar` reports both as exit 2/3 with empty output on RAR4 and exposes no signal to separate them; the bias is toward the actionable message (§2.3) |
+| An SFX archive that is also split is unreadable from every one of its files | **archivey** | Sibling discovery needs the archive extension immediately before the part number; an SFX first member replaces it. Shared with 7z and ZIP — [`open-issues.md`](../open-issues.md) P17 and [`topics/prefixed-archives.md`](../topics/prefixed-archives.md) §6 |
+| A RAR on a pipe or socket cannot be opened at all, in either access mode | **format** | Block headers are chained forward but the walk still seeks; nothing is buffered for you (ADR [0010](../decisions/0010-no-silent-buffer-nonseekable.md)) |
+| `encoding=` is accepted and has no effect | **archivey** | RAR names are decoded by the parser. The call logs that the value will not be applied rather than failing |
+| A compressed member reports its compression as `UNKNOWN` with a level | **format** | The algorithm is proprietary and has no public name to report. The level is what the header carries |
```
davitf: interesting. should we have a "RAR" stream format for reporting these?

dev-docs/formats/rar.md
```diff
@@ -0,0 +1,625 @@
...
+| Every RAR5 symlink and hard link has no `hashes` entry, where ZIP and 7z have one | **format** | The stored field covers zero bytes, so the only honest answer is no digest (§2.2). RAR3/4 keeps its genuine one |
```
davitf: it's arguable whether we should keep hashes for RAR3/4. it's the hash of the link target name, not its contents, right? that's not very useful and may be misleading. potential open issue

dev-docs/formats/rar.md
davitf: Checked, and no — it's not different from what you said. `zip.md` links specs as bare references in three places (`packaging-and-extras` for the extras naming, `safe-extraction` in §2.4, the three in §9) and reasons *from* spec text exactly once, in §2.5, where the point is historical: `format-zip` used to specify streaming write and no longer does.

This page departed from that in three places, all now rewritten to state behaviour:

- The At-a-glance box argued SHALL vs MAY and quoted the requirement text. It now just says the two things a reader might expect and won't find: no version floor on the `unrar` banner check, and nothing amortizing repeated solid random reads.
- §4 cited "the spec's constrain-unrar-argv-by-call-site requirement" as the authority the argv hardening answers to. Now it says how the hole survived a rule written to prevent it, which is the part worth remembering.
- A §5 row called the temp-directory strategy "the spec's". Now: nothing caches the decode.

Spec links survive in §9 and at the two reference sites `zip.md` also uses.

One knock-on worth flagging: `pair-workflow.md` §Format page structure still says At-a-glance "distinguishes *claimed* from *shipped* — **a spec requirement** for something that does not exist says so here". That phrasing assumes specs are the thing making claims. If they're being phased out, that line wants rewording too — probably toward "anything a reader would expect and won't find". I haven't touched it; say the word and I will.

---

dev-docs/formats/rar.md
davitf: Not fixed — considered and left at the conservative half. Verified against the current tree: `_member_include_switch` (`rar_unrar.py`) still raises `UnsupportedFeatureError` on `*` or `?`, and `test_unrar_member_include_switch_rejects_wildcards` pins that.

Your approach is right and I've recorded it. The mask can carry the wildcard through — `a*.txt` does match a member literally named `a*.txt` — and the parser already has the full member list at that point, so it can compute which *other* members the mask matches and skip their bytes in the pipe. That makes it archivey's gap rather than unrar's, so I've retagged the §5 row **archivey — fixable** and it's item 3 in the new temporary §10.

One wrinkle for whoever picks it up: the skipping has to be demux-shaped, not filter-shaped. `unrar p` emits the matched members concatenated with no framing, so the fix reuses the same declared-size slicing the solid pipe uses — and inherits the same emission-policy caveat (§4), since a matched *link* member contributes zero bytes in both generations while its declared size is non-zero.

---

